PR #68, opened 2013-08-05
[JENKINS-17850] Add multiple Gerrit server support

mathieu-wang:
Refactor methods related to one Gerrit server to a new class GerritServer and
keep a list of GerritServer objects in PluginImpl.
Refactor all related classes and tests.


gerrithudsontrigger/src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer/remove.jelly
```diff
@@ -11,6 +11,9 @@
         <l:main-panel>
             <form method="post" action="removeConfirm">
                 <j:choose>
+                    <j:when test="${it.isLastServer()}">
+                        Cannot remove the last server!
+                    </j:when>
```
fredg02: Hi, I was wondering why the last Gerrit server should not be removed. Can you explain the reasoning behind it?

rsandell: Oofh, that was a long time ago :)
But IIRC, it is because of legacy code throughout the plugin that expects there to be at least one server configured (because there used to be only one configured). So to be safe and not break too much this safeguard was added.
I don't know if there is anything left of that legacy code anymore though ;)

fredg02: Thanks for the insight. Legacy strikes again. ;) Would be nice to remove it at some point.

PR #194, opened 2015-01-07
[JENKINS-26323] Fix "Build Current Patches Only" by scanning for event

HedAurabesh:
The "Build Current Patches Only" feature currently identifies previous
builds based on their parameters as they were present when the build was
first scheduled.

If you use a plugin that alters the parameters after the job has started,
the cancelling of previous builds does not work, since the
GerritTrigger#cancelJob() method does not find the job anymore.

This is easily and trivially fixed, by not scanning for the parameters, but
instead for the GerritEvent that is stored in the GerritCause used to start
a build.

The only downside to this is, that the current patch-set scans for identity,
but not equality of the two events. After a Jenkins restart, they might not
be identical anymore, since Jenkins reschedules queued builds and can't
unify the two events via Stapler; as they are stored in separate files.

It'd be trivially easy to alter this to Event.equals(otherEvent), in case
this degradation is not wanted.


HedAurabesh:
Closed, because this was targeted against outdated code.

Pull request #195 is the resubmission.

PR #195, opened 2015-01-07
[JENKINS-26323] Fix "Build Current Patches Only" by scanning for event

HedAurabesh:
The "Build Current Patches Only" feature currently identifies previous
builds based on their parameters as they were present when the build was
first scheduled.

If you use a plugin that alters the parameters after the job has started,
the cancelling of previous builds does not work, since the
GerritTrigger#cancelJob() method does not find the job anymore.

This is easily and trivially fixed, by not scanning for the parameters, but
instead for the GerritEvent that is stored in the GerritCause used to start
a build.

The only downside to this is, that the current patch-set scans for identity,
but not equality of the two events. After a Jenkins restart, they might not
be identical anymore, since Jenkins reschedules queued builds and can't
unify the two events via Stapler; as they are stored in separate files.

It'd be trivially easy to alter this to Event.equals(otherEvent), in case
this degradation is not wanted.

[EDIT]
Duplicate of pull request #194. That one was aborted, since it targeted an outdated release tag.

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
@@ -63,6 +63,7 @@ import com.sonymobile.tools.gerrit.gerritevents.dto.events.CommentAdded;
 import com.sonymobile.tools.gerrit.gerritevents.dto.events.GerritTriggeredEvent;
 import com.sonymobile.tools.gerrit.gerritevents.dto.events.RefUpdated;
 import com.sonymobile.tools.gerrit.gerritevents.dto.rest.Notify;
+
 import hudson.Extension;
 import hudson.ExtensionList;
 import hudson.Util;
@@ -107,6 +108,8 @@ import java.util.Map.Entry;
 import java.util.Set;
 import java.util.StringTokenizer;
 import java.util.regex.PatternSyntaxException;
+
+import jenkins.model.Jenkins;
 
 import org.kohsuke.stapler.AncestorInPath;
 import org.kohsuke.stapler.DataBoundConstructor;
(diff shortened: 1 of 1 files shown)


rinrinne:
Did you test with manual trigger?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -1849,23 +1852,27 @@ public synchronized void scheduled(ChangeBasedEvent event, ParametersAction para
...
+        private void cancelJob(GerritTriggeredEvent event) {
             // Remove any jobs in the build queue.
-            List<hudson.model.Queue.Item> itemsInQueue = Queue.getInstance().getItems(job);
-            for (hudson.model.Queue.Item item  : itemsInQueue) {
-                List<ParametersAction> params = item.getActions(ParametersAction.class);
-                for (ParametersAction param : params) {
-                    if (param.equals(parameters)) {
-                        Queue.getInstance().cancel(item);
-                    }
+            List<hudson.model.Queue.Item> itemsInQueue = Queue.getInstance().getItems(myProject);
```
rsandell: myProject doesn't exist anymore since the last merged pull request.
Use getJob or job instead.

HedAurabesh: Thanks, will fix it ASAP. That's what you get for re-basing and just looking at the mergetool. :P

HedAurabesh:
@rinrinne Yes, I tested it. Manual triggers always abort all previous runs started for that change, no matter which patch-set was restarted or was previously running. It also does not matter, whether you restart via the special "manual trigger" page or via the "Retrigger / Retrigger All" page in the side panel. All previous runs get immediately aborted.

As far as I can tell, this is the same behaviour as in the previous implementation, where it checked the used parameters for equality.

Of course, manual runs and runs based on other change IDs are completely unaffected and continue running normally.


As a sidenote: I have no clue why the Jenkins test failed:
https://jenkins.ci.cloudbees.com/job/plugins/job/gerrit-trigger-plugin/392/

From my limited point of view, it does not look like it's related to the changes in this commit (especially because only one single test case failed).

scoheb:
Hi,

I had the same test failure happen to me while preparing a PR myself. This was going to be my fix:

```Diff
@@ -102,15 +103,11 @@ public class BuildCompletedRestCommandJobHudsonTest {
 
         PluginImpl.getInstance().getHandler().post(event);
 
-        //CS IGNORE MagicNumber FOR NEXT 9 LINES. REASON: TestData
-        long startedWait = System.currentTimeMillis();
-        while (project.getLastCompletedBuild() == null) {
-            if (System.currentTimeMillis() - startedWait > 10000) {
-                throw new RuntimeException("Timeout!");
-            }
-            System.out.println("waiting for build to complete...");
-            Thread.sleep(1000);
-        }
+        TestUtils.waitForBuilds(project, 1);
+
+        //CS IGNORE MagicNumber FOR NEXT 3 LINES. REASON: TestData
+        System.out.println("waiting for post build notification...");
+        Thread.sleep(5000);
```

It seems it is a timing issue. We cannot just simply wait for the build to be completed...We also need to wait until the post build notification is received. Hence, the additional "sleep".

Hope that helps!

Scott

rsandell:
Thanks @scoheb I'm currently debugging some flaky tests and this was one of them.
I'll push it directly onto master if that's ok.

scoheb:
By all means @rsandell !

Thanks

Scott

PR #197, opened 2015-01-12
Findbugs: null pointer dereference fixes.

rsandell:
The bump in core version introduced @CheckforNull on Jenkins.getInstance
and escalated to PluginImpl.getInstance.
Added some static helper/shortcut methods in PluginImpl so not too many
new if statements littering the rest of the code.

This commit should fix all current, including old,
"possible null pointer dereference" warnings.

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritManagement.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritManagement.java
@@ -46,8 +46,10 @@ import hudson.util.FormValidation;
 import java.io.IOException;
 import java.net.URLDecoder;
 import java.net.URLEncoder;
+import java.util.Collections;
 import java.util.List;
 
+import javax.annotation.CheckForNull;
 import javax.servlet.ServletException;
 
 import jenkins.model.Jenkins;
@@ -112,8 +114,10 @@ public class GerritManagement extends ManagementLink implements StaplerProxy, De
 
     @Override
     public ContextMenu doContextMenu(StaplerRequest request, StaplerResponse response) throws Exception {
+        Jenkins jenkins = Jenkins.getInstance();
+        assert jenkins != null;
         ContextMenu menu = new ContextMenu();
-        menu.add("newServer", Functions.joinPath(Jenkins.getInstance().getRootUrl(), Functions.getResourcePath(),
+        menu.add("newServer", Functions.joinPath(jenkins.getRootUrl(), Functions.getResourcePath(),
                 "images", "24x24", "new-package.png"), Messages.AddNewServer());
         for (GerritServer server : getServers()) {
             menu.add(server);
@@ -138,7 +142,9 @@ public class GerritManagement extends ManagementLink implements StaplerProxy, De
          * @return the list of descriptors containing GerritServer's descriptor.
          */
         public static DescriptorExtensionList<GerritServer, GerritServer.DescriptorImpl> serverDescriptorList() {
-            return Jenkins.getInstance()
+            Jenkins jenkins = Jenkins.getInstance();
+            assert jenkins != null;
+            return jenkins
                     .<GerritServer, GerritServer.DescriptorImpl>getDescriptorList(GerritServer.class);
         }
 
@@ -151,9 +157,12 @@ public class GerritManagement extends ManagementLink implements StaplerProxy, De
         public AutoCompletionCandidates doAutoCompleteCopyNewItemFrom(@QueryParameter final String value) {
             final AutoCompletionCandidates r = new AutoCompletionCandidates();
 
-            for (String s : PluginImpl.getInstance().getServerNames()) {
-                if (s.startsWith(value)) {
-                    r.add(s);
+            PluginImpl plugin = PluginImpl.getInstance();
+            if (plugin != null) {
+                for (String s : plugin.getServerNames()) {
+                    if (s.startsWith(value)) {
+                        r.add(s);
+                    }
                 }
             }
             return r;
@@ -167,7 +176,11 @@ public class GerritManagement extends ManagementLink implements StaplerProxy, De
     */
     @SuppressWarnings("unused") //Called from Jelly
     public List<GerritServer> getServers() {
-        return PluginImpl.getInstance().getServers();
+        PluginImpl plugin = PluginImpl.getInstance();
+        if (plugin == null) {
+            return Collections.emptyList();
+        }
+        return plugin.getServers();
     }
 
     /**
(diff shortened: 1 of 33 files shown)


rsandell:
Hoping to get some review help. Not sure I have all annotations set correctly for example...

@TWestling @jglick @stephenc @recampbell ?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritManagement.java
```diff
@@ -138,7 +142,9 @@ public String getDisplayName() {
          * @return the list of descriptors containing GerritServer's descriptor.
          */
         public static DescriptorExtensionList<GerritServer, GerritServer.DescriptorImpl> serverDescriptorList() {
-            return Jenkins.getInstance()
+            Jenkins jenkins = Jenkins.getInstance();
+            assert jenkins != null;
```
jglick: Note that it _is_ valid for `Jenkins.getInstance` to be null during shutdown, so in general you need to think about whether your code might be running in this mode and behave gracefully if so.

There is no good solution for `DescriptorExtensionList`. For `ExtensionList` you can use the new `lookup` method to behave smoothly during shutdown.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/PluginImpl.java
```diff
@@ -126,11 +130,20 @@ public Api getApi() {
...
      * @return the instance.
      */
+    @CheckForNull
     public static PluginImpl getInstance() {
-        return Jenkins.getInstance().getPlugin(PluginImpl.class);
+        Jenkins jenkins = Jenkins.getInstance();
+        if (jenkins != null) {
+            return jenkins.getPlugin(PluginImpl.class);
+        } else {
+            logger.error("INITIALIZATION Error, Jenkins could not be found, so no plugin!");
```
jglick: Probably better to just shut up.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/dependency/DependencyQueueTaskDispatcher.java
```diff
@@ -102,8 +101,13 @@ public DependencyQueueTaskDispatcher() {
...
      */
     public static DependencyQueueTaskDispatcher getInstance() {
+        Jenkins jenkins = Jenkins.getInstance();
+        if (jenkins == null) {
+            logger.error("INITIALIZATION ERROR? Could not find the Jenkins instance.");
+            return null;
+        }
         ExtensionList<DependencyQueueTaskDispatcher> dispatchers =
-                Jenkins.getInstance().getExtensionList(DependencyQueueTaskDispatcher.class);
+                jenkins.getExtensionList(DependencyQueueTaskDispatcher.class);
```
jglick: Use `ExtensionList.lookup` if your baseline is sufficiently new.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ToGerritRunListener.java
```diff
@@ -81,10 +86,10 @@ public static ToGerritRunListener getInstance() {
     }
 
     @Override
-    public synchronized void onCompleted(AbstractBuild r, TaskListener listener) {
+    public synchronized void onCompleted(AbstractBuild r, @Nonnull TaskListener listener) {
         GerritCause cause = getCause(r);
         logger.debug("Completed. Build: {} Cause: {}", r, cause);
-        if (cause != null) {
+        if (r != null && cause != null) {
```
jglick: Why?

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTriggerParametersTest.java
```diff
@@ -76,15 +64,12 @@
...
     public void setup() {
-        mockStatic(PluginImpl.class);
-        PluginImpl plugin = mock(PluginImpl.class);
-        when(PluginImpl.getInstance()).thenReturn(plugin);
-
         config = Setup.createConfig();
         GerritServer server = new GerritServer(PluginImpl.DEFAULT_SERVER_NAME);
         server.setConfig(config);
-        when(plugin.getServer(eq(PluginImpl.DEFAULT_SERVER_NAME))).thenReturn(server);
-        when(plugin.getFirstServer()).thenReturn(server);
```
jglick: Any particular reason you are deleting all this?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ToGerritRunListener.java
rsandell: Minimal effort to get rid of the warning that r might be null.
The method in RunListener is declared as ```public void onCompleted(R r, @Nonnull TaskListener listener) {}```

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTriggerParametersTest.java
rsandell: Yes, at some point in time it had become some strange abomination with mixed JenkinsRule and PowerMock, when the test didn't pass I started with simplifying the test code to understand what was going wrong in the tests and that fixed the issues.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/dependency/DependencyQueueTaskDispatcher.java
rsandell: The baseline is 1.565.3 which doesn't seem to have it.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritManagement.java
rsandell: That depends on when ```Jenkins.getInstance()``` becomes null, is it before or after ```Plugin.stop()```?

rsandell:
@reviewbybees

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ToGerritRunListener.java
jglick: But in fact it should be `@Nonnull Run`. Can fix that later in core, but in the meantime I think you can just add this annotation to your override.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritManagement.java
jglick: Not sure offhand.

PR #198, opened 2015-02-02
Parameter provider of Base64EncodedString is implemented for Rebuilder.

conan-z:
The new provider class helps rebuild plugin to correctly display
Base64EncodedString parameters. When rebuilder plugin rebuilt a Gerrit
event triggered build, the parameters page was not displayed properly
without this provider class.

Please consider the following test scenario of the provider class:

Given a Jenkins instance
 And Gerrit Trigger plugin installed on that Jenkins
 And an existing Gerrit instance configured on that Jenkins
 And Parameterized Trigger plugin installed on that Jenkins
 And Rebuilder plugin installed on that Jenkins
 And an upstream job that is triggered by Gerrit events
 And 'Commit message as human readable' is NOT set in the upstream job
 And once a build is finished, upstream job triggers a parameterized
     build of downstream job with 'current parameters'
When upstream job is triggered by a Gerrit event
 And the first build of both job completes
 And GERRIT_CHANGE_COMMIT_MESSAGE parameter is displayed properly in
     first build of both jobs
 And I rebuild the first build of the downstream job
Then the second build of downstream job should complete
 And GERRIT_CHANGE_COMMIT_MESSAGE parameter is displayed properly in
     second build of the downstream job

The test scenario is beyond scope of a unit test as it requires multiple
plugins preinstalled and it tests compatibility of these plugins.
Therefore its test should be written in Jenkins ATH rather than in
Gerrit Trigger project.

Our goal is to add such a test to ATH in the near future.


scoheb:
+1 for me.

conan-z:
The test failure is caused by changing dependency of rebuild plugin version from 1.1 to 1.22. And this test fails at Rebuuilder 1.21 too.

However, only Rebuilder 1.21+ provides extension point for other plugins to support rebuild plugin.

rinrinne:
You should fix test too. Perhaps it is not code but resource. Please check `src/test/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/matrix_build.xml`

PR #199, opened 2015-02-03
Parameter provider of Base64EncodedString is implemented for Rebuilder.

conan-z:
The new provider class helps rebuild plugin to correctly display
Base64EncodedString parameters. When rebuilder plugin rebuilt a Gerrit
event triggered build, the parameters page was not displayed properly
without this provider class. matrix_build.xml is modified to pass a
failed test after changing dependency of Rebuild from v1.1 to v1.22.

Please consider the following test scenario of the provider class:

Given a Jenkins instance
 And Gerrit Trigger plugin installed on that Jenkins
 And an existing Gerrit instance configured on that Jenkins
 And Parameterized Trigger plugin installed on that Jenkins
 And Rebuilder plugin installed on that Jenkins
 And an upstream job that is triggered by Gerrit events
 And 'Commit message as human readable' is NOT set in the upstream job
 And once a build is finished, upstream job triggers a parameterized
     build of downstream job with 'current parameters'
When upstream job is triggered by a Gerrit event
 And the first build of both job completes
 And GERRIT_CHANGE_COMMIT_MESSAGE parameter is displayed properly in
     first build of both jobs
 And I rebuild the first build of the downstream job
Then the second build of downstream job should complete
 And GERRIT_CHANGE_COMMIT_MESSAGE parameter is displayed properly in
     second build of the downstream job

The test scenario is beyond scope of a unit test as it requires multiple
plugins preinstalled and it tests compatibility of these plugins.
Therefore its test should be written in Jenkins ATH rather than in
Gerrit Trigger project.

Our goal is to add such a test to ATH in the near future.

[FIXED JENKINS-26756]

### DIFF
--- a/src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/parameters/Base64EncodedStringParameterValue/Base64EncodedStringParameterValue.jelly
+++ b/src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/parameters/Base64EncodedStringParameterValue/Base64EncodedStringParameterValue.jelly
@@ -27,6 +27,9 @@ THE SOFTWARE.
         xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:f="/lib/form"
         xmlns:i="jelly:fmt" xmlns:p="/lib/hudson/project">
         <f:entry title="${it.name}" description="${it.description}">
+            <div name="parameter" description="${it.description}">
+                <input type="hidden" name="name" value="${it.name}" />
                 <f:textbox name="value" value="${it.value}" readonly="true" />
+            </div>
         </f:entry>
 </j:jelly>
--- a/src/test/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/matrix_build.xml
+++ b/src/test/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/matrix_build.xml
@@ -86,9 +86,6 @@
         </hudson.model.Cause_-UpstreamCause>
       </causes>
     </hudson.model.CauseAction>
-    <com.sonyericsson.rebuild.RebuildAction>
-      <rebuildurl>../buildWithParameters?GERRIT_BRANCH=master&amp;GERRIT_CHANGE_ID=Icbd12345&amp;GERRIT_CHANGE_NUMBER=38506&amp;GERRIT_CHANGE_SUBJECT=Removing the obsolete graphics&amp;GERRIT_CHANGE_URL=http://gerrit/38506&amp;GERRIT_PATCHSET_NUMBER=1&amp;GERRIT_PATCHSET_REVISION=cbd12345&amp;GERRIT_PROJECT=platform/project&amp;GERRIT_REFSPEC=refs/changes/06/38506/1</rebuildurl>
-    </com.sonyericsson.rebuild.RebuildAction>
   </actions>
   <number>102</number>
   <result>FAILURE</result>
--- a/pom.xml
+++ b/pom.xml
@@ -79,7 +79,7 @@
         <dependency>
             <groupId>com.sonyericsson.hudson.plugins.rebuild</groupId>
             <artifactId>rebuild</artifactId>
-            <version>1.1</version>
+            <version>1.22</version>
             <optional>true</optional>
         </dependency>
         <dependency>
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/parameters/Base64EncodedStringParameterProvider.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/parameters/Base64EncodedStringParameterProvider.java
@@ -1,0 +1,56 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2014 Ericsson
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package com.sonyericsson.hudson.plugins.gerrit.trigger.hudsontrigger.parameters;
+
+import hudson.Extension;
+import hudson.model.ParameterValue;
+
+import com.sonyericsson.rebuild.RebuildParameterPage;
+import com.sonyericsson.rebuild.RebuildParameterProvider;
+
+/**
+ * An extension class to inject {@link Base64EncodedStringParameterProvider} to rebuild-plugin.
+ *
+ * @author Scott.Hebert &lt;Scott.Hebert@ericsson.com&gt;.
+ * @author Bowen.Cheng &lt;Bowen.Cheng@ericsson.com&gt;.
+ */
+@Extension(optional = true)
+public class Base64EncodedStringParameterProvider extends RebuildParameterProvider {
+    /**
+     * @param value ParameterValue to have rebuild page created for it.
+     * @return rebuild page for Base64EncodedStringParameterValue
+     * @see com.sonyericsson.rebuild.RebuildParameterProvider#getRebuildPage(hudson.model.ParameterValue)
+     */
+    @Override
+    public RebuildParameterPage getRebuildPage(ParameterValue value) {
+        if (value instanceof Base64EncodedStringParameterValue) {
+            return new RebuildParameterPage(Base64EncodedStringParameterValue.class,
+                "Base64EncodedStringParameterValue.jelly");
+        }
+
+        return null;
+    }
+
+}
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/parameters/Base64EncodedStringParameterValue.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/parameters/Base64EncodedStringParameterValue.java
@@ -39,7 +39,7 @@ public class Base64EncodedStringParameterValue extends StringParameterValue {
     private static final int ABBREVIATE_LENGTH = 10;
 
     /**
-     * Default constractor.
+     * Default constructor.
      *
      * @param name the name.
      * @param value the value.
@@ -50,7 +50,7 @@ public class Base64EncodedStringParameterValue extends StringParameterValue {
     }
 
     /**
-     * Default constractor.
+     * Default constructor.
      * @param name the name.
      * @param value the value.
      * @param description the description.


scoheb:
+1

src/test/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/matrix_build.xml
```diff
@@ -86,9 +86,6 @@
         </hudson.model.Cause_-UpstreamCause>
       </causes>
     </hudson.model.CauseAction>
```
rsandell: Normally removing xml data from test cases _(especially back compat test cases)_ would defeat the purpose of the test as is normally tests that loading of the data works, and if it doesn't code should be fixed and not data.

In this case its probably OK as the Action probably shouldn't be there due to the veto extension point. But I'll need to dig into the tests a bit first before I can say for sure.

src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/parameters/Base64EncodedStringParameterValue/Base64EncodedStringParameterValue.jelly
```diff
@@ -27,6 +27,9 @@ THE SOFTWARE.
         xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:f="/lib/form"
         xmlns:i="jelly:fmt" xmlns:p="/lib/hudson/project">
         <f:entry title="${it.name}" description="${it.description}">
+            <div name="parameter" description="${it.description}">
```
rsandell: does this actually work with the form posting? it looks like the json data would be ```name.parameter.name``` and ```name.parameter.value```. And description is added twice?

src/test/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/matrix_build.xml
conan-z: I agree that we normally should fix the code not he data.

In this case, the XML is only used by this single test so nothing else is affected by this change. The test case itself is not testing Rebuild plugin at all. Therefore, the test should survive version changes of such plugins. Also, the outer tag '...RebuildAction' still exists in rebuild but the inner tag 'buildurl' has been deprecated long time ago. Its corresponding variable only exists in the class for backwards compatibility reason.

src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/parameters/Base64EncodedStringParameterValue/Base64EncodedStringParameterValue.jelly
conan-z: Yes, it passes the manual test proposed in the commit message.

The div tag was added later as a bug fix for this commit even before it was pushed to GitHub.
Please take a look at this file from Rebuild: https://github.com/jenkinsci/rebuild-plugin/blob/master/src/main/resources/com/sonyericsson/rebuild/RebuildAction/StringParameterValue.jelly

rinrinne: I think we should follow Jenkins core. See https://github.com/jenkinsci/jenkins/blob/master/core/src/main/resources/hudson/model/StringParameterValue/value.jelly

rsandell: So did I merge too soon?

conan-z: The *div* element has to present in the jelly file, otherwise the JSON object cannot be parsed properly. The bug fix I mentioned is actually adding the *div* element. The *name* attribute tells Stapler  Request the jelly file serves as one entry if the *parameter* JSON object array. Without the div element, *Base64EncodedString* parameter is outstanding in the JSON and  cannot be picked up as one of the array element.
 Please take a look at line 289 of RebuildAction where the array elements are extracted: https://github.com/jenkinsci/rebuild-plugin/blob/master/src/main/java/com/sonyericsson/rebuild/RebuildAction.java#L289

rinrinne: I have 2 questions:

1. Why did you rename value.jelly to value.jelly to Base64EncodedStringParameterValue.jelly ?
2. If such `div` element is needed for handling parameters as an array by rebuild plugin, I think that dedicated jelly file should be provided. Because this plugin itself don't need it actually. WDYT?

I have concern that HTML structure for displaying this plugin own object has strong dependency with other plugin.

conan-z: Hi, thank you for the comments.

For Q1: I simply followed the convention from Rebuild plugin. All Jelly files for different types of parameters are under [this folder] (https://github.com/jenkinsci/rebuild-plugin/tree/master/src/main/resources/com/sonyericsson/rebuild/RebuildAction). I think by following the convention, this name is also more self-explainable. if you insist. I will change the name back to *value.jelly*.

For Q2: I understand your concern. But from Rebuild plugin's point, it might be impossible to provide a generic jelly file because only the project that defines the customized parameter knows how it should be displayed.

Maybe the following proposal could solve the problem?
We have two Jelly file (instead one) for the two different purposes. E.g. *value.jelly* displays Gerrit Trigger's own objects. The other one (*Base64EncodedStringParameterValue.jelly*) helps Rebuild renders the page properly?
I think in this way, we can solve the conflict and the strong dependency issue.

rinrinne: @conan-z I checked rebuild-plugin then found it has various jelly files for ParameterValue. It includes one for non-core class. e.g. MavenMetadataParameterValue.jelly for jenkins-maven-metadata-plugin, NodeParameterValue.jelly for nodelabelparameter-plugin.

And RebuildAction in rebuild plugin has a logic to load internal jelly files with the simple name of class. So I think that this issue can be solved without any patches for this plugin.

Maybe, rebuild plugin has changed the policy to support ParameterValue extended by other plugins?

BTW, I cannot find plugins that RebuildParameterProvider has been implemented.(googling...)
Do you know such plugin?

conan-z: @rinrinne Since renamed jelly file would never be loaded in Gerrit Trigger, is it better if the name is changed back to *value.jelly*? I fear this would potentially cause future bugs.

No, sorry, I don't know any plugins that implements this class either. I need to google it to find out too.

rinrinne: @conan-z Googling, I confirmed that almost all plugins which implements the extended ParameterValue class has `value.jelly` only in short class name directory. (Of course including jenkins-maven-metadata-plugin and nodelabelparameter-plugin too.)

So, my proposal for this issue is:
* Revert this patch (No change for GT)
* Create new PR for rebuild plugin that `Base64EncodedStringParameterValue.jelly` is added to the resource of RebuildAction.

conan-z: Hi, sorry for the late reply. I have been busy with some other things.

I have also done some searching. There are two other projects that I found providing Rebuild plugin support the exact same way as we do.

[matrix-combinations-plugin](https://github.com/jenkinsci/matrix-combinations-plugin)
* [MatrixCombinationsRebuildParameterProvider.java](https://github.com/jenkinsci/matrix-combinations-plugin/blob/master/src/main/java/hudson/plugins/matrix_configuration_parameter/MatrixCombinationsRebuildParameterProvider.java)
* [rebuild.groovy](https://github.com/jenkinsci/matrix-combinations-plugin/blob/master/src/main/resources/hudson/plugins/matrix_configuration_parameter/MatrixCombinationsParameterValue/rebuild.groovy)

[promoted-builds-plugin](https://github.com/jenkinsci/promoted-builds-plugin)
* [PromotedBuildRebuildParameterProvider.java](https://github.com/jenkinsci/promoted-builds-plugin/blob/master/src/main/java/hudson/plugins/promoted_builds/parameters/PromotedBuildRebuildParameterProvider.java)
* [value.jelly](https://github.com/jenkinsci/promoted-builds-plugin/blob/master/src/main/resources/hudson/plugins/promoted_builds/parameters/PromotedBuildParameterValue/value.jelly)

I we revert this PR, adding `Base64EncodedStringParameterValue.jelly` to the resource of `RebuildAction` in Rebuild plugin is not enough. A parameter provider class is needed in rebuild too. However, Rebuild doesn't really need such a provider class. So I don't think there is any problem adding changes to Gerrit Trigger.

rinrinne: @conan-z Thanks for example. I confirmed both of them have either `value.jelly` or `value.groovy` in the resource directory for `ParameterValue` class.

* In [matrix-combinations-plugin](https://github.com/jenkinsci/matrix-combinations-plugin/blob/master/src/main/resources/hudson/plugins/matrix_configuration_parameter/MatrixCombinationsParameterValue/value.groovy)
* In [promoted-builds-plugin](https://github.com/jenkinsci/promoted-builds-plugin/blob/master/src/main/resources/hudson/plugins/promoted_builds/parameters/PromotedBuildParameterValue/value.jelly)

So I think you should add dedicated page resource file instead renaming.

> A parameter provider class is needed in rebuild too in order to support Gerrit Trigger plugin.

If you are correct, you must show the reason why Rebuild Plugin can deal with [NodeParameterValue](https://github.com/jenkinsci/rebuild-plugin/blob/master/src/main/resources/com/sonyericsson/rebuild/RebuildAction/NodeParameterValue.jelly) correctly though [NodeLabel Parameter Plugin](https://github.com/jenkinsci/promoted-builds-plugin) which provides [NodeParameterValue](https://github.com/jenkinsci/nodelabelparameter-plugin/blob/master/src/main/java/org/jvnet/jenkins/plugins/nodelabelparameter/NodeParameterValue.java) does not have any [RebuildParameterProvider](https://github.com/jenkinsci/rebuild-plugin/blob/master/src/main/java/com/sonyericsson/rebuild/RebuildParameterProvider.java) extended class.

Certainly, Rebuild Plugin's implementation for `NodeParameterValue` is not enough because type information is lost. But I don't agree with you as of now because I think the reason you mentioned is not enough as well.

rinrinne: I found the issue.

I created `value.jelly` with the below contents on this patch.
```
<?jelly escape-by-default='true'?>
<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define"
        xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:f="/lib/form"
        xmlns:i="jelly:fmt" xmlns:p="/lib/hudson/project">
        <f:entry title="Test ${it.name}" description="${it.description}">
            <input type="hidden" name="name" value="${it.name}" />
            <f:textbox name="value" value="${it.value}" readonly="true" />
        </f:entry>
</j:jelly>
```
You can see the string `Test` in the value of `title` attribute for `f:entry` element.

On Jenkins with built plugin, you could see `GERRIT_CHANGE_COMMIT_MESSAGE` like below at parameters page in triggered build.
value-jelly-cut2
It means that renamed jelly file has not been used for original purpose since this patch. Perhaps core can display it as of now because core has fallback solution that displays it as `StringParameterValue`.

conan-z: @rinrinne Thank you for the comments.
As I proposed in previous comments, I agree with the solution (restoring the value.jelley). Renaming the file was careless. I will be more careful in the future.

conan-z: @rinrinne 
> If you are correct, you must show the reason why Rebuild Plugin can deal with NodeParameterValue correctly though NodeLabel Parameter Plugin which provides NodeParameterValue does not have any RebuildParameterProvider extended class.

##### Why it works?
When no provider is available, Rebuild plugin use an existing view provided by Rebuild plugin itself.
Please see code [here](https://github.com/jenkinsci/rebuild-plugin/blob/master/src/main/java/com/sonyericsson/rebuild/RebuildAction.java#L468).

##### Why it is not good anymore?
It is recommended to set values to `RebuildParameterPage` by implementing `RebuildParameterProvider` according to its [documentation](https://github.com/jenkinsci/rebuild-plugin/blob/master/src/main/java/com/sonyericsson/rebuild/RebuildParameterProvider.java#L49).
All page files under resource folder of Rebuild should be internal. Resource folder of Rebuild plugin should not be a "warehouse" for page files from every single plugin that provides support for Rebuild.

##### Why didn't the committer follow recommended instruction?
Please pay attention to the following dates:
* `RebuildParameterProvider` was first introduced by
[this commit](https://github.com/jenkinsci/rebuild-plugin/commit/9da2497ab558cc19dfe63e12b8adbcb89a7c2000) on Oct 6, 2013
* `NodeParameterValue.jelly` was added to Rebuild plugin by [this commit](https://github.com/jenkinsci/rebuild-plugin/commit/32d15659e677f7c22bafd05cee4e2fb2d143d6a4) on Jul 8, 2012

This means when `NodeParameterValue.jelly` was added to Rebuild plugin, there was no extension point provided by Rebuild yet. Adding a jelly file directly to Rebuild was probably  the only way (or hack) to provide Rebuild plugin support at the time.

PR #200, opened 2015-02-06
Enable background threads to impersonate SYSTEM

scoheb:
Jenkins plugin listeners and background threads should normally run as the SYSTEM user.

This commit allows the EventThreads to impersonate ACL.SYSTEM.

Addresses regression in [JENKINS-23152]

Change-Id: Ia0f5a223f70497ad44529c390aff0c711e414f67

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -61,7 +61,7 @@
         <dependency>
             <groupId>com.sonymobile.tools.gerrit</groupId>
             <artifactId>gerrit-events</artifactId>
-            <version>2.4.2</version>
+            <version>2.6.0</version>
             <!-- New source is here: https://github.com/sonyxperiadev/gerrit-events -->
         </dependency>
         <dependency>
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/JenkinsAwareGerritHandler.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/JenkinsAwareGerritHandler.java
@@ -28,29 +28,43 @@ import org.slf4j.LoggerFactory;
 
 import com.sonymobile.tools.gerrit.gerritevents.GerritHandler;
 import com.sonymobile.tools.gerrit.gerritevents.dto.GerritEvent;
+import com.sonymobile.tools.gerrit.gerritevents.workers.EventThread;
 import com.sonyericsson.hudson.plugins.gerrit.trigger.events.lifecycle.GerritEventLifecycle;
 
 /**
- * Specialization of GerritHandler that supports gerrit event's lifecycle.
- * @author Hugo Arès &lt;hugo.ares@ericsson.com&gt;
+ * Specialization of GerritHandler that supports gerrit event's
+ * lifecycle and takes care of custom EventThread creation.
  *
+ * @author Hugo Arès &lt;hugo.ares@ericsson.com&gt;
  */
-public class GerritHandlerLifecycle extends GerritHandler {
+public class JenkinsAwareGerritHandler extends GerritHandler {
 
-    private static final Logger logger = LoggerFactory.getLogger(GerritHandlerLifecycle.class);
+    private static final Logger logger = LoggerFactory.getLogger(JenkinsAwareGerritHandler.class);
 
     /**
      * Standard Constructor.
      *
-     * @param numberOfWorkerThreads the number of event threads.
+     * @param numberOfWorkerThreads
+     *            the number of event threads.
      */
-    public GerritHandlerLifecycle(int numberOfWorkerThreads) {
+    public JenkinsAwareGerritHandler(int numberOfWorkerThreads) {
         super(numberOfWorkerThreads);
+    }
+
+    /**
+     * Here we override the EventThread creation with
+     * one that impersonates System.
+     * @param threadName name of thread.
+     * @return new EventThread.
+     */
+    @Override
+    protected EventThread createEventThread(String threadName) {
+        return new SystemEventThread(this, threadName);
     }
 
     @Override
     public void notifyListeners(GerritEvent event) {
-        //Notify lifecycle listeners.
+        // Notify lifecycle listeners.
         if (event instanceof GerritEventLifecycle) {
             try {
                 ((GerritEventLifecycle)event).fireTriggerScanStarting();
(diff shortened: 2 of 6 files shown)


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/SystemEventThread.java
```diff
@@ -0,0 +1,65 @@
...
+        super(coordinator);
+    }
+
+    /**
+     * We perform the impersonation of the System user
+     * prior to execution.
+     */
+    @Override
+    public void run() {
+        ACL.impersonate(ACL.SYSTEM);
```
rsandell: ```ACL.impersonate(ACL.SYSTEM);``` will set the thread context to SYSTEM "forever" the normal pattern to use is

```
ACL.SYSTEM.impersonate(new Runnable() {
  public void run() {
    ...
  }
});
```

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/JenkinsAwareGerritHandler.java
```diff
@@ -28,29 +28,44 @@
...
 /**
- * Specialization of GerritHandler that supports gerrit event's lifecycle.
- * @author Hugo Arès &lt;hugo.ares@ericsson.com&gt;
+ * Specialization of GerritHandler that supports gerrit event's
+ * lifecycle and takes care of custom EventThread creation.
  *
+ * @author Hugo Arès &lt;hugo.ares@ericsson.com&gt;
  */
-public class GerritHandlerLifecycle extends GerritHandler {
+public class JenkinsAwareGerritHandler extends GerritHandler implements Coordinator {
```
rsandell: Renaming public classes like this will break backwards compatibility quite severely.

Although a rename is probably in order here the old class should probably be kept but deprecated.

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/LockedDownGerritEventTest.java
```diff
@@ -0,0 +1,147 @@
...
+
+    /**
+     * Lock down the instance.
+     * @throws Exception throw if so.
+     */
+    private void lockDown() throws Exception {
+        GlobalMatrixAuthorizationStrategy auth = new GlobalMatrixAuthorizationStrategy();
+        j.jenkins.setAuthorizationStrategy(auth);
+        j.jenkins.setCrumbIssuer(null);
+        HudsonPrivateSecurityRealm realm = new HudsonPrivateSecurityRealm(false);
```
rsandell: There is a security realm mock in the test harness prepared for these kinds of things that is a bit easier/cleaner to use.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/SystemEventThread.java
rinrinne: @rsandell Seems his code is the same as hudson.model.Executor.

pom.xml
```diff
@@ -61,7 +61,7 @@
         <dependency>
             <groupId>com.sonymobile.tools.gerrit</groupId>
             <artifactId>gerrit-events</artifactId>
-            <version>2.4.2</version>
+            <version>2.5.1</version>
```
rsandell: released as 2.6.0

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/SystemEventThread.java
rsandell: Yes, I don't think its harmful as it's written now. 
@reviewbybees  ?

stephenc: Always safer to restore the old context... though since this is not a thread pool it's more of a stylistic issue... given the amount of copy&paste example taking in Jenkins plugins, I would think it is safer to always to it the "right" way or else give a comment saying why you don't need to

jglick: Agreed with @stephenc, it is safer to restore the former value just in case this code is copied or refactored somehow.

```java
SecurityContext old = ACL.impersonate(ACL.SYSTEM);
try {
    super.run();
} finally {
    SecurityContextHolder.setSecurityContext(old);
}
```

scoheb: Thanks. Will use your snippet Jesse.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/JenkinsAwareGerritHandler.java
hugares: I agree that renaming public class can break backward compatibility but I would be surprised to find one plugin that depends on Gerrit-Trigger and use that class.  For this one, I am not conviced that we need to keep the old one.

hugares: You did not need to add "implements Coordinator", GerritHandler already implement that interface

rsandell: I sincerely doubt that usage too, but if a rename can be avoided I think we should.
But if @scoheb thinks it becomes too messy too keep the original class with my suggested workarounds then go ahead and keep the rename.

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/LockedDownGerritEventTest.java
rsandell: Example of usage:
https://github.com/jenkinsci/build-failure-analyzer-plugin/blob/master/src/test/java/com/sonyericsson/jenkins/plugins/bfa/CauseManagementPermissionTest.java#L40

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/JenkinsAwareGerritHandler.java
scoheb: I kept the rename

rsandell:
Looks good, just waiting to hear from Jenkins and then I'll try a new release.

scoheb:
Seems like the job might be stuck... See https://jenkins.ci.cloudbees.com/job/plugins/job/gerrit-trigger-plugin/414/console

rsandell:
There is a risk that the same issue will appear when onCompleted etc is called with a different security context like with this plugin https://wiki.jenkins-ci.org/display/JENKINS/Authorize+Project+plugin

PR #201, opened 2015-02-06
Honor manual trigger even if it would have been deemed uninteresting

lanfeust69:
Case in point : suppose a new change isn't validated for a transient reason
(Jenkins offline for maintenance, or a problem in the workspace), then a new
patchset is created which doesn't trigger a build, such as an edit of the
commit message, with "Exclude No Code Change" checked in the job configuration,
then there is no way to get a validation : a retrigger of patchset 1 will give
a validation on an obsolete patchset, and a manual trigger of patchset 2
doesn't work either since it is filtered by "Exclude No Code Change".

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/events/PluginPatchsetCreatedEvent.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/events/PluginPatchsetCreatedEvent.java
@@ -27,9 +27,12 @@ import com.sonymobile.tools.gerrit.gerritevents.dto.GerritChangeKind;
 import com.sonymobile.tools.gerrit.gerritevents.dto.events.GerritTriggeredEvent;
 import com.sonymobile.tools.gerrit.gerritevents.dto.events.PatchsetCreated;
 import com.sonyericsson.hudson.plugins.gerrit.trigger.Messages;
+import com.sonyericsson.hudson.plugins.gerrit.trigger.events.ManualPatchsetCreated;
+
 import hudson.Extension;
 import hudson.model.Descriptor;
 import hudson.model.Hudson;
+
 import org.kohsuke.stapler.DataBoundConstructor;
 
 import java.io.Serializable;
@@ -108,6 +111,10 @@ public class PluginPatchsetCreatedEvent extends PluginGerritEvent implements Ser
     public boolean shouldTriggerOn(GerritTriggeredEvent event) {
         if (!super.shouldTriggerOn(event)) {
             return false;
+        }
+        // if it was a manually triggered event, always honor it
+        if (event instanceof ManualPatchsetCreated) {
+            return true;
         }
         if (excludeDrafts && ((PatchsetCreated)event).getPatchSet().isDraft()) {
             return false;


rsandell:
Hmm, that ```BackCompat252HudsonTest``` is flaky as hell, most likely the reason that the build timed out.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/events/PluginPatchsetCreatedEvent.java
```diff
@@ -109,6 +112,10 @@ public boolean shouldTriggerOn(GerritTriggeredEvent event) {
         if (!super.shouldTriggerOn(event)) {
             return false;
         }
+        // if it was a manually triggered event, always honor it
```
rsandell: Should this really be always on, or should it be configurable?

rsandell: I get the point of trivial rebase and no code change, but draft?!

lanfeust69: Of course it's debatable.
My idea was that when you explicitely trigger something, you definitely expect something to happen.
We don't exclude drafts in our settings, and I would suspect that doing so means that you want to have very early drafts that often don't compile. But in this use-case, you probably still want to be able to verify a draft manually.
OTOH, if there is no job watching the corresponding project, you still won't have any build launched...
It may make sense to have this behavior configurable, but then should it be a checkbox such as "Latest Patchset Only" on the manual trigger screen, or a part of the job configuration, or both (and then which one takes precedence) ? I went for simple and (to my eyes at least) reasonable enough, since I really wanted to be able to edit commit messages at (almost) no cost : some of our devs have *a lot* of progress margin there !

rsandell: Well you might have one job that triggers on drafts (hidden from unknowing people) and one that triggers on non drafts.
If you then select a patch set that isn't a draft that would cause both jobs to trigger and that's probably not what you want.

lanfeust69: You can't setup a job to trigger *only* on drafts (at least currently), can you ?
But indeed you could have a "light" job that always triggers, and a "heavier" that runs only on non-drafts, and if someone manually triggers, you can't guess if it's meant to (re)trigger the normal "light" job, or to exceptionally launch the "heavy" one.
Actually, I initially wrote the patch without skipping the draft check, but then thought it was probably better to have *some* way of running a "non-draft" job on a draft (instead of *no* way), even at the cost of an occasional unintended run.
I can of course narrow the feature to "trivial rebases/no code change" if you think it desirable, but I'm afraid I'm not familiar enough with the GUI handling to find the time to have it configurable.

oharboe:
I ran into this too. See my review #216. I'll try to update my review so that it only triggers skipped builds on manual rebuild.

Too many configuration options... Isn't there a way we can avoid configuration options here and just do the right thing always?

rsandell:
If I get the conversation correctly; there currently is no *right* way. Some wants it one way and someone else wants it another. And you are also right in that there is currently too many config options to set and its getting more bloated.

We should at some point do something like what git 2.0 did with added "behaviours" that eased up the config screen a bit. But I don't think that's today :)

rsandell:
Merged #216 instead.

PR #204, opened 2015-02-20
Correct infinite loops caused by CommentAdded

scoheb:
CommentAdded events were not be processed with the correct
gerritEvent() signature. This caused builds be triggered needlessly
and sometimes caused infinite loops between Jenkins and Gerrit.
    
This fix exposes the EventListener as a public class so that the Gerrit Events
GerritHandler can correctly invoke the appropriate methods.
    
This also corrects a NPE during a BuildStarted notification.
    
[JENKINS-23152]


rsandell:
The test errors seems to be related to Java 8

rsandell:
But the Jenkins build was running with JDK 7, so I'm a bit confused.

rsandell:
I found and fixed the test problem, a missed mock that got revealed with the new null check.

scoheb:
Do you want me to do anything?

rsandell:
nope, I think I got it, thanks. I'm pushing to master shortly.

PR #205, opened 2015-03-03
Branch name

thewtex:
Fixes for the BRANCH parameter expansion and adding the TOPIC parameter expansion.


rsandell:
It looks ok, but I would like to hold off merging this untill I've figured out how to handle the current -beta release.

PR #206, opened 2015-03-05
Remove unused test methods

dpursehouse:
Change-Id: I444a68f621cda972557eed49ac5da5ddbdf4524b


src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spec/SpecGerritTriggerHudsonTest.java
```diff
@@ -240,30 +236,6 @@ public void testDynamicTriggeredBuild() throws Exception {
     }
 
     /**
-     * Forces GC by causing an OOM and then verifies the given {@link WeakReference} has been garbage collected.
-     * @param reference object used to verify garbage collection.
-     */
-    @SuppressWarnings("DLS_DEAD_LOCAL_STORE_OF_NULL")
-    // TODO use version from MemoryAssert in 1.519+
```
rsandell: Please move this comment up to the method call above, so we don't miss to look into that TODO

PR #208, opened 2015-03-05
Remove unused logger

dpursehouse:
Change-Id: I9c5e32bf21cd4b6ae3689df0e042d1267fe9bbbc


rsandell:
I think this change should be the other way around; i.e. start using the logger instead.

dpursehouse:
> I think this change should be the other way around; i.e. start using the logger instead.

OK.  Will close this one.  If I have time I'll check where the logger should be used, and do another PR later.

PR #210, opened 2015-03-06
Introduce Google Truth

dpursehouse:
Truth [1] is a testing framework designed to make tests and their error
messages more readable and discoverable, while being extensible to new
types of objects.

Add dependency on truth in the pom file and convert some of the trigger
tests to use it.

This also gets rid of several deprecation warnings which were caused
by using junit.framework.Assert

[1] http://google.github.io/truth/

Change-Id: I6e2ed8e646fa4831a41639ad8a12b5b0a3d98f03

### DIFF
--- a/src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServerHudsonTest.java
+++ b/src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServerHudsonTest.java
@@ -44,10 +44,8 @@ import java.io.IOException;
 import java.net.URL;
 import java.util.List;
 
-
-
-//CS IGNORE AvoidStarImport FOR NEXT 1 LINES. REASON: UnitTest.
-import static org.junit.Assert.*;
+import static com.google.common.truth.Truth.assertThat;
+import static com.google.common.truth.Truth.assert_;
 
 
 /**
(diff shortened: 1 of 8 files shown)


dpursehouse:
In this initial commit I only converted a few of the tests.  If this change is accepted I will convert the others in follow-up commits.

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServerHudsonTest.java
```diff
@@ -150,14 +148,16 @@ public void testTriggeringFromMultipleGerritServers() throws Exception {
         TestUtils.waitForBuilds(projectTwo, 1);
 
         FreeStyleBuild buildOne = projectOne.getLastCompletedBuild();
-        assertSame(Result.SUCCESS, buildOne.getResult());
-        assertEquals(1, projectOne.getLastCompletedBuild().getNumber());
-        assertSame(gerritServerOneName, buildOne.getCause(GerritCause.class).getEvent().getProvider().getName());
+        assertThat(buildOne.getResult()).isEqualTo(Result.SUCCESS);
```
rsandell: same != equalTo

dpursehouse: is `assertSame` actually correct in these cases?

rsandell:
I see no problem with introducing another assertion framework, whatever floats your boat :)

But rewriting tests is problematic as you have to verify that the test still captures the error/failure that they are supposed to catch _(by "rebraking" the code to verify that the test still works)_.

So if you have any new tests you would like to add with this new framework, then go for it. But please leave the old tests as they are :)

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServerHudsonTest.java
rsandell: IIRC ```Result``` isn't an enum, but a set of singeltons, so checking reference equals is more correct than instance equals IMHO.

dpursehouse:
OK.  I will close this one then, as I don't currently have any specific new tests that I want to add.

I'll do another pull request to fix the usage of deprecated Assert.

PR #211, opened 2015-03-06
Update checkstyle config to output results to the console

dpursehouse:
This makes it easier to run checkstyle from the command line with:

  mvn checkstyle:checkstyle

Change-Id: Ia09777576127fd9514b8cc64e1b867c8de826662

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,11 +43,15 @@ fail during the verification phase if you don't follow them.
 
     mvn clean package
     
-Run findbugs for future reference
-or to make sure you haven't introduced any new warnings
+Run findbugs for future reference or to make sure you haven't introduced any
+new warnings
 
     mvn findbugs:findbugs
-    
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -43,11 +43,15 @@ fail during the verification phase if you don't follow them.
...
-or to make sure you haven't introduced any new warnings
+Run findbugs for future reference or to make sure you haven't introduced any
+new warnings
 
     mvn findbugs:findbugs
-    
+
+Run checkstyle
+
+    mvn checkstyle:checkstyle
```
rsandell: ```checkstyle:checkstyle``` is run during the compile phase and ```checkstyle:check``` is run during the verify phase, so there is not need to run it again.

dpursehouse: Right, but running it explicitly means you get the log output without all the noise of the compile and test output.

PR #212, opened 2015-03-06
Build with Java 7

dpursehouse:
Public updates for Java 6 ended in Feb 2013 [1].

Update the project to build with Java 7.

Java 7 has support for type inference, so we no longer need to give
explicit specification of type arguments.  For example code like:

   List<String> list = new List<String>();

will now cause a warning "Redundant specification of type arguments".

Code using this syntax can be changed to:

   List<String> list = new List<>();

which silences the warning and makes the code a bit more compact.

[1] http://www.oracle.com/technetwork/java/eol-135779.html
[2] http://docs.oracle.com/javase/tutorial/java/generics/genTypeInference.html

Change-Id: Ifd72081ae7d6090594cd3d5cb219a639da64424a

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ Each job can be configured with one Gerrit server.
 
 # Environments
 * `linux`
-    * `java-1.6`
+    * `java-1.7`
         * `maven-3.0.4`
 
 You should have no problem running the plugin on a Windows server.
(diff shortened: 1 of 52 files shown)


rsandell:
:-1: The community still wants to keep Java 6 compatibility. https://wiki.jenkins-ci.org/display/JENKINS/Java+5+Compatibility

You can still use jdk 7 to build the plugin. And in later hpi plugin version tests can be written in java 8 if desired _(but IIUC there is not IDE that supports different language versions between tests and production)_

Although It is up to each plugin to decide to keep java 6 or not, I think I'd still prefer it for a while longer.

PR #213, opened 2015-03-12
Missed Events Playback on re-connect

scoheb:
The missed events playback manager is now able to:

- maintain a last known alive timestamp of events that were received by the
Gerrit Server connection.
- Upon re-connect, a request is made to the Gerrrit Events-log plugin installed on the Gerrit
 Server to determine which events may have been missed while the connection was down.
- The events are then added to the Gerrit Trigger event queue to be processed.

This feature is only enabled if:

- The REST api is configured.
- The Gerrit Events-log plugin is installed on the Gerrit Server (See https://gerrit.googlesource.com/plugins/events-log/)

[JENKINS-23871]

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer.java
@@ -111,6 +111,7 @@ import com.sonyericsson.hudson.plugins.gerrit.trigger.hudsontrigger.GerritConnec
 import com.sonyericsson.hudson.plugins.gerrit.trigger.hudsontrigger.UnreviewedPatchesListener;
 import com.sonyericsson.hudson.plugins.gerrit.trigger.hudsontrigger.GerritTrigger;
 import com.sonyericsson.hudson.plugins.gerrit.trigger.hudsontrigger.data.GerritSlave;
+import com.sonyericsson.hudson.plugins.gerrit.trigger.playback.GerritMissedEventsPlaybackManager;
 import com.sonyericsson.hudson.plugins.gerrit.trigger.version.GerritVersionChecker;
 
 /**
@@ -152,6 +153,7 @@ public class GerritServer implements Describable<GerritServer>, Action {
     private transient UnreviewedPatchesListener unreviewedPatchesListener;
     private IGerritHudsonTriggerConfig config;
     private transient GerritConnectionListener gerritConnectionListener;
+    private transient GerritMissedEventsPlaybackManager missedEventsPlaybackManager;
 
     @Override
     public DescriptorImpl getDescriptor() {
@@ -159,6 +161,14 @@ public class GerritServer implements Describable<GerritServer>, Action {
     }
 
     /**
+     * Returns the Missed Events playback manager.
+     * @return GerritMissedEventsPlaybackManager
+     */
+    public GerritMissedEventsPlaybackManager getMissedEventsPlaybackManager() {
+        return missedEventsPlaybackManager;
+    }
+
+     /**
      * Convenience method for jelly to get url of the server list's page relative to root.
      * @link {@link GerritManagement#getUrlName()}.
      *
@@ -391,6 +401,10 @@ public class GerritServer implements Describable<GerritServer>, Action {
         config.setCategories(categories);
         gerritEventManager = PluginImpl.getHandler_();
 
+        if (missedEventsPlaybackManager == null) {
+            missedEventsPlaybackManager = new GerritMissedEventsPlaybackManager(name);
+        }
+
         initializeConnectionListener();
 
         projectListUpdater = new GerritProjectListUpdater(name);
@@ -398,6 +412,10 @@ public class GerritServer implements Describable<GerritServer>, Action {
 
         //Starts unreviewed patches listener
         unreviewedPatchesListener = new UnreviewedPatchesListener(name);
+
+        if (missedEventsPlaybackManager.isSupported()) {
+            addListener((GerritEventListener)missedEventsPlaybackManager);
+        }
 
         logger.info(name + " started");
         started = true;
@@ -433,6 +451,10 @@ public class GerritServer implements Describable<GerritServer>, Action {
         if (unreviewedPatchesListener != null) {
             unreviewedPatchesListener.shutdown();
             unreviewedPatchesListener = null;
+        }
+
+        if (missedEventsPlaybackManager != null) {
+            missedEventsPlaybackManager.shutdown();
         }
 
         if (gerritConnection != null) {
@@ -502,6 +524,10 @@ public class GerritServer implements Describable<GerritServer>, Action {
                 gerritConnection.setHandler(gerritEventManager);
                 gerritConnection.addListener(gerritConnectionListener);
                 gerritConnection.addListener(projectListUpdater);
+                if (missedEventsPlaybackManager == null) {
+                    missedEventsPlaybackManager = new GerritMissedEventsPlaybackManager(name);
+                }
+                gerritConnection.addListener(missedEventsPlaybackManager);
                 gerritConnection.start();
             } else {
                 logger.warn("Already started!");
@@ -517,6 +543,7 @@ public class GerritServer implements Describable<GerritServer>, Action {
         if (gerritConnection != null) {
             gerritConnection.shutdown(true);
             gerritConnection.removeListener(gerritConnectionListener);
+            gerritConnection.removeListener(missedEventsPlaybackManager);
             gerritConnection = null;
             gerritEventManager.setIgnoreEMail(name, null);
         } else {
@@ -1065,6 +1092,19 @@ public class GerritServer implements Describable<GerritServer>, Action {
             if (gerritConnectionListener.isSnapShotGerrit()) {
                 return true;
             }
+        }
+        return false;
+    }
+
+    /**
+     * If Gerrit Missed Events Playback is supported.
+     *
+     * @return true if so, false otherwise.
+     */
+    @JavaScriptMethod
+    public boolean isGerritMissedEventsSupported() {
+        if (gerritConnectionListener.isConnected()) {
+            return missedEventsPlaybackManager.isSupported();
         }
         return false;
     }
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/EventTimeSlice.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/EventTimeSlice.java
@@ -1,0 +1,81 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2014 Ericsson
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package com.sonyericsson.hudson.plugins.gerrit.trigger.playback;
+
+import com.sonymobile.tools.gerrit.gerritevents.dto.events.GerritTriggeredEvent;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * This class holds events that were processed by the MissedEventPlaybackManager
+ * for the most recent timestamp.
+ *
+ * For example, if 3 events were received at t1, they would be present in the Event Slice.
+ * However, if another event is processed at t2, then the Event Slice would evict the previous
+ * events and only keep the new event at t2.
+ *
+ * Created by scott.hebert@ericsson.com on 12/12/14.
+ */
+public class EventTimeSlice {
+
+    private long timeSlice;
+    /**
+     * events to persist.
+     */
+    protected List<GerritTriggeredEvent> events = Collections.synchronizedList(new ArrayList<GerritTriggeredEvent>());
+
+    /**
+     *
+     * @param ts Time slice in ms to hold events for.
+     */
+    public EventTimeSlice(long ts) {
+        this.timeSlice = ts;
+    }
+
+    /**
+     * Get the time slice in ms.
+     * @return this time slice
+     */
+    public long getTimeSlice() {
+        return timeSlice;
+    }
+    /**
+     * Add an event to the list.
+     * @param evt Event to be persisted.
+     */
+    public void addEvent(GerritTriggeredEvent evt) {
+        events.add(evt);
+    }
+
+    /**
+     * get the events for this time slice.
+     * @return events that pertain to the time slice.
+     */
+    public List<GerritTriggeredEvent> getEvents() {
+        return events;
+    }
+}
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsPlaybackManager.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsPlaybackManager.java
@@ -1,0 +1,497 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2014 Ericsson
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package com.sonyericsson.hudson.plugins.gerrit.trigger.playback;
+
+import com.sonyericsson.hudson.plugins.gerrit.trigger.GerritServer;
+import com.sonyericsson.hudson.plugins.gerrit.trigger.PluginImpl;
+import com.sonyericsson.hudson.plugins.gerrit.trigger.config.IGerritHudsonTriggerConfig;
+import com.sonyericsson.hudson.plugins.gerrit.trigger.utils.GerritPluginChecker;
+import com.sonyericsson.hudson.plugins.gerrit.trigger.utils.HttpUtils;
+import com.sonymobile.tools.gerrit.gerritevents.ConnectionListener;
+import com.sonymobile.tools.gerrit.gerritevents.GerritEventListener;
+import com.sonymobile.tools.gerrit.gerritevents.GerritJsonEventFactory;
+import com.sonymobile.tools.gerrit.gerritevents.dto.GerritEvent;
+import com.sonymobile.tools.gerrit.gerritevents.dto.attr.Provider;
+import com.sonymobile.tools.gerrit.gerritevents.dto.events.ChangeBasedEvent;
+import com.sonymobile.tools.gerrit.gerritevents.dto.events.GerritTriggeredEvent;
+
+import hudson.Util;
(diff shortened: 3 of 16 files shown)


hugares:
In the commit, your refer to the Audit plugin, you should change that to Events-log plugin

rinrinne:
In event-log, event visibility check is done per project. But Gerrit can handle access control per branch. So that plugin has the possibility to playback unnecessary events.

In this pluigin, playback process is started just after Gerrit connection is established. It means that events from both are received in the same duration. If "Build Current Patchset Only" feature is enabled, it might not work correctly because event order based on timestamp is broken.

hugares:
@rinrinne You are right, currently, the visibility check is done per project in the events-log plugin but we could easily modify it to do it per branch also.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer.java
```diff
@@ -1066,6 +1093,19 @@ public boolean isGerritSnapshotVersion() {
     }
 
     /**
+     * If Gerrit Missed Events Playback is supported.
+     *
+     * @return true if so, false otherwise.
+     */
+    @JavaScriptMethod
+    public boolean isGerritMissedEventsNotSupported() {
```
rsandell: The javadoc seems to say something different from what the name of the method and its implementation suggests.

And inverse logic like this is usually hard to grok when you read the code, so try to be careful with the naming to not confuse too much.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsPlaybackManager.java
```diff
@@ -0,0 +1,438 @@
...
+    protected void load() throws IOException {
+        XmlFile xml = PlaybackUtils.getConfigXml();
+        if (xml.exists()) {
+            serverTimestamps = (ConcurrentHashMap<String, EventTimeSlice>)xml.unmarshal(serverTimestamps);
+        }
+    }
+
+    /**
+     * get DateRange from current and last known time.
+     * @return diff between current time and last known
```
rsandell: This doesn't seem to produce a diff, only the last or current Date.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsPlaybackManager.java
```diff
@@ -0,0 +1,438 @@
...
+ * some events while the connection was down.
+ *
+ * Once the server is re-connected, the missed event will be played back as if they had been
+ * received orginally.
+ *
+ * @author scott.hebert@ericsson.com
+ */
+public class GerritMissedEventsPlaybackManager implements ConnectionListener, GerritEventListener {
+
+    private static final Logger logger = LoggerFactory.getLogger(GerritMissedEventsPlaybackManager.class);
```
rsandell: I think all the log levels in this class needs to be lowered.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsPlaybackManager.java
```diff
@@ -0,0 +1,438 @@
...
+     */
+    private Vector<ChangeBasedEvent> createEventsFromString(String eventsString) {
+        Vector<ChangeBasedEvent> events = new Vector<ChangeBasedEvent>();
+        Scanner scanner = new Scanner(eventsString);
+        while (scanner.hasNextLine()) {
+            String line = scanner.nextLine();
+            logger.debug("found line: {}", line);
+            JSONObject jsonObject = JSONObject.fromObject(line);
+            GerritEvent evt = GerritJsonEventFactory.getEvent(jsonObject);
+            if (evt instanceof ChangeBasedEvent) {
```
rsandell: so why only ```ChangeBasedEvent```s? why not ref-updated as well?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsPlaybackManager.java
```diff
@@ -0,0 +1,438 @@
...
+
+    private String serverName;
+    /**
+     * Map that contains timestamps.
+     */
+    protected Map<String, EventTimeSlice> serverTimestamps = new ConcurrentHashMap<String, EventTimeSlice>();
+    /**
+     * Vector that contains received Gerrit Events.
+     */
+    protected Vector<ChangeBasedEvent> receivedEventCache = new Vector<ChangeBasedEvent>();
```
rsandell: I would prefer you use ```List l = Collections.synchronizedList(new SomeListImplementation())``` as it doesn't expose any specific implementation as well as clearly state synchronization intent instead of giving Java 1.1 flashbacks.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsPlaybackManager.java
```diff
@@ -0,0 +1,438 @@
...
+            return;
+        }
+
+        if (event instanceof GerritTriggeredEvent) {
+            logger.info("Recording timestamp due to an event {} for server: {}", event, serverName);
+            GerritTriggeredEvent triggeredEvent = (GerritTriggeredEvent)event;
+            persist(triggeredEvent);
+            //add to cache
+            if (!playBackComplete) {
+                if (!receivedEventCache.contains(event)) {
```
rsandell: Using a ```synchronizedSet``` here instead could maybe lower the complexity.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsPlaybackManager.java
```diff
@@ -0,0 +1,438 @@
...
+                    receivedEventCache.add((ChangeBasedEvent)event);
+                    logger.debug("Added event {} to received cache for server: {}", event, serverName);
+                } else {
+                    logger.debug("Event {} ALREADY in received cache for server: {}", event, serverName);
+                }
+            } else {
+                receivedEventCache = new Vector<ChangeBasedEvent>();
```
rsandell: This will create a new object for every event coming in during normal operations, it seems a bit unnecessary exercising of the heap.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsPlaybackManager.java
```diff
@@ -0,0 +1,438 @@
...
+
+    /**
+     * Get events for a given lower bound date.
+     * @param lowerDate lower bound for which to request missed events.
+     * @return collection of gerrit events.
+     * @throws IOException if HTTP errors occur
+     */
+    protected Vector<ChangeBasedEvent> getEventsFromDateRange(Date lowerDate) throws IOException {
+
+        GerritServer server = PluginImpl.getInstance().getServer(serverName);
```
rsandell: Findbugs will have an issue with this in the current core version we are depending on as ```PluginImpl.getInstance()``` is annotated with ```CheckForNull```.

rsandell: Use ```PluginImpl._getServer(...)``` as a simple shortcut to prevent that, but check ```server``` for ```null```.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsPlaybackManager.java
```diff
@@ -0,0 +1,438 @@
...
+            return "";
+        }
+
+        int statusCode = execute.getStatusLine().getStatusCode();
+        logger.info("Received status code: {} for server: {}", statusCode, serverName);
+
+        if (statusCode == HttpURLConnection.HTTP_OK) {
+            try {
+                InputStream bodyStream = execute.getEntity().getContent();
+                String body = IOUtils.toString(bodyStream, "UTF-8");
```
rsandell: I don't think you can be sure that the server returned the content in UTF-8, use the content type from the response instead.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsPlaybackManager.java
```diff
@@ -0,0 +1,438 @@
...
+            return false;
+        }
+        return true;
+    }
+
+    /**
+     * Shutdown the listener.
+     */
+    public void shutdown() {
+        GerritServer server = PluginImpl.getInstance().getServer(serverName);
```
rsandell: ```CheckForNull```

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsPlaybackManager.java
```diff
@@ -0,0 +1,438 @@
...
+
+    private static final Logger logger = LoggerFactory.getLogger(GerritMissedEventsPlaybackManager.class);
+    static final String EVENTS_LOG_PLUGIN_NAME = "events-log";
+    private static final String EVENTS_LOG_PLUGIN_URL = "a/plugins/" + EVENTS_LOG_PLUGIN_NAME + "/events/";
+
+    private String serverName;
+    /**
+     * Map that contains timestamps.
+     */
+    protected Map<String, EventTimeSlice> serverTimestamps = new ConcurrentHashMap<String, EventTimeSlice>();
```
rsandell: From what I can see there is one ```GerritMissedEventsPlaybackManager``` per ```GerritServer``` and you are only using serverName as a key for putting and reading values. So it seems like only one value will be ever stored in this map?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/PlaybackUtils.java
```diff
@@ -0,0 +1,53 @@
...
+     */
+    private PlaybackUtils() {
+    }
+
+    /**
+     * @return XmlFile corresponding to gerrit-trigger-server-timestamps.xml.
+     */
+    public static XmlFile getConfigXml() {
+        return new XmlFile(Jenkins.XSTREAM,
+                new File(Jenkins.getInstance().getRootDir(), "gerrit-trigger-server-timestamps.xml"));
```
rsandell: CheckForNull

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/PlaybackUtils.java
```diff
@@ -0,0 +1,53 @@
...
+    /**
+     * Playback utils.
+     */
+    private PlaybackUtils() {
+    }
+
+    /**
+     * @return XmlFile corresponding to gerrit-trigger-server-timestamps.xml.
+     */
+    public static XmlFile getConfigXml() {
```
rsandell: Since this is the only method it could be moved into ```GerritMissedEventsPlaybackManager```.
But now I get why you thought of using a map, but you are risking each server instance overwriting this file so other servers will loose their data.

A directory with one file per server would be more safe I think.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/PlaybackUtils.java
rsandell: the file name should be a constant.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/utils/GerritPluginChecker.java
```diff
@@ -0,0 +1,146 @@
...
+     */
+    private static boolean decodeStatus(int statusCode, String pluginName) {
+        switch (statusCode) {
+            case HttpURLConnection.HTTP_OK:
+                logger.info("Gerrit Plugin " + pluginName + " is installed");
+                return true;
+            case HttpURLConnection.HTTP_NOT_FOUND:
+                logger.info("Gerrit Plugin " + pluginName + " is NOT installed");
+                return false;
+            case HttpURLConnection.HTTP_UNAUTHORIZED:
```
rsandell: you could simplify these by parameterized messages. ```Messages.PluginHttpConnectionUnauthorized(pluginName)```

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/utils/GerritPluginChecker.java
```diff
@@ -0,0 +1,146 @@
...
+        if (restUrl == null) {
+            logger.warn(NOT_ABLE_TO_VERIFY_IF_GERRIT_PLUGIN + pluginName
+                    + " is installed. REST URL cannot be null. Is the REST API Enabled?");
+            return false;
+        }
+        logger.trace("{}plugins/{}/", restUrl, pluginName);
+
+        DefaultHttpClient httpclient = new DefaultHttpClient();
+        HttpGet httpGet = new HttpGet(restUrl + "plugins/" + pluginName + "/");
+        httpclient.getCredentialsProvider().setCredentials(new AuthScope(null, -1),
```
rsandell: You forgot to set the proxy.

I count now three places where a HttpClient is created in a similar manner with most likely cut & paste code.
A utility method in the config would help this a lot I think.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsPlaybackManager.java
```diff
@@ -0,0 +1,438 @@
...
+    public GerritMissedEventsPlaybackManager(String name) {
+        this.serverName = name;
+        checkIfEventsLogPluginSupported();
+    }
+
+    /**
+     * method to verify if plugin is supported
+     */
+    private void checkIfEventsLogPluginSupported() {
+        GerritServer server = PluginImpl.getInstance().getServer(serverName);
```
rsandell: CheckForNull

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsPlaybackManager.java
```diff
@@ -0,0 +1,438 @@
...
+     * get DateRange from current and last known time.
+     * @return diff between current time and last known
+     * timestamp.
+     */
+    protected synchronized Date getDateFromTimestamp() {
+        //get timestamp for server
+        if (serverTimestamps != null) {
+            EventTimeSlice myTimeStamp = serverTimestamps.get(serverName);
+            if (myTimeStamp != null) {
+                Date myDate = new Date(myTimeStamp.getTimeSlice());
```
rsandell: ```getTimeSlice``` is claming to use seconds, but ```Date``` is using ```ms``` right?

scoheb:
Hi @rsandell,

I think I addressed all your comments.

Please review at your convenience

Thanks,

Scott

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/EventTimeSlice.java
```diff
@@ -0,0 +1,76 @@
...
+ * For example, if 3 events were received at t1, they would be present in the Event Slice.
+ * However, if another event is processed at t2, then the Event Slice would evict the previous
+ * events and only keep the new event at t2.
+ *
+ * Created by scott.hebert@ericsson.com on 12/12/14.
+ */
+public class EventTimeSlice {
+
+    private long timeSlice;
+    private Vector<GerritTriggeredEvent> events = new Vector<GerritTriggeredEvent>();
```
rsandell: Use a synchronized List implementation instead

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsPlaybackManager.java
```diff
@@ -0,0 +1,486 @@
...
+        try {
+            load();
+        } catch (IOException e) {
+            logger.error("Failed to load in timestamps for server {}", serverName);
+            logger.error("Exception: {}", e.getMessage(), e);
+            playBackComplete = true;
+            return;
+        }
+        Date timeStampDate = getDateFromTimestamp();
+        long diff = new Date().getTime() - timeStampDate.getTime();
```
rsandell: ```System.currentTimeMillis()```

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsPlaybackManager.java
```diff
@@ -0,0 +1,486 @@
...
+        if (diff > 0) {
+            logger.debug("Non-zero date range from last-alive timestamp exists for server " + serverName
```
rsandell:
```
logger.debug("Non-zero date range from last-alive timestamp exists for server {}:{}", serverName, Util.getPastTimeString(diff));
```

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsPlaybackManager.java
```diff
@@ -0,0 +1,486 @@
...
+                    + ": " + Util.getPastTimeString(diff));
+        } else {
+            logger.debug("Zero date range from last-alive timestamp for server " + serverName);
```
rsandell: logger.debug("Zero date range from last-alive timestamp for server {}", serverName);

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsPlaybackManager.java
```diff
@@ -0,0 +1,486 @@
...
+            playBackComplete = true;
+            return;
+        }
+        try {
+            List<GerritTriggeredEvent> events = getEventsFromDateRange(timeStampDate);
+            logger.info("({}) missed events to process for server: {} ...", events.size(), serverName);
+            for (GerritTriggeredEvent evt: events) {
+                logger.info("({}) Processing missed event {}", serverName, evt);
```
rsandell: I guess this should be debug level as well?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsPlaybackManager.java
```diff
@@ -0,0 +1,486 @@
...
+
+        if (ts == 0) {
+            logger.warn("Event CreatedOn is 0...Gerrit Server does not support attribute eventCreateOn. "
+                    + "Will NOT persist this event and Missed Events will be disabled!");
+            isSupported = false;
+            return false;
+        }
+
+        if (serverTimestamp != null && ts < serverTimestamp.getTimeSlice()) {
+            logger.debug("Event has same time slice {} or is earlier...NOT Updating time slice.", ts);
```
rsandell: so maybe you should return here instead of persisting to disk?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsPlaybackManager.java
```diff
@@ -0,0 +1,486 @@
...
+    public EventTimeSlice getServerTimestamp() {
+        return serverTimestamp;
+    }
+
+    /**
+     * @param serverName The Name of the Gerrit Server to load config for.
+     * @return XmlFile corresponding to gerrit-trigger-server-timestamps.xml.
+     */
+    public static XmlFile getConfigXml(String serverName) {
+        File xmlFile = new File(Jenkins.getInstance().getRootDir() + GERRIT_SERVER_EVENT_DATA_FOLDER
```
rsandell: CheckForNull on ```Jenkins.getInstance()```

Try using the ```new File(parent, child)``` construct instead of string concatenation. Or ```jenkins.getRootPath().child(x).child(y)```.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsPlaybackManager.java
```diff
@@ -0,0 +1,486 @@
...
+                    logger.info("({}) Triggering: {}", serverName, evt);
+                    GerritServer server = PluginImpl.getServer_(serverName);
+                    if (server == null) {
+                        logger.error("Server for {} could not be found. Skipping this event", serverName);
+                        continue;
+                    }
+                    server.triggerEvent(evt);
+                    receivedEventCache.add(evt);
+                    logger.debug("Added event {} to received cache for server: {}", evt, serverName);
+                    persist(evt);
```
rsandell: Is there some way you can avoid writing to disk for every single event? If Jenkins has been down for 24h there could be thousands of events, and that would take a really long time to process if each event should also be persisted to disk before the next one can be processed.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsPlaybackManager.java
scoheb: I would argue maybe only the one inside the loop...the main one could be seen as valuable upon server startup.

rsandell: Yes, the outer one would only turn up once per server and is a good indication, the inner one though has a risk of polluting the log.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsPlaybackManager.java
scoheb: We need some way of knowing what timestamp to use to request missed events from Gerrit when the jenkins instance comes back up. At some point, we need to persist the timestamp.

I can look into something like MapDB that maybe more performant.

Do you have any ideas? I'll gladly check them out.

rsandell: Well, on line 253 there is also a call to ```persist(triggeredEvent);``` So I'm thinking that its not even needed here. IIRC ```server.triggerEvent(evt);``` will put the event on the event queue so the persist on line 253 is in a different thread and shouldn't slow us down as much.

scoheb: ok, I'll re-run my load tests with line 219 commented out.

Thanks!

scoheb:
I removed the persist() call and re-ran the stress tests. They ran fine.

Thanks @rsandell

rsandell:
Looks good, just fix the compilation error in the test and I think we are good to merge.

scoheb:
Sorry about that @rsandell ...it has been fixed.

rsandell:
W00t!

scoheb:
Now the question is when will this be in a release? :)

rsandell:
My plan is to get #227 merged and then do a beta release, let that stew for a week and hopefully get some beta testing on it. And then release 2.14.

scoheb:
I'd also like to remove the "unreviewed patches listener" since this PR is a replacement for it.

PR #214, opened 2015-03-19
JENKINS-27438: fix NPE when gerritProjects is null

ydubreuil:
This NPE is triggered when saving a job where user deleted all Gerrit
Project filters and checked 'Check Non-Reviewed Patchsets'. The job is
saved anyway.

On Jenkins start, it's also triggered when job is loaded, preventing
such jobs to be loaded.

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
@@ -426,9 +426,10 @@ public class GerritTrigger extends Trigger<AbstractProject> {
 
         GerritProjectList.removeTriggerFromProjectList(this);
         if (allowTriggeringUnreviewedPatches) {
-            for (GerritProject p : gerritProjects) {
-                GerritProjectList.addProject(p, this);
-            }
+            if (gerritProjects != null)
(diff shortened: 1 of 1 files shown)


jglick:
Looks good insofar as it goes, but there are still two missing pieces.

First, `AbstractProject.onLoad` needs to be patched to catch and report this kind of plugin bug. The fix of [JENKINS-22811](https://issues.jenkins-ci.org/browse/JENKINS-22811) shows the idea.

Second, there is the general problem (just confirmed by experimentation in `ui-samples-plugin`) that `<f:repeatableProperty>` (or the lower-level `<f:repeatable>` used in this plugin, probably unwisely), `<f:repeatableHeteroProperty>`, and perhaps others like `<f:hetero-list>` will be data-bound to `null` rather than an empty list if there are no entries, which is an antipattern prone to producing this kind of NPE. I suspect the problem lies in `org.kohsuke.stapler.RequestImpl.bindJSON`: the form has no entries, so the `buildFormTree` handler in JavaScript omits any mention of the field in the submitted JSON, yet a `@DataBoundConstructor` is called with `null` for a `List`-typed parameter.

ydubreuil:
Thanks Jesse, will work on fixing the two other problems.

jglick:
https://github.com/jenkinsci/jenkins/pull/1616 for the first.

PR #215, opened 2015-03-19
Fix for "Build Current Patches Only"

TWestling:
The feature "Build Current Patches Only" cancels
builds that are already running for the same
change. I've added checks for if one of the triggers
was manual trigger as well as if an older patch set
gets triggered (due to a retrigger),
where the builds won't get cancelled.

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/config/Config.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/config/Config.java
@@ -24,6 +24,7 @@
 package com.sonyericsson.hudson.plugins.gerrit.trigger.config;
 
 import com.google.common.primitives.Ints;
+import com.sonyericsson.hudson.plugins.gerrit.trigger.hudsontrigger.data.BuildCancellationPolicy;
 import com.sonymobile.tools.gerrit.gerritevents.GerritDefaultValues;
 import com.sonymobile.tools.gerrit.gerritevents.dto.attr.Provider;
 import com.sonymobile.tools.gerrit.gerritevents.dto.events.ChangeBasedEvent;
@@ -135,16 +136,6 @@ public class Config implements IGerritHudsonTriggerConfig {
      */
     public static final Notify DEFAULT_NOTIFICATION_LEVEL = Notify.ALL;
 
-    /**
-     * Default value for {@link #isGerritAbortNewPatchsets()}.
-     */
-    public static final boolean DEFAULT_ABORT_NEW_PATCHSETS = true;
-
-    /**
-     * Default value for {@link #isGerritAbortManualPatchsets()}.
-     */
-    private static final boolean DEFAULT_ABORT_MANUAL_PATCHSETS = true;
-
     private String gerritHostName;
     private int gerritSshPort;
     private String gerritProxy;
@@ -157,9 +148,8 @@ public class Config implements IGerritHudsonTriggerConfig {
     private Secret gerritHttpPassword;
     private boolean restCodeReview;
     private boolean restVerified;
-    private boolean gerritBuildCurrentPatchesOnly;
-    private boolean gerritAbortNewPatchsets;
-    private boolean gerritAbortManualPatchsets;
+    @Deprecated
+    private transient boolean gerritBuildCurrentPatchesOnly;
     @Deprecated
     private transient int numberOfWorkerThreads;
     private String gerritVerifiedCmdBuildSuccessful;
@@ -189,6 +179,7 @@ public class Config implements IGerritHudsonTriggerConfig {
     private int watchdogTimeoutMinutes;
     private WatchTimeExceptionData watchTimeExceptionData;
     private Notify notificationLevel;
+    private BuildCancellationPolicy buildCurrentPatchesOnly;
 
 
     /**
@@ -220,8 +211,6 @@ public class Config implements IGerritHudsonTriggerConfig {
         restCodeReview = config.isRestCodeReview();
         restVerified = config.isRestVerified();
         gerritBuildCurrentPatchesOnly = config.isGerritBuildCurrentPatchesOnly();
-        gerritAbortNewPatchsets = config.isGerritAbortNewPatchsets();
-        gerritAbortManualPatchsets = config.isGerritAbortManualPatchsets();
         numberOfWorkerThreads = config.getNumberOfReceivingWorkerThreads();
         numberOfSendingWorkerThreads = config.getNumberOfSendingWorkerThreads();
         gerritBuildStartedVerifiedValue = config.getGerritBuildStartedVerifiedValue();
@@ -276,18 +265,12 @@ public class Config implements IGerritHudsonTriggerConfig {
                 "gerritAuthKeyFilePassword",
                 DEFAULT_GERRIT_AUTH_KEY_FILE_PASSWORD));
 
-        gerritBuildCurrentPatchesOnly = formData.optBoolean(
-                "gerritBuildCurrentPatchesOnly",
-                DEFAULT_BUILD_CURRENT_PATCHES_ONLY);
-
-        gerritAbortNewPatchsets = formData.optBoolean(
-                "gerritAbortNewPatchsets",
-                DEFAULT_ABORT_NEW_PATCHSETS);
-
-        gerritAbortManualPatchsets = formData.optBoolean(
-                "gerritAbortManualPatchsets",
-                DEFAULT_ABORT_MANUAL_PATCHSETS);
-
+        if (formData.has("buildCurrentPatchesOnly")) {
+            JSONObject currentPatchesOnly = formData.getJSONObject("buildCurrentPatchesOnly");
+            buildCurrentPatchesOnly = BuildCancellationPolicy.createPolicyFromJSON(currentPatchesOnly);
+        } else {
+            buildCurrentPatchesOnly = new BuildCancellationPolicy();
+        }
 
         numberOfWorkerThreads = formData.optInt(
                 "numberOfReceivingWorkerThreads",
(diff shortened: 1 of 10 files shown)


scoheb:
Hi,

I believe both of the new conditions that you have introduced should be added a configurable options. We have internal customers how would like to retain the current behavior.

Thanks

Scott

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -1846,15 +1847,22 @@ public String getHelpFile() {
...
                 // Find all entries in runningJobs with the same Change #.
+                // Ignore all manual patchsets and don't cancel builds due to
+                // a retrigger of an older build.
                 if (pairs.getKey() instanceof ChangeBasedEvent) {
-                    if (((ChangeBasedEvent)pairs.getKey()).getChange().equals(event.getChange())) {
+                    ChangeBasedEvent runningChangeBasedEvent = ((ChangeBasedEvent)pairs.getKey());
+                    if (!(runningChangeBasedEvent instanceof ManualPatchsetCreated)
+                            && runningChangeBasedEvent.getChange().equals(event.getChange())
+                            && Integer.parseInt(runningChangeBasedEvent.getPatchSet().getNumber())
+                            < Integer.parseInt(event.getPatchSet().getNumber())) {
```
rsandell: having ```runningChangeBasedEvent.getPatchSet().compareTo(event.getPatchSet()) < 0``` or something along those lines would have been nicer.

rsandell:
I was debating this *(at least with myself)* when the new behaviour change got introduced in a pull request. But couldn't motivate it good enough in my head. But being able to configure weather to ignore retriggers and manual triggered seems like a good thing since different people prefer different behaviours.

TWestling:
I will upload a new change with configuration for this (as well as improve the comparison between new and old patchsets).

oharboe:
We're currently struggling with the manual rebuild, because it doesn't retrigger a build for trivial rebase patchsets.

To trigger builds for trivial rebase patchsets by default swamps our slaves with unnecessary work, because developers frequently have a string of reviews where they only work on the latest review, but they need to rebase all the reviews on top of the newest server branch.

Skipping rebuilds on trivial rebase is problematic when there is a failure on the server branch that interacts with a review. If there is a false positive or negative, then the developer *does* want a rebuild even if the rebase was trivial.

#216 makes it easy for developers to manually request a build in this case because #216 will always trigger a build on manual rebuild, even if the rebase was a trivial rebase. I don't want to have a configuration or GUI option here if we can avoid it. It seems like clutter.

I mention this because this pull-request is somewhat related to #216.

Thoughts?

xbeta:
I don't even understand why we can't override this behavior in manual trigger. It's manual so that means we know we want to force it to build it no matter of what.   Am I missing something here?  I would like to join this discussion as to why this behavior has changed.

oharboe:
What changed is that a new feature has been added: that builds can be
skipped for trivial rebase, comment only and draft reviews.

IMO, manual trigger should continue to do what it has always done, even
though the new skip build feature has been added: always trigger a build.

There *are* ways around it using new voting labels, topics, labels and
detailed instructions to developers, but they already know how to use
manual build and I don't particularly want to make things any more
complicated than necessary.


On Thu, Mar 26, 2015 at 8:41 PM, Sam Xiao <notifications@github.com> wrote:

> I don't even understand why we can't override this behavior in manual
> trigger. It's manual so that means we know we want to force it to build it
> no matter of what. Am I missing something here? I would like to join this
> discussion as to why this behavior has changed.
>
> —
> <https://github.com/jenkinsci/gerrit-trigger-plugin/pull/215#issuecomment-86687139>
> .
>



-- 
http://www.zylin.com/

xbeta:
+1 with @oharboe

oharboe:
Look at #216, it's a focused pullrequest to fix the problem with manual
retrigger not retriggering builds.

#215 handles other aspects of manual retrigger and canceling builds that
are already started.

On Fri, Mar 27, 2015 at 6:07 AM, Sam Xiao <notifications@github.com> wrote:

> +1 with @oharboe <https://github.com/oharboe>
>
> —
> <https://github.com/jenkinsci/gerrit-trigger-plugin/pull/215#issuecomment-86826542>
> .
>



-- 
http://www.zylin.com/

TWestling:
This doesn't change whether a manual trigger will build, only if it will cancel other running builds( and be cancelled when another patch set is triggered ).

Regarding having this as an option, I agree that it feels like clutter( the reason I didn't make it optional in the first place ) but I didn't know that people wanted the behavior the way it was ( old patch sets cancelling new ones and manual patch sets cancelling anything ). 

I feel like it's safer to add an option under advanced than to force the behavior.

TWestling:
I can't reproduce the failed test locally.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/config/Config.java
```diff
@@ -149,6 +158,8 @@
     private boolean restCodeReview;
     private boolean restVerified;
     private boolean gerritBuildCurrentPatchesOnly;
+    private boolean gerritAbortNewPatchsets;
```
rsandell: When upgrading from a previous version these fields will become ```false```, is that the intended behaviour?

src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer/index.jelly
```diff
@@ -264,6 +264,18 @@
                                 </table>
                             </f:repeatable>
                         </f:entry>
+                        <f:entry title="${%Abort new patch sets}"
```
rsandell: ```Build Current Patches Only``` should be changed to an optionalBlock with these values hidden behind.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/config/Config.java
```diff
@@ -149,6 +158,8 @@
     private boolean restCodeReview;
     private boolean restVerified;
     private boolean gerritBuildCurrentPatchesOnly;
```
rsandell: A slightly better approach, to slightly lesser the amount of fields in this class would be to deprecate ```gerritBuildCurrentPatchesOnly```, make it transient and replace it with an object containing ```AbortNewPatchsets``` and ```AbortManualPatchsets```. It would also make the optionalBlock _(commented below)_ simpler to implement I believe.

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spec/SpecGerritTriggerHudsonTest.java
```diff
@@ -465,10 +466,13 @@ public void testTripleTriggeredBuildWithProjects() throws Exception {
      *
      * @throws Exception if so.
      */
+    @Test
     @LocalData
     public void testBuildLatestPatchsetOnly() throws Exception {
         GerritServer gerritServer = PluginImpl.getInstance().getServer(PluginImpl.DEFAULT_SERVER_NAME);
         ((Config)gerritServer.getConfig()).setGerritBuildCurrentPatchesOnly(true);
+        ((Config)gerritServer.getConfig()).setGerritAbortManualPatchsets(false);
```
rsandell: there are four different states to test here;
* both ```false```
* both ```true```
* one ```false``` and the other ```true```

rsandell:
Since people seem to expect different behaviours from the cancelling of builds I tend to lean on the _"provide options with sensible defaults"_ side of the fence. So that makes me unsure on how to handle #216.
While this pr seems like a sensible approach.

oharboe:
Can we separate concerns to make progress here?

I think #216 is ready to be merged as-is. Nobody will be affected by #216 unless they take explicit user-action and #216 implements the default expected behaviour: always trigger a build upon a manually triggered build.

Personally, I think we subtract from the gerrit-trigger plugin by adding options for #216, but if someone really needs it, then I think it can be addressed afterwards as a separate concern.

src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer/index.jelly
TWestling: I tried to keep this change small but I agree, it is the better way to solve this. I'll change it.

TWestling:
ping @rsandell do you want to take another look?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -1846,15 +1848,26 @@ public String getHelpFile() {
          */
         public synchronized void scheduled(ChangeBasedEvent event, ParametersAction parameters, String projectName) {
             IGerritHudsonTriggerConfig serverConfig = getServerConfig(event);
-            if (serverConfig != null && !serverConfig.isGerritBuildCurrentPatchesOnly()) {
+            BuildCancellationPolicy buildCurrentPatchesOnly = serverConfig.getBuildCurrentPatchesOnly();
```
rsandell: possible NPE when serverConfig == null

rsandell:
@TWestling Just one, Sorry for the delay, I don't think you need me to merge once fixed ;)

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
TWestling: Good catch, I'll add the null check again.

TWestling:
I did the null check but also made sure that the event is put into runningJobs even if no serverConfig exists. Otherwise, if serverConfig is changed to not be null anymore while the build is running, the build wouldn't be able to be aborted.

scoheb:
Hi,

This PR causes builds to be aborted for non-related changes.

I have fixed this in the following PR: https://github.com/jenkinsci/gerrit-trigger-plugin/pull/230

Thanks

Scott

PR #216, opened 2015-03-20
manual trigger: always trigger upon manually triggered build

oharboe:
When skip on trivial rebase is enabled, it is useful to manually
retrigger build of patchsets that failed due to an unrelated failure
on the branch that the review is against.


rsandell:
This is a behaviour change in a similar manner to what we are discussing in #215 and it would be good to have a debate on if this should be a forced behaviour or not.

PR #217, opened 2015-03-20
Abortbuilds

oharboe:
new patchsets: allows users to choose if currently running builds shuld be aborted


src/main/webapp/help-GerritAbortCurrentBuilds.html
```diff
@@ -0,0 +1,4 @@
+<p>Sets <b>abort current builds</b> mode.</p>
+
+<p>If this is enabled, all currently running builds for previous patchsets
```
rsandell: It should be a sub-setting to the cancel setting, it is quite hard to grok what it does as is. But from what I can see it is a subset of that setting. And the help text should explain that as well.

oharboe: I agree that a sub-setting would be good here, but I don't know currently
know how to implement that.

Ideally I'd like there to be no configuration options or user-settings here.

The only reason why I want this option, is that there is a bug Jenkins in
how it aborts builds.

See: https://issues.jenkins-ci.org/browse/JENKINS-17116


Perhaps we should just make it an environment variable?

-Dworkaround17716=true

?



On Fri, Mar 20, 2015 at 3:40 PM, Robert Sandell <notifications@github.com>
wrote:

> In src/main/webapp/help-GerritAbortCurrentBuilds.html
> <https://github.com/jenkinsci/gerrit-trigger-plugin/pull/217#discussion_r26844128>
> :
>
> > @@ -0,0 +1,4 @@
> > +<p>Sets <b>abort current builds</b> mode.</p>
> > +
> > +<p>If this is enabled, all currently running builds for previous patchsets
>
> It should be a sub-setting to the cancel setting, it is quite hard to grok
> what it does as is. But from what I can see it is a subset of that setting.
> And the help text should explain that as well.
>
> —
> <https://github.com/jenkinsci/gerrit-trigger-plugin/pull/217/files#r26844128>
> .
>



-- 
http://www.zylin.com/

oharboe:
#224 is superior to this patch...

PR #218, opened 2015-03-27
[FIXED JENKINS-27651] handle job rename

ndeloof:
see [https://issues.jenkins-ci.org/browse/JENKINS-27651](https://issues.jenkins-ci.org/browse/JENKINS-27651)
@reviewbybees

PR #219, opened 2015-03-27
checkstyle: fixes checkstyle error that snuck into master branch

oharboe:
A tiny checkstyle error made it into master so now all pull requests will fail builds before this is fixed.

BTW, I can't see the artifacts published anywhere on 'details' of a failed build. This makes it hard to figure out what the checkstyle error was.


oharboe:
If #216 has a false positive in the validation, then this review should show that.

This is the test that failed on #216. com.sonyericsson.hudson.plugins.gerrit.trigger.LockedDownGerritEventTest.testTriggerWithLockedDownInstance(LockedDownGerritEventTest.java:148).

rsandell:
#218 contains this fix as well. But this is small and self contained.

oharboe:
Thanks! While i recall: I believe master is unstable currently though. I
dont have the details now but try creating an empty review on master. I
think it will fail ca one in three.

PR #220, opened 2015-03-30
Support configuration of when and how the project list is fetched

GLundh:
Introduces a new configuration section where the admin can configure
how often the project list is queried from Gerrit (used for project
auto completion) and if it should be fetched during startup.

Change-Id: Id4493dc4381d59b9db789e1469c0783bcb2a31f1

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritProjectListUpdater.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritProjectListUpdater.java
@@ -138,7 +138,15 @@ public class GerritProjectListUpdater extends Thread implements ConnectionListen
 
             try {
                 synchronized (this) {
-                    wait(TimeUnit.SECONDS.toMillis(getConfig().getProjectListRefreshInterval()));
+                    long startTime = System.nanoTime();
+                    // Continuously refetching the refresh interval allows us to pick up configuration
+                    // changes on the fly, keeping us from getting stuck on accidentally entered very
+                    // high values.
+                    while (System.nanoTime() - startTime
+                            < TimeUnit.SECONDS.toNanos(getConfig().getProjectListRefreshInterval())
+                            && !shutdown) {
+                        wait(TimeUnit.SECONDS.toMillis(1));
+                    }
                 }
             } catch (InterruptedException ex) {
                 logger.warn("InterruptedException: ", ex);
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritProjectListUpdater.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritProjectListUpdater.java
@@ -33,6 +33,8 @@ import java.io.IOException;
 import java.io.Reader;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.TimeUnit;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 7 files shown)


GLundh:
Tests work locally. Seems like an unstable test case.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritProjectListUpdater.java
```diff
@@ -130,7 +138,7 @@ public void run() {
 
             try {
                 synchronized (this) {
-                    wait(UPDATE_DELAY);
+                    wait(TimeUnit.SECONDS.toMillis(getConfig().getProjectListRefreshInterval()));
```
rsandell: This value could be changed during runtime to ```0``` which should abort it, not cause the thread to never sleep.

How should a change from ```0``` to ```>0``` be handled? I guess admins are used to clicking that reconnect button?

src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer/index.jelly
```diff
@@ -230,6 +230,31 @@
                                         checked="${it.config.enableManualTrigger}"
                                         default="${true}"/>
                         </f:entry>
+                        <f:entry title="${%Fetch Project List Settings}"
+                                 help="/plugin/gerrit-trigger/help-GerritFetchProjectList.html">
```
rsandell: I would prefer we use standard jelly taglib functions as much as possible, since a lot of UI changes are coming. And too much custom html in plugin forms could get broken. Using the taglibs we should get those changes to be compatible "for free".

I think an ```optionalBlock``` here could work https://jenkins-ci.org/maven-site/jenkins-core/jelly-taglib-ref.html#form:optionalBlock

rsandell:
@TWestling I see this as a new feature, so holding off the merge until 2.13 strategy is determined.

rsandell:
The build also failed due to ```There are 1 checkstyle errors.``` but that could be the fixed in #219

GLundh:
Thanks for the comments, Bobby :)

Yes, the checkstyle error are not mine. Those appeared after a rebase. So I'm guessing #219 fixes it.

src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer/index.jelly
GLundh: optionalBlock works great when the checkbox should control the visibility of other elements. But in this case the textbox is _not_ dependent on the checkbox. E.g. I would like both to be visible. So I used the <divs> since I really wanted to align the elements in rows and using a table would not work at all (f:entry + table = broken form submission. I tried everything. It seems to work in hetrolists and repeatables, but not directly under  <f:entries>).

So I'm out of ideas. I see tables are used extensively by the Gerrit-Trigger. While it is not optimal, I'm guessing we can make exceptions when the core libs do not want to "play ball"?

GLundh:
@rsandell: Please take a new look.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritProjectListUpdater.java
```diff
@@ -130,7 +138,15 @@ public void run() {
 
             try {
                 synchronized (this) {
-                    wait(UPDATE_DELAY);
+                    long startTime = System.nanoTime();
```
rsandell: Why ```nanoTime```? Allegedly on certain Windows systems this is a very expensive operation.

src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer/index.jelly
rsandell: ok

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritProjectListUpdater.java
GLundh: I started with currentTimeMillis, but found several articles that recommends nanoTime instead. It should apparently handle machine time adjustments better. Like someone setting a new date/time on the machine.

PR #226, opened 2015-05-07
Fix JENKINS-28161

rsandell:
This just fixes the NPE symptoms, but I'm afraid that
the reason that an event has a null value has some deeper issue somewhere.

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/model/BuildMemory.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/model/BuildMemory.java
@@ -60,11 +60,19 @@ public class BuildMemory {
      *
      * @author James E. Blair &lt;jeblair@hp.com&gt;
      */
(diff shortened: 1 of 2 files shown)


rsandell:
@reviewbybees

jglick:
BTW please include a meaningful summary of the fix in the PR, and link to the JIRA issue from the description.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/model/BuildMemory.java
```diff
@@ -60,11 +60,19 @@
...
+            if (o1 == null && o2 == null) {
+                return 0;
+            }
+            if (o1 != null && o2 == null) {
+                return 1;
+            }
+            if (o1 == null && o2 != null) {
+                return -1;
+            }
+            return new Integer(o1.hashCode()).compareTo(o2.hashCode());
```
stephenc: I know this is an aside from the current issue, but comparing events based on a 32 bit integer in order to 

> This ensures that every event received from Gerrit is kept track of individually

Seems a bit naïve

stephenc:
:+1: on making more stable

PR #227, opened 2015-05-07
Add incremental update of project list

sselberg:
Through project-created event, introduced in Gerrit v2.12.

Testcases updated since this change adds one more default
GerritEventListener.

Change-Id: I22e5f2d622c154f7079c59f23c8fb0a71d45cf10

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/config/Config.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/config/Config.java
@@ -1,7 +1,7 @@
 /*
  *  The MIT License
  *
- *  Copyright 2010, 2015 Sony Mobile Communications Inc. All rights reserved.
+ *  Copyright 2010 Sony Mobile Communications Inc. All rights reserved.
  *
  *  Permission is hereby granted, free of charge, to any person obtaining a copy
  *  of this software and associated documentation files (the "Software"), to deal
@@ -115,11 +115,16 @@ public class Config implements IGerritHudsonTriggerConfig {
     public static final int DEFAULT_GERRIT_BUILD_NOT_BUILT_CODE_REVIEW_VALUE = 0;
 
     /**
-     * Default value indicating if the project list should be fetched on startup.
+     * Default value indicating if the Gerrit server should be used to fetch project names.
      */
-    public static final boolean DEFAULT_LOAD_PROJECT_LIST_ON_STARTUP = true;
+    public static final boolean DEFAULT_ENABLE_PROJECT_AUTO_COMPLETION = true;
     /**
-     * Default value showing how often the project list should be updated.
+     * Default value showing how many seconds between startup and initial project list fetch.
+     */
+    public static final int DEFAULT_PROJECT_LIST_FETCH_DELAY = 0;
+    /**
+     * Default value showing how many seconds between project list fetches.
+     * <p>Only used for Gerrit servers with version &lt; 2.12</p>
      */
     public static final int DEFAULT_PROJECT_LIST_REFRESH_INTERVAL = 3600;
     /**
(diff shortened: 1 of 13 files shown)


sselberg:
@rsandell

src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer/index.jelly
```diff
@@ -241,31 +241,48 @@
...
-                                        value="${it.config.projectListRefreshInterval}"
-                                        default="${com.sonyericsson.gerrithudsontrigger.config.Config.DEFAULT_PROJECT_LIST_REFRESH_INTERVAL}"
-                                        checkUrl="'${rootURL}/${serverURL}/projectListRefreshIntervalCheck?value='+escape(this.value)"
-                                        style="max-width: 100px;"/>
-                                </div>
-                            </div>
-                        </f:entry>
+                            </f:entry>
+                        </j:if>
+                        <j:if test="${instance.isProjectCreatedEventsSupported()}">
```
rsandell: I don't think this behaves well when you create a new server config for the first time i.e. when you don't have any connection details and so version is unknown.

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spec/DuplicateGerritListenersPreloadedProjectHudsonTestCase.java
```diff
@@ -355,13 +350,16 @@ XmlPage loadConfigXmlViaHttp(TopLevelItem job) throws IOException, SAXException
...
         Collection<GerritEventListener> gerritEventListeners = getGerritEventListeners();
-        // DependencyQueueTaskDispatcher adds 1 listener
-        // ReplicationQueueTaskDispatcher adds 1 listener
-        assertEquals(number, gerritEventListeners.size());
+        int nbrOfListeners = extra;
+        nbrOfListeners++; // EventListener adds 1 listener
+        nbrOfListeners++; // DependencyQueueTaskDispatcher adds 1 listener
+        nbrOfListeners++; // ReplicationQueueTaskDispatcher adds 1 listener
+        nbrOfListeners++; // GerritProjectListUpdater adds 1 listener
+        assertEquals(nbrOfListeners, gerritEventListeners.size());
```
rsandell: +:100:

sselberg: Does this mean good or bad....?

src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer/index.jelly
sselberg: Now that you mention it, it complains and moans a bit...

What would be the Jenkins-way?
* Surrounding it with a catch tag and just don't display anything until there's a connection.
or
* Catch the blowup in isProjectCreatedEventsSupported() and fall back to the "Default" options?
or
* None of the above...

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spec/DuplicateGerritListenersPreloadedProjectHudsonTestCase.java
rsandell: That means good :) Its alot more cleaner and easier to read than what the test did previously.

src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer/index.jelly
rsandell: If you aren't connected I guess you'd need to show some hybrid of both possible options, so it might be easier to do something simpler and always show both in some way.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/config/Config.java
```diff
@@ -183,15 +183,16 @@
     private int buildScheduleDelay;
     private int dynamicConfigRefreshInterval;
     private boolean loadProjectListOnStartup;
-    private int projectListRefreshInterval;
+    @Deprecated
+    private transient int projectListRefreshInterval;
```
rsandell: I think you should add to the readResolve method a conversion from ```projectListRefreshInterval``` to ```projectListFetchDelay```. Unless this is the new field that @GLundh added in his previous pr? In that case you don't need to deprecate and make transient because it has never been in a released version of the plugin IIRC.

sselberg:
My configuration logic is lacking.
With the present logic you can't load project list on startup with incremental update without setting a delay...
I'm thinking about setting:
delay = 0 => on startup
delay = <int> => after <int> seconds
and replacing loadOnStartUp with disableProjectListFetching.

sselberg:
@rsandell
Changed config values so that it makes sense, IMOHO.

sselberg:
Updated Commit message.

rsandell:
It would be preferred to have some unit tests for this feature as well. They can come in a separate pr.

PR #228, opened 2015-05-12
Remove Unreviewed Patches feature

scoheb:
This is replaced by the Missed Event Playback Manager feature


src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger/config.jelly
```diff
@@ -182,11 +182,6 @@
            <f:checkbox name="silentMode"
                        checked="${it.silentMode}"/>
         </f:entry>
-        <f:entry title="${%Check Non-Reviewed Patchsets}"
-                 field="allowTriggeringUnreviewedPatches"
```
rsandell: You should also deprecate the field and make it transient in GerritTrigger.java

PR #230, opened 2015-05-15
Change should be checked for BuildCurrentPatchsets

scoheb:
If Build Current Patchsets only is enabled, then the Change in question must first be
checked to see if a build should be aborted.

Without this fix, it is possible for a build to be aborted because of a NON-related patchset
from ANOTHER change.

This change also makes the two new options disabled by default to be compatible
with the previous behaviour.

Change-Id: I958a4974756f398d87c7d60fb23cdd1b978a0ea1

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
@@ -1895,22 +1895,34 @@ public class GerritTrigger extends Trigger<AbstractProject> {
                 // a retrigger of an older build.
                 if (pairs.getKey() instanceof ChangeBasedEvent) {
                     ChangeBasedEvent runningChangeBasedEvent = ((ChangeBasedEvent)pairs.getKey());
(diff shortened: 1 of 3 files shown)


oharboe:
I've created pull request #231 to make 1), 2) and 3) above separate commits.

scoheb:
I will continue with this PR and add some tests while refactoring as suggested.

Another commit is forthcoming.

scoheb:
@rsandell ... ready for your review.

rsandell:
So, are we ready for a 2.14-beta or do we need something more first?

scoheb:
I am ready now for a 2.14 beta!

oharboe:
I don't have anything pending that particularly belongs in the 2.14 release.

On Wed, May 20, 2015 at 12:11 PM, Scott Hebert <notifications@github.com>
wrote:

> I am ready now for a 2.14 beta!
>
> —
> <https://github.com/jenkinsci/gerrit-trigger-plugin/pull/230#issuecomment-103834390>
> .
>



-- 
http://www.zylin.com/

scoheb:
Can you provide an estimate when the beta will be ready?

rsandell:
Just released a minute or so ago :)

scoheb:
When does the Wiki page with the Changelog get updated? I want to point my customers to it.

rsandell:
now :)

PR #233, opened 2015-05-18
README: added section for development dependencies

oharboe:


rsandell:
There is already a literate description of this in the Environments section

oharboe:
I see.

I'll see if I can't reword my patch to distinguish between officially
supported environments and development environments that can be used.

The advantage of allowing users to use Java 1.7 is that it increases the
reach of who can do a bit of casual work on the gerrit-trigger-plugin. That
was my thinking anyway.

On Mon, May 18, 2015 at 6:12 PM, Robert Sandell <notifications@github.com>
wrote:

> There is already a literate description of this in the Environments section
>
> —
> <https://github.com/jenkinsci/gerrit-trigger-plugin/pull/233#issuecomment-103115953>
> .
>



-- 
http://www.zylin.com/

rsandell:
Well, java 1.7 should probably be set as the official dev environment with java 1.6 as runtime (standard community guidelines) When the bump to core > 1.60x java 6 support should be removed.

PR #234, opened 2015-05-18
Fix eclipse build errors

oharboe:
Improve interoperability with Eclipse.

For casual patch-work, one might be using Java 8 and it's nice that the gerrit-trigger-plugin builds in Eclipse out of the box, even if testing is done with Java 7 and mvn install on the command line.


rsandell:
Contains the commit from #233

PR #238, opened 2015-05-26
Improve backwards compatibilty of Project Lister

scoheb:
In moving from 2.13 to 2.14, older gerrit-trigger config files fail to load.
This fix enables it to be backwards compatible.

Change-Id: I0abc74e7880091d148d8d4af892d66ea90e6d8c9


scoheb:
will create a new PR to replace this.

PR #239, opened 2015-05-26
Improve backwards compatibilty of Project Lister

scoheb:
In moving from 2.13 to 2.14, older gerrit-trigger config files fail to load.
This fix enables it to be backwards compatible.

[FIXED JENKINS-28583]

Change-Id: I0abc74e7880091d148d8d4af892d66ea90e6d8c9


rsandell:
Darn it, I thought that field came in after 2.13 and that it was safe to remove.

We should probably try to convert the old value to the new settings with sensible defaults, but his will do until I magically get some time to do it ;)

PR #240, opened 2015-07-10
[JENKINS-26010] Workflow compatibility

tfennelly:
[See JENKINS-26010](https://issues.jenkins-ci.org/browse/JENKINS-26010)

__Current status__: Ready for review. No more changes planned.

- [x] `AbstractProject` to `Job` in core code
- [x] Upgrade `rebuild-plugin`
- [x] Upgrade `git-plugin`
- [x] `AbstractBuild` to `Run` in core code
- [x] Fix tests
- [x] Manual tests
- [x] Resolve TODOs
- [x] Build status "Verified" event back to Gerrit Server.
- [x] Backward Compatibility .
- [x] Run acceptance tests
- [x] Create JIRA for adding acceptance tests (Freestyle and Workflow jobs). See [JENKINS-29836](https://issues.jenkins-ci.org/browse/JENKINS-29836).
- [x] Create JIRA for adding code checkout feature i.e. an easy way to trigger checkout of the correct branch/revision of code that triggered the build. see [JENKINS-29838](https://issues.jenkins-ci.org/browse/JENKINS-29838).
- [x] Document how to checkout the correct branch/revision of code that triggered the build. See [wiki docs](https://wiki.jenkins-ci.org/display/JENKINS/Gerrit+Trigger#GerritTrigger-WorkflowJobs).
- [x] Create test plugin to test backward compatibility.

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer.java
@@ -917,8 +917,10 @@ public class GerritServer implements Describable<GerritServer>, Action {
     private void changeSelectedServerInJobs(String oldName) {
         for (Job job : PluginImpl.getConfiguredJobs_(oldName)) {
 
-            // TODO: find a way to handle add/remove of triggers
             if (!(job instanceof AbstractProject)) {
+                logger.info("Unable to modify Gerrit Trigger configurations for job [" + job.getName()
+                        + "] after Gerrit server has been renamed from [" + oldName + "] to [" + name + "]." +
+                        " This feature is only supported for AbstractProject types e.g. Freestyle Jobs.");
                 return;
             }
             AbstractProject project = (AbstractProject)job;
@@ -944,8 +946,9 @@ public class GerritServer implements Describable<GerritServer>, Action {
     private void removeGerritTriggerInJobs() {
         for (Job job : getConfiguredJobs()) {
 
-            // TODO: find a way to handle add/remove of triggers
             if (!(job instanceof AbstractProject)) {
+                logger.info("Unable to remove Gerrit Trigger ffrom job [" + job.getName() + "]. " +
+                        " This feature is only supported for AbstractProject types e.g. Freestyle Jobs.");
                 return;
             }
             AbstractProject project = (AbstractProject)job;
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer.java
@@ -919,8 +919,8 @@ public class GerritServer implements Describable<GerritServer>, Action {
 
             if (!(job instanceof AbstractProject)) {
                 logger.info("Unable to modify Gerrit Trigger configurations for job [" + job.getName()
-                        + "] after Gerrit server has been renamed from [" + oldName + "] to [" + name + "]." +
-                        " This feature is only supported for AbstractProject types e.g. Freestyle Jobs.");
+                        + "] after Gerrit server has been renamed from [" + oldName + "] to [" + name + "]."
+                        + " This feature is only supported for AbstractProject types e.g. Freestyle Jobs.");
                 return;
             }
             AbstractProject project = (AbstractProject)job;
@@ -947,8 +947,8 @@ public class GerritServer implements Describable<GerritServer>, Action {
         for (Job job : getConfiguredJobs()) {
 
             if (!(job instanceof AbstractProject)) {
-                logger.info("Unable to remove Gerrit Trigger ffrom job [" + job.getName() + "]. " +
-                        " This feature is only supported for AbstractProject types e.g. Freestyle Jobs.");
+                logger.info("Unable to remove Gerrit Trigger ffrom job [" + job.getName() + "]. "
+                        + " This feature is only supported for AbstractProject types e.g. Freestyle Jobs.");
                 return;
             }
             AbstractProject project = (AbstractProject)job;
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer.java
@@ -918,7 +918,7 @@ public class GerritServer implements Describable<GerritServer>, Action {
         for (Job job : PluginImpl.getConfiguredJobs_(oldName)) {
 
             if (!(job instanceof AbstractProject)) {
-                logger.info("Unable to modify Gerrit Trigger configurations for job [" + job.getName()
+                logger.warn("Unable to modify Gerrit Trigger configurations for job [" + job.getName()
                         + "] after Gerrit server has been renamed from [" + oldName + "] to [" + name + "]."
                         + " This feature is only supported for AbstractProject types e.g. Freestyle Jobs.");
                 return;
@@ -947,7 +947,7 @@ public class GerritServer implements Describable<GerritServer>, Action {
         for (Job job : getConfiguredJobs()) {
 
             if (!(job instanceof AbstractProject)) {
-                logger.info("Unable to remove Gerrit Trigger ffrom job [" + job.getName() + "]. "
+                logger.warn("Unable to remove Gerrit Trigger ffrom job [" + job.getName() + "]. "
                         + " This feature is only supported for AbstractProject types e.g. Freestyle Jobs.");
                 return;
             }
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer.java
@@ -33,6 +33,7 @@ import hudson.Extension;
 import hudson.Functions;
 import hudson.RelativePath;
 import hudson.model.AbstractProject;
+import hudson.model.Job;
 import hudson.model.Action;
 import hudson.model.Describable;
 import hudson.model.Failure;
@@ -904,7 +905,7 @@ public class GerritServer implements Describable<GerritServer>, Action {
      *
      * @return the list of jobs configured with this server.
      */
-    public List<AbstractProject> getConfiguredJobs() {
+    public List<Job> getConfiguredJobs() {
         return PluginImpl.getConfiguredJobs_(name);
     }
 
@@ -914,14 +915,21 @@ public class GerritServer implements Describable<GerritServer>, Action {
      * @param oldName the old name of the Gerrit server
      */
     private void changeSelectedServerInJobs(String oldName) {
-        for (AbstractProject job : PluginImpl.getConfiguredJobs_(oldName)) {
-            GerritTrigger trigger = (GerritTrigger)job.getTrigger(GerritTrigger.class);
+        for (Job job : PluginImpl.getConfiguredJobs_(oldName)) {
+
+            // TODO: find a way to handle add/remove of triggers
+            if (!(job instanceof AbstractProject)) {
+                return;
+            }
+            AbstractProject project = (AbstractProject)job;
+
+            GerritTrigger trigger = (GerritTrigger)project.getTrigger(GerritTrigger.class);
             if (trigger != null) {
                 try {
                     trigger.setServerName(name);
                     trigger.start(job, false);
-                    job.addTrigger(trigger);
-                    job.save();
+                    project.addTrigger(trigger);
+                    project.save();
                 } catch (IOException e) {
                     logger.error("Error saving Gerrit Trigger configurations for job [" + job.getName()
                             + "] after Gerrit server has been renamed from [" + oldName + "] to [" + name + "]");
@@ -934,11 +942,18 @@ public class GerritServer implements Describable<GerritServer>, Action {
      * Remove "Gerrit event" as a trigger in all jobs selecting this server.
      */
     private void removeGerritTriggerInJobs() {
-        for (AbstractProject job : getConfiguredJobs()) {
-            GerritTrigger trigger = (GerritTrigger)job.getTrigger(GerritTrigger.class);
+        for (Job job : getConfiguredJobs()) {
+
+            // TODO: find a way to handle add/remove of triggers
+            if (!(job instanceof AbstractProject)) {
+                return;
+            }
+            AbstractProject project = (AbstractProject)job;
+
+            GerritTrigger trigger = (GerritTrigger)project.getTrigger(GerritTrigger.class);
             trigger.stop();
             try {
-                job.removeTrigger(trigger.getDescriptor());
+                project.removeTrigger(trigger.getDescriptor());
             } catch (IOException e) {
                 logger.error("Error removing Gerrit trigger from job [" + job.getName()
                         + "]. Please check job config");
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/PluginImpl.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/PluginImpl.java
@@ -36,6 +36,7 @@ import com.sonyericsson.hudson.plugins.gerrit.trigger.hudsontrigger.data.Trigger
 import com.sonymobile.tools.gerrit.gerritevents.dto.attr.Provider;
 import com.sonymobile.tools.gerrit.gerritevents.dto.events.GerritTriggeredEvent;
 import hudson.Plugin;
+import hudson.model.Item;
 import hudson.model.Job;
 import hudson.model.Api;
 import hudson.model.Hudson;
@@ -96,7 +97,7 @@ public class PluginImpl extends Plugin {
     public static final Permission RETRIGGER = new Permission(PERMISSION_GROUP,
             "Retrigger",
             Messages._RetriggerPermissionDescription(),
-            Job.BUILD);
+            Item.BUILD);
 
     private static final Logger logger = LoggerFactory.getLogger(PluginImpl.class);
     private final List<GerritServer> servers = new CopyOnWriteArrayList<GerritServer>();
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/PluginImpl.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/PluginImpl.java
@@ -439,7 +439,7 @@ public class PluginImpl extends Plugin {
      */
     public List<Job> getConfiguredJobs(String serverName) {
         LinkedList<Job> configuredJobs = new LinkedList<Job>();
-        for (Job<?, ?> project : Hudson.getInstance().getItems(Job.class)) { //get the jobs
+        for (Job<?, ?> project : Jenkins.getInstance().getItems(Job.class)) { //get the jobs
             GerritTrigger gerritTrigger = GerritTrigger.getTrigger(project);
 
             //if the job has a gerrit trigger, check whether the trigger has selected this server:
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/PluginImpl.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/PluginImpl.java
@@ -36,7 +36,7 @@ import com.sonyericsson.hudson.plugins.gerrit.trigger.hudsontrigger.data.Trigger
 import com.sonymobile.tools.gerrit.gerritevents.dto.attr.Provider;
 import com.sonymobile.tools.gerrit.gerritevents.dto.events.GerritTriggeredEvent;
 import hudson.Plugin;
-import hudson.model.AbstractProject;
+import hudson.model.Job;
 import hudson.model.Api;
 import hudson.model.Hudson;
 import hudson.model.Items;
@@ -96,7 +96,7 @@ public class PluginImpl extends Plugin {
     public static final Permission RETRIGGER = new Permission(PERMISSION_GROUP,
             "Retrigger",
             Messages._RetriggerPermissionDescription(),
-            AbstractProject.BUILD);
+            Job.BUILD);
 
     private static final Logger logger = LoggerFactory.getLogger(PluginImpl.class);
     private final List<GerritServer> servers = new CopyOnWriteArrayList<GerritServer>();
@@ -436,10 +436,10 @@ public class PluginImpl extends Plugin {
      * @param serverName the name of the Gerrit server.
      * @return the list of jobs configured with this server.
      */
-    public List<AbstractProject> getConfiguredJobs(String serverName) {
-        LinkedList<AbstractProject> configuredJobs = new LinkedList<AbstractProject>();
-        for (AbstractProject<?, ?> project : Hudson.getInstance().getItems(AbstractProject.class)) { //get the jobs
-            GerritTrigger gerritTrigger = project.getTrigger(GerritTrigger.class);
+    public List<Job> getConfiguredJobs(String serverName) {
+        LinkedList<Job> configuredJobs = new LinkedList<Job>();
+        for (Job<?, ?> project : Hudson.getInstance().getItems(Job.class)) { //get the jobs
+            GerritTrigger gerritTrigger = GerritTrigger.getTrigger(project);
 
             //if the job has a gerrit trigger, check whether the trigger has selected this server:
             if (gerritTrigger != null && gerritTrigger.getServerName().equals(serverName)) {
@@ -458,7 +458,7 @@ public class PluginImpl extends Plugin {
      */
     @Nonnull
     //CS IGNORE MethodName FOR NEXT 1 LINES. REASON: Static equivalent marker.
-    public static List<AbstractProject> getConfiguredJobs_(String serverName) {
+    public static List<Job> getConfiguredJobs_(String serverName) {
         PluginImpl plugin = getInstance();
         if (plugin == null) {
             logger.debug("Error, plugin instance could not be found!");
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/dependency/DependencyQueueTaskDispatcher.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/dependency/DependencyQueueTaskDispatcher.java
@@ -117,9 +117,9 @@ public final class DependencyQueueTaskDispatcher extends QueueTaskDispatcher
 
     @Override
     public CauseOfBlockage canRun(Queue.Item item) {
-        //AbstractProject check
+        //Job check
         if (!(item.task instanceof Job)) {
-            logger.debug("Not an abstract project: {}", item.task);
+            logger.debug("Not an Job instance: {}", item.task);
             return null;
         }
 
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/dependency/DependencyQueueTaskDispatcher.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/dependency/DependencyQueueTaskDispatcher.java
@@ -25,11 +25,11 @@ package com.sonyericsson.hudson.plugins.gerrit.trigger.dependency;
 
 import hudson.Extension;
 import hudson.ExtensionList;
-import hudson.model.AbstractBuild;
-import hudson.model.AbstractProject;
 import hudson.model.Cause;
 import hudson.model.Item;
+import hudson.model.Job;
 import hudson.model.Queue;
+import hudson.model.Run;
 import hudson.model.queue.QueueTaskDispatcher;
 import hudson.model.queue.CauseOfBlockage;
 import jenkins.model.Jenkins;
@@ -118,7 +118,7 @@ public final class DependencyQueueTaskDispatcher extends QueueTaskDispatcher
     @Override
     public CauseOfBlockage canRun(Queue.Item item) {
         //AbstractProject check
-        if (!(item.task instanceof AbstractProject)) {
+        if (!(item.task instanceof Job)) {
             logger.debug("Not an abstract project: {}", item.task);
             return null;
         }
@@ -143,7 +143,7 @@ public final class DependencyQueueTaskDispatcher extends QueueTaskDispatcher
             logger.debug("Item is already buildable");
             return null;
         }
-        AbstractProject p = (AbstractProject)item.task;
+        Job p = (Job)item.task;
         GerritTrigger trigger = GerritTrigger.getTrigger(p);
         //The project being checked has no Gerrit Trigger
         if (trigger == null) {
@@ -151,7 +151,7 @@ public final class DependencyQueueTaskDispatcher extends QueueTaskDispatcher
             return null;
         }
         //Dependency projects in the build queue
-        List<AbstractProject> dependencies = getProjectsFromString(trigger.getDependencyJobsNames(),
+        List<Job> dependencies = getProjectsFromString(trigger.getDependencyJobsNames(),
                 (Item)p);
         if ((dependencies == null) || (dependencies.size() == 0)) {
             logger.debug("No dependencies on project: {}", p);
@@ -182,7 +182,7 @@ public final class DependencyQueueTaskDispatcher extends QueueTaskDispatcher
         }
 
 
-        List<AbstractProject> blockingProjects = getBlockingDependencyProjects(dependencies, event);
+        List<Job> blockingProjects = getBlockingDependencyProjects(dependencies, event);
 
         if (blockingProjects.size() > 0) {
             return new BecauseDependentBuildIsBuilding(blockingProjects);
@@ -198,12 +198,12 @@ public final class DependencyQueueTaskDispatcher extends QueueTaskDispatcher
      * @param event The event should have also caused the blocking builds.
      * @return the sublist of dependencies which need to be completed before this event is resolved.
      */
-    protected List<AbstractProject> getBlockingDependencyProjects(List<AbstractProject> dependencies,
+    protected List<Job> getBlockingDependencyProjects(List<Job> dependencies,
             GerritTriggeredEvent event) {
-        List<AbstractProject> blockingProjects = new ArrayList<AbstractProject>();
+        List<Job> blockingProjects = new ArrayList<Job>();
         ToGerritRunListener toGerritRunListener = ToGerritRunListener.getInstance();
         if (toGerritRunListener != null) {
-            for (AbstractProject dependency : dependencies) {
+            for (Job dependency : dependencies) {
                 if (toGerritRunListener.isProjectTriggeredAndIncomplete(dependency, event)) {
                     blockingProjects.add(dependency);
                 }
@@ -232,8 +232,8 @@ public final class DependencyQueueTaskDispatcher extends QueueTaskDispatcher
      * @param context The context in which to read the string
      * @return the list of projects
      */
-    public static List<AbstractProject> getProjectsFromString(String projects, Item context) {
-        List<AbstractProject> dependencyJobs = new ArrayList<AbstractProject>();
+    public static List<Job> getProjectsFromString(String projects, Item context) {
+        List<Job> dependencyJobs = new ArrayList<Job>();
         if ((projects == null) || projects.equals("")) {
             return null;
         } else {
@@ -244,9 +244,9 @@ public final class DependencyQueueTaskDispatcher extends QueueTaskDispatcher
                 String projectName = tokens.nextToken().trim();
                 if (!projectName.equals("")) {
                     Item item = jenkins.getItem(projectName, context, Item.class);
-                    if ((item != null) && (item instanceof AbstractProject)) {
-                        dependencyJobs.add((AbstractProject)item);
-                        logger.debug("project dependency job added : {}", (AbstractProject)item);
+                    if ((item != null) && (item instanceof Job)) {
+                        dependencyJobs.add((Job)item);
+                        logger.debug("project dependency job added : {}", (Job)item);
                     }
                 }
             }
@@ -331,15 +331,15 @@ public final class DependencyQueueTaskDispatcher extends QueueTaskDispatcher
     }
 
     @Override
-    public void projectTriggered(GerritEvent event, AbstractProject project) {
+    public void projectTriggered(GerritEvent event, Job project) {
     }
 
     @Override
-    public void buildStarted(GerritEvent event, AbstractBuild build) {
+    public void buildStarted(GerritEvent event, Run build) {
     }
 
     @Override
-    public void buildCompleted(GerritEvent event, AbstractBuild build) {
+    public void buildCompleted(GerritEvent event, Run build) {
     }
 
     @Override
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/GerritMessageProvider.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/GerritMessageProvider.java
@@ -135,7 +135,7 @@ public abstract class GerritMessageProvider implements Serializable, ExtensionPo
         if (thisOverrides("getFileComments", Run.class)) {
             return getFileComments((Run)build);
         }
-        return null;
+        return Collections.emptyList();
     }
 
      /**
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/GerritMessageProvider.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/GerritMessageProvider.java
@@ -25,8 +25,8 @@ package com.sonyericsson.hudson.plugins.gerrit.trigger.gerritnotifier;
 
 import com.sonymobile.tools.gerrit.gerritevents.dto.rest.CommentedFile;
 import hudson.ExtensionPoint;
-import hudson.model.AbstractBuild;
 import hudson.model.Hudson;
+import hudson.model.Run;
 
 import java.io.Serializable;
 import java.util.Collection;
@@ -49,7 +49,7 @@ public abstract class GerritMessageProvider implements Serializable, ExtensionPo
      * @param build Triggered build to provide custom message for
      * @return the custom message
      */
-    public String getBuildStartedMessage(AbstractBuild build) {
+    public String getBuildStartedMessage(Run build) {
         return null;
     }
 
@@ -61,7 +61,7 @@ public abstract class GerritMessageProvider implements Serializable, ExtensionPo
      * @param build Triggered build to provide custom message for
      * @return the custom message
      */
-    public String getBuildCompletedMessage(AbstractBuild build) {
+    public String getBuildCompletedMessage(Run build) {
         return null;
     }
 
@@ -71,7 +71,7 @@ public abstract class GerritMessageProvider implements Serializable, ExtensionPo
      * @param build the build to complain about
      * @return the file comments, default is an empty list.
      */
-    public Collection<CommentedFile> getFileComments(AbstractBuild build) {
+    public Collection<CommentedFile> getFileComments(Run build) {
         return Collections.emptyList();
     }
 
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/GerritMessageProvider.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/GerritMessageProvider.java
@@ -25,6 +25,8 @@ package com.sonyericsson.hudson.plugins.gerrit.trigger.gerritnotifier;
 
 import com.sonymobile.tools.gerrit.gerritevents.dto.rest.CommentedFile;
 import hudson.ExtensionPoint;
+import hudson.Util;
+import hudson.model.AbstractBuild;
 import hudson.model.Hudson;
 import hudson.model.Run;
 
@@ -50,6 +52,27 @@ public abstract class GerritMessageProvider implements Serializable, ExtensionPo
      * @return the custom message
      */
     public String getBuildStartedMessage(Run build) {
+        // If the overriding class still overrides the deprecated method, call that.
+        if (build instanceof AbstractBuild && thisOverrides("getBuildStartedMessage", AbstractBuild.class)) {
+            return getBuildStartedMessage((AbstractBuild)build);
+        }
+        return null;
+    }
+
+    /**
+     * Method allowing plug-ins to provide extra custom messages to Gerrit when a build is started.
+     *
+     * Return null if no message should be added.
+     *
+     * @param build Triggered build to provide custom message for
+     * @return the custom message
+     * @deprecated Use {@link #getBuildStartedMessage(hudson.model.Run)}
+     */
+    @Deprecated
+    public String getBuildStartedMessage(AbstractBuild build) {
+        if (thisOverrides("getBuildStartedMessage", Run.class)) {
+            return getBuildStartedMessage((Run)build);
+        }
         return null;
     }
 
@@ -62,6 +85,27 @@ public abstract class GerritMessageProvider implements Serializable, ExtensionPo
      * @return the custom message
      */
     public String getBuildCompletedMessage(Run build) {
+        // If the overriding class still overrides the deprecated method, call that.
+        if (build instanceof AbstractBuild && thisOverrides("getBuildCompletedMessage", AbstractBuild.class)) {
+            return getBuildCompletedMessage((AbstractBuild)build);
+        }
+        return null;
+    }
+
+    /**
+     * Method allowing plug-ins to provide extra custom messages to Gerrit when a build is completed.
+     *
+     * Return null if no message should be added.
+     *
+     * @param build Triggered build to provide custom message for
+     * @return the custom message
+     * @deprecated Use {@link #getBuildCompletedMessage(hudson.model.Run)}
+     */
+    @Deprecated
+    public String getBuildCompletedMessage(AbstractBuild build) {
+        if (thisOverrides("getBuildCompletedMessage", Run.class)) {
+            return getBuildCompletedMessage((Run)build);
+        }
         return null;
     }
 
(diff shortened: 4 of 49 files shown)


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/dependency/DependencyQueueTaskDispatcher.java
```diff
@@ -118,7 +118,7 @@ public static DependencyQueueTaskDispatcher getInstance() {
     @Override
     public CauseOfBlockage canRun(Queue.Item item) {
         //AbstractProject check
-        if (!(item.task instanceof AbstractProject)) {
+        if (!(item.task instanceof Job)) {
             logger.debug("Not an abstract project: {}", item.task);
```
rsandell: The log is a lie :)

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/GerritMessageProvider.java
```diff
@@ -49,7 +49,7 @@
      * @param build Triggered build to provide custom message for
      * @return the custom message
      */
-    public String getBuildStartedMessage(AbstractBuild build) {
+    public String getBuildStartedMessage(Run build) {
```
rsandell: Will these changes be binary compatible with https://github.com/jenkinsci/build-failure-analyzer-plugin/blob/master/src/main/java/com/sonyericsson/jenkins/plugins/bfa/GerritMessageProviderExtension.java ?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/dependency/DependencyQueueTaskDispatcher.java
tfennelly: Thanks.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ToGerritRunListener.java
```diff
@@ -428,15 +429,18 @@ private String obtainFailureMessage(@Nullable GerritTriggeredEvent event,
...
-                // Check for ANT-style file path
-                FilePath[] matches = this.getMatchingWorkspaceFiles(build.getWorkspace(), filepath);
-                logger.debug("Found matching workspace files: {}", matches);
-
-                if (matches.length > 0) {
-                    // Use the first match
-                    FilePath path = matches[0];
-                    content = this.getExpandedContent(path, envVars);
-                    logger.info("Obtained failure message from file: {}", content);
+                // TODO: what will we do here?
```
rsandell: I've seen code something like ```getCurrentExecutor(build).getWorkspace()``` appear now and then.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/GerritMessageProvider.java
tfennelly: I hadn't seen that one, but I know there are some issues lying ahead for some other external sony lib that assumes `AbstractBuild` or `AbstractProject`. I forget the name of it right now but will tell you when I hit it again.

tfennelly: I hope we can fix those and cut new releases of them, yeah?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/EventListener.java
```diff
@@ -164,25 +166,39 @@ protected void schedule(GerritTrigger t, GerritCause cause, GerritTriggeredEvent
...
+        if (project instanceof AbstractProject) {
+            build = ((AbstractProject)project).scheduleBuild2(
                 projectbuildDelay,
                 cause,
                 badgeAction,
                 new RetriggerAction(cause.getContext()),
                 new RetriggerAllAction(cause.getContext()),
                 parameters);
+        } else if (project instanceof SCMTriggerItem) {
+            // TODO: Using SCMTriggerItem smells bad here. What should we be using?
```
rsandell: Similar JobMixin trick that dockerhub trigger does?
https://github.com/jenkinsci/dockerhub-notification-plugin/blob/master/src/main/java/org/jenkinsci/plugins/dockerhub/notification/DockerHubWebHook.java#L167

rsandell:
@scoheb, @GLundh, @TWestling  I'm going on vacation today, so I won't be much help to @tfennelly during that time. Could you help him out with some friendly review?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/GerritMessageProvider.java
rsandell: Yes I'm a maintainer of gerrit-events as well :)

rsandell: But that lib should have no dependency on Jenkins at all.

GLundh:
@rsandell: I would love to help out, but I'm going on vacation myself today, and wont be able to spend much time on it until early august.

scoheb:
@rsandell ... I can definitely lend a hand!

jacob-keller:
@tfennelly  I'm willing to attempt running a version as well, to see how it goes. Let me know what I can do to help.

tfennelly:
@scoheb @GLundh @jacob-keller Thanks guys :beer:  ... I'll ping ye as soon as I have something testable.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/EventListener.java
```diff
@@ -164,25 +166,39 @@ protected void schedule(GerritTrigger t, GerritCause cause, GerritTriggeredEvent
...
         BadgeAction badgeAction = new BadgeAction(event);
         //during low traffic we still don't want to spam Gerrit, 3 is a nice number, isn't it?
         int projectbuildDelay = t.getBuildScheduleDelay();
         if (cause instanceof GerritUserCause) {
             // it's a manual trigger, no need for a quiet period
             projectbuildDelay = 0;
-        } else if (project.getHasCustomQuietPeriod()
-                && project.getQuietPeriod() > projectbuildDelay) {
-            projectbuildDelay = project.getQuietPeriod();
+        } else if (project instanceof AbstractProject) {
```
rsandell: ```ParameterizedJobMixing.ParameterizedJob``` which both AbstractProject and Workflow implements has quietPeriod methods IIRC.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -377,11 +377,21 @@ void onJobRenamed(String oldFullName, String newFullName) {
...
      */
-    public static GerritTrigger getTrigger(@Nullable AbstractProject project) {
+    public static GerritTrigger getTrigger(@Nullable Job project) {
         if (project == null) {
             return null;
         }
-        return (GerritTrigger)project.getTrigger(GerritTrigger.class);
+
+        // TODO: I'd imagine this needs to be optimized?
+        DescribableList<Trigger<?>, TriggerDescriptor> triggers =
```
rsandell: ```ParameterizedJobMixing.ParameterizedJob``` has a ```getTriggers()``` method.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -1918,8 +1928,12 @@ public synchronized void scheduled(ChangeBasedEvent event, ParametersAction para
          *            The event that originally triggered the build.
          */
         private void cancelJob(GerritTriggeredEvent event) {
+            if (!(job instanceof Queue.Task)) {
+                logger.error("Error canceling job. The job is not of type Task.");
+            }
```
rsandell: I guess you should return directly after this?

And please log the task as well, so the admin knows which of the 10k jobs the log is talking about :)

rsandell:
@scoheb @jacob-keller Thanks guys! Then @tfennelly is in good hands while I relax with some :beers: myself :)

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/EventListener.java
tfennelly: Rights, but it doesn't have `AbstractProject.getHasCustomQuietPeriod`. We'd have to drop that. Would that be a terrible thing and cause regressions?

tfennelly: Note `getQuietPeriod` defaults if not set.

jacob-keller: I personally don't think that's an issue.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer.java
```diff
@@ -914,14 +915,21 @@ private void rename(String newName) {
      * @param oldName the old name of the Gerrit server
      */
     private void changeSelectedServerInJobs(String oldName) {
-        for (AbstractProject job : PluginImpl.getConfiguredJobs_(oldName)) {
-            GerritTrigger trigger = (GerritTrigger)job.getTrigger(GerritTrigger.class);
+        for (Job job : PluginImpl.getConfiguredJobs_(oldName)) {
+
+            // TODO: find a way to handle add/remove of triggers
+            if (!(job instanceof AbstractProject)) {
+                return;
```
amuniz: What would happen if it's a ```WorkflowJob```?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/EventListener.java
rinrinne: If not check `AbstractProject.getHasCustomQuietPeriod`, `ParameterizedJob.getQuietPeriod` might return system value given by `Jenkins.getQuietPeriod` if `quietPeriod` in project is not set. system's default is 5.

BTW, can we use GT's own period only?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ToGerritRunListener.java
amuniz: Or ```build.getCurrentExecutor().getWorkspace()``` but ```getCurrentExecutor()``` will always return null unless there were a build in progress.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/EventListener.java
amuniz: Could something like this be used here?
```
ParameterizedJobMixIn scheduler = new ParameterizedJobMixIn() {
    @Override protected Job asJob() {
        return project;
    }
};
scheduler.scheduleBuild(projectbuildDelay)
```

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/EventListener.java
jacob-keller: So, if I understand this right, the old behavior was something like:

2. If we're manual trigger, don't delay at all
3. If the job has a custom quiet period (ie: not default) AND that period is larger than the ScheduleDelay, use this
4. Otherwise, use the ScheduleDelay 

So what's wrong with using the quietPeriodDelay even if it's not customized? Why can't we also check if we're above the default?

Is it so that GT's custom delay overrides the global default, but not per-jobs?

I'd rather use the longest of all configured delays, not the shortest. Ie: GT's delay only overrides global if the global is shorter?

We could also use only GT's trigger period, without worrying about per-job settings?

Thoughts?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer.java
jacob-keller: Does WorkflowJob have a way to add and remove triggers? I assume we'd want something similar to that? I know that the ParameterizedJobMixin does not have support for add triggers, but maybe we can just instanceOf to WorkflowJob here? 

According to the source for WorkflowJob.java, it appears that WorkflowJob has an "addTrigger" function and getTriggers function, but this isn't exposed by the Mixin API. It would feel really awkward to instanceof and then use the same code both times.... Also, it doesn't support "getTrigger" standalone only the getTriggers...

Kind of ugly either way.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/EventListener.java
tfennelly: @jacob-keller @rinrinne guys, I'm afraid I don't really have a valuable opinion on stuff like this. I will be relying on ye to tell me what makes most sense from a Gerrit pov.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer.java
amuniz: The ```getTriggers()``` method is defined in ```ParameterizedJobMixIn.ParameterizedJob```. We can cast to this one better than ```WorkflowJob```.

tfennelly: @amuniz At the moment, nothing with happen for a Workflow Job i.e. add/remove of triggers is not supported for workflow as things stand. That's why I marked it with a `TODO`. We need to find a way of handling it for workflow. `AbstractProject` has methods for adding and removing triggers, but that's no help for a Workflow job.

Trigger `rename` might be fixable by finding the current trigger and renaming the server on it Vs  removing and readding the trigger (at the moment, `rename` is supported by removing the trigger and adding a new one). Not sure yet how `remove` will be supported (when a Gerrit server is removed we need to find all Jobs with triggers for that server and remove them).

tfennelly:
> Didn't @rsandell suggest also to update the error to include the job name?

@jacob-keller Where's that?

tfennelly:
The tests compile now, but there are failures galore :) Maybe you guys can help me make sense of these?

jacob-keller:
@tfennelly, it was suggested that you add the job name as part of the failure message, but you didn't get to that yet.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer.java
jacob-keller: you can add a trigger easily. It seems like the trigger add code will remove the old trigger already. Sadly WorkflowJob.java does not export a remove trigger function like the code does. Maybe we just don't bother removing the old trigger at all when a server is removed?

What's the reason that we actually have remove/add trigger code here specifically? I presume to prevent a bug? What do other trigger plugins do to prevent this sort of issue?

tfennelly:
@rsandell @scoheb @jacob-keller @TWestling @rinrinne We have a [build with passing test suite](https://jenkins.ci.cloudbees.com/job/plugins/job/gerrit-trigger-plugin/571/). Are you guys in a position to test it?

scoheb:
I have started testing!

tfennelly:
@scoheb Thanks Scott. I suppose the first most important thing is that I haven't broken anything for non Workflow builds (Freestyle etc).

scoheb:
Will do!

jacob-keller:
Same here I can run it.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer.java
amuniz: Triggers are not dynamically added/removed in other plugins (for example GitHub plugin, which includes a trigger on push feature). The trigger is added or removed in the UI (a checkbox).

I don't know if there is some special case with Gerrit that requires to add/remove triggers dynamically.

tfennelly:
@scoheb @jacob-keller thanks guys.

jacob-keller:
Biggest complaint so far is the lack of modification to the workflow page. AbstractProject web pages show things like "re-trigger" links, and such, these don't appear to be setup for Workflow jobs, only regular ones?

jacob-keller:
Secondly, it doesn't appear to insert the GERRIT_* environment variables into the workflow environment.

It does appear to trigger correctly, and it does run the job, so that part is good so far.

jacob-keller:
It appears that parameters aren't being set in workflow.. according to documentation, workflow should be able to access parameters via their name as variables:

echo "${GERRIT_BRANCH}"

But it causes a crash, without finding the property.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/EventListener.java
jacob-keller: This causes a regression for Workflows because we don't actually include the parameters.

We can pass all the actions, and at minimum we should do that even if we stick to SCMTriggerItem. You failed to pass badgeAction, and both Retrigger actions, as well as the parameters.

We can't pass the cause directly though, at least not to SCMTriggerItem

tfennelly:
@jacob-keller thanks a lot for that Jacob. There are issues, but it doesn't sound too bad; we're making progress I think :beers: 

I can't work on this for a day or two so it will be then before I get to address the issues you have identified.

scoheb:
My testing has been positive so far. I mainly have some high volume freestyle jobs that are triggered by Gerrit Replication Events.

jacob-keller:
I submitted a pull request to your branch that should deal with almost all
of the issues I was seeing.

tfennelly:
@jacob-keller That's awesome. Thanks Jacob. Your PR is merged.

tfennelly:
Forcing a rebuild. Looks like a flaky test failed.

jacob-keller:
Ok, I installed this version of HPI, and the gerrit trigger icons and parameters now work as expected. A minor issue is ofcourse the workflow jobs don't get the cause set, but that is not a deal breaker.

Other than that, I'll keep testing this and update if I see any other issues.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer.java
jglick: As @amuniz says, programmatically removing a trigger is eyebrow-raising. Normally a human should be in charge of the trigger list—or at least `job-dsl-plugin`, `cloudbees-template`, etc.

jacob-keller: What do other plugins that provide triggers do? This is the case where a gerrit server is being removed.. so do we just not do this at all? That makes sense to me at least...

tfennelly: Auto removing zombie triggers after a Gerrit server is removed seems to make sense to me.

In any case (and imo), the debate on the pros and cons of programmatically removing a trigger is something that should happen separate to this PR. IMO, our job here is to integrate workflow and maintain the same feature set. It shouldn't be about proposing fundamental plugin design changes.

jacob-keller: On Jul 27, 2015 12:48 AM, "Tom Fennelly" <notifications@github.com> wrote:
>
> In
src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer.java:
>
> > @@ -914,14 +915,21 @@ private void rename(String newName) {
> >       * @param oldName the old name of the Gerrit server
> >       */
> >      private void changeSelectedServerInJobs(String oldName) {
> > -        for (AbstractProject job :
PluginImpl.getConfiguredJobs_(oldName)) {
> > -            GerritTrigger trigger =
(GerritTrigger)job.getTrigger(GerritTrigger.class);
> > +        for (Job job : PluginImpl.getConfiguredJobs_(oldName)) {
> > +
> > +            // TODO: find a way to handle add/remove of triggers
> > +            if (!(job instanceof AbstractProject)) {
> > +                return;
>
> Auto removing zombie triggers after a Gerrit server is removed seems to
make sense to me.
>
> In any case (and imo), the debate on the pros and cons of
programmatically removing a trigger is something that should happen
separate to this PR. IMO, our job here is to integrate workflow and
maintain the same feature set. It shouldn't be about proposing fundamental
plugin design changes.
>
> —
> Reply to this email directly or view it on GitHub.

While that is true.. I don't know of a good way to implement this
functionality and it may make sense to do the PR to remove such a feature
first unless we can actually do this change..

One way might be to optionally depend on work flow plug in.  I have seen
others do this but I don't know how they did it.

tfennelly: @jacob-keller Right, I didn't see a way of doing it for workflow. So, the easiest thing (for now) might be to mark it as a missing feature for Workflow jobs and not have that point block this PR?

Then, separately have a discussion on the merits/demerits etc of auto-removing triggers. This will either result in a task to remove the feature completely (for all build types), or a task to find a way of making it work for workflow.

jacob-keller: I think that is fine myself. I would rather not block trigger support just
because we are missing a relatively minor feature.

tfennelly:
@jacob-keller @scoheb While testing I see the `nulljob` in the event being sent back to the gerrit server (from the gerrit plugin):

screenshot 2015-07-28 09 24 56

Does that look like a regression or a config issue?

I had to tweak away from default settings in order to get a patchset created event that the gerrit plugin found "interesting" e.g. I had to specify a "Branches Pattern" of type "Path" with value "**". Is that the norm?

screenshot 2015-07-28 09 29 53

tfennelly:
Ignore above comment re `nulljob`. It's because I'm running the plugin via HPI.

amuniz:
@tfennelly what's the problem with hpi? Is it a know issue?

scoheb:
Hi,

The **nulljob** comes from the Jenkins URL not being set in the Global Config section. Or you have never tried to persist the settings to by Saving or Applying.

Also, you putting a ** for Path is entirely normal:

Plain: The exact name in Gerrit, case sensitive equality.
Path: ANT style pattern. Ex: "**/base/*"

Hope that helps!

Scott

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer.java
jglick: Agreed with @tfennelly’s summary.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/PluginImpl.java
```diff
@@ -96,7 +96,7 @@
     public static final Permission RETRIGGER = new Permission(PERMISSION_GROUP,
             "Retrigger",
             Messages._RetriggerPermissionDescription(),
-            AbstractProject.BUILD);
+            Job.BUILD);
```
jglick: FWIW this is defined in `Item`.

jglick:
:bug: for the acknowledged issue of possible binary incompatibilities for plugins depending on this one. Alas there is no well-defined API exported by the plugin; anyone out there might be using any public member. [This search](https://github.com/search?utf8=%E2%9C%93&q=user%3Ajenkinsci+user%3Acloudbees+-repo%3Ajenkinsci%2Fgerrit-trigger-plugin+%22com.sonyericsson.hudson.plugins.gerrit.trigger.%22&type=Code&ref=searchresults) turns up only a few hits among likely candidates. [This search](https://github.com/search?utf8=%E2%9C%93&q=-repo%3Ajenkinsci%2Fgerrit-trigger-plugin+language%3Ajava+-filename%3ATest+%22com.sonyericsson.hudson.plugins.gerrit.trigger.%22&type=Code&ref=searchresults) has more to wade through (mostly forks?) but could be considered too.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ToGerritRunListener.java
jglick: There is no `getCurrentExecutor` anywhere in Jenkins core, so I am not sure what you could mean by that.

Currently there is nothing else to be done here except check for an instance of `AbstractBuild`. [JENKINS-26138](https://issues.jenkins-ci.org/browse/JENKINS-26138) might involve introducing a core API whereby you could enumerate workspaces associated with a `Run`.

In the meantime, it seems reasonable to just fall back to a message that does not rely on knowledge of workspace files.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/EventListener.java
```diff
@@ -164,25 +166,45 @@ protected void schedule(GerritTrigger t, GerritCause cause, GerritTriggeredEvent
...
+            // TODO: How about using ParameterizedJobMixIn.ParameterizedJob here?
```
jglick: Yes, that is the usual interface for finding a `quietPeriod`.

I do not follow the logic relating to `hasCustomQuietPeriod`; this is definitely unusual. One of many “snowflake” features in this plugin…

More broadly, my impression is that quiet periods are of little importance to people not using archaic systems like CVS that lack atomic commits.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/EventListener.java
jglick: You need to use `ParameterizedJobMixIn.scheduleBuild2`. This handles all (schedulable!) project types. Then just add a `TODO` comment reminding the plugin maintainer to use [the simplified `static` convenience method](https://github.com/jenkinsci/jenkins/blob/e55f356ed0132b966200e5266a5e382a1c019388/core/src/main/java/jenkins/model/ParameterizedJobMixIn.java#L108-127) as of 1.621.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -377,11 +377,21 @@ void onJobRenamed(String oldFullName, String newFullName) {
...
         if (project == null) {
             return null;
         }
-        return (GerritTrigger)project.getTrigger(GerritTrigger.class);
+
+        if (project instanceof ParameterizedJob) {
+            ParameterizedJob parameterizedJob = (ParameterizedJob)project;
+            for (Trigger p : parameterizedJob.getTriggers().values()) {
+                if (GerritTrigger.class.isInstance(p)) {
+                    return GerritTrigger.class.cast(p);
```
jglick:
```java
// TODO use convenience method as of 1.621+
```

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/replication/ReplicationFailedHandler.java
```diff
@@ -40,6 +40,7 @@
 @Extension
 public class ReplicationFailedHandler extends RunListener<AbstractBuild<?, ?>> {
 
+    // TODO: scooby doo'd here wrt AbstractBuild. What are the implications?
```
jglick: There is no equivalent to `RunListener.setUpEnvironment` that would work for Workflow builds, nor probably could there be one; the implicit assumption would not hold.

Anyway the comment below shows that the usage here has nothing to do with environment variables, the plugin just wants to kill builds that have this action present, and this is the only apparent API to do that.

I do not really follow what `ReplicationQueueTaskDispatcher` is doing. Under some conditions it adds this action, which would cause the build to fail if it were scheduled, yet then returns `null` from `canRun`, indicating that the build may be scheduled. Mystery.

jglick: Offhand it sounds like an esoteric feature that few Workflow users would miss. If you really need to try something similar, at least override `onStarted` to print a warning, and call `setResult(FAILURE)`; the build will continue but perhaps it will die later.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/EventListener.java
jglick: > We can't pass the cause directly though

Just use `new CauseAction(cause)`.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/EventListener.java
jacob-keller: Agreed. The only other possibility would be if the user pushes several separate commits using separate git-push...

I think that's rather not worth it to care much about it, but the plugin already does support things this way...

I don't really get the idea here, but i think it's something like "if job has its own quiet period, use that, unless we set a higher build delay ourselves, then use that instead.  But if a project doesn't have a customized build delay and only has the global quiet period, just use our value instead of the larger of the values.

That seems really convoluted logic, and we could just use "max" of quiet period or build delay.

I think the primary reason we have build delay is incase of git mirrors, where you clone from a mirror, but monitor based on a gerrit server, and want to wait long enough for the mirror to be updated so that it has the refs you need to pull.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/replication/ReplicationFailedHandler.java
jacob-keller: I'm not even sure what this feature is for.

tfennelly:
> :bug: for the acknowledged issue of possible binary incompatibilities for plugins depending on this one. 

Yeah, this is the biggest issue. What do @rsandell @jacob-keller @scoheb @GLundh @TWestling think? Is it essential that backward compatibility is maintained? What has been your policy on this in the past? Is there an unofficial API that we could focus on here i.e. a set of classes that we maintain backward compatibility for and, after that, backward compatibility is not offered?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/EventListener.java
tfennelly: Ok, I'm going to switch to `ParameterizedJobMixIn.ParameterizedJob` and not use `v` for `WorkflowJob`. We'll see if anyone complains :wink:

tfennelly:
As for backward compatibility, only two `@ExtensionPoint`s are defined in this plugin: `GerritMessageProvider` and `GerritTriggeredBuildListener`. Of those, only `GerritMessageProvider` has changed. I'll update it at least and provide backward compat.

tfennelly:
@reviewbybees (sorry ... I didn't realise I forgot to add this)

jglick:
:bug: you need to use [this idiom](https://github.com/jenkinsci/jenkins/blob/e55f356ed0132b966200e5266a5e382a1c019388/core/src/main/java/jenkins/model/ParameterizedJobMixIn.java#L108-127) (`TODO 1.621+` etc.) and pass a `new CauseAction(cause)`.

jglick:
Also recommend fixing `TriggerDescriptor.isApplicable` to check for `ParameterizedJob`.

jglick:
:bug: add a Workflow-based [acceptance test](https://github.com/jenkinsci/acceptance-test-harness/blob/c375f31e67766e0f2867d269aedc94a0379652de/src/test/java/plugins/GerritTriggerTest.java#L57-63).

tfennelly:
@jacob-keller not sure I know what you mean. Can you show an example?

tfennelly:
@jacob-keller @scoheb hey guys ... how either of you ever run the [acceptance-test-harness](https://github.com/jenkinsci/acceptance-test-harness) for this plugin? If yes, any tips on how to run it?

jacob-keller:
I haven't run it.

scoheb:
I can run it tomorrow and provide you with a small howto                                                                                                                                                                                                                                                                                                                                        Sent from my BlackBerry 10 smartphone.                                                                                                                                                                                                                From: Tom FennellySent: Wednesday, August 5, 2015 12:26To: jenkinsci/gerrit-trigger-pluginReply To: jenkinsci/gerrit-trigger-pluginCc: Scott HebertSubject: Re: [gerrit-trigger-plugin] [WiP] [JENKINS-26010] Workflow compatibility (#240)@jacob-keller @scoheb hey guys ... how either of you ever run the acceptance-test-harness for this plugin? If yes, any tips on how to run it?

—Reply to this email directly or view it on GitHub.

tfennelly:
@scoheb Hey Scot, sorry for the late reply. No need, I figured it out. Thanks.

tfennelly:
@rsandell @jacob-keller @scoheb @reviewbybees (and anyone else) please review

jglick:
:bee:

amuniz:
:bee: !!

tfennelly:
@rsandell @jacob-keller @scoheb calling ye again to have a look at this and give it a thumbs up/down for merging. Thanks.

jacob-keller:
My preference would be a rebase on this to squash in fixes so that bugs are not introduced and then fixed in a later sequence. However, I don't know how strongly the jenkins project cares about such efforts.

The end result looks great now.

rsandell:
@tfennelly the backwardcompat strategy we've had in the past is to do a "best effort" to keep it but if it turns out to be too much work then its ok to ignore, and as you said; its a must for the extension points, there is one class called _Something**Api**_ I don't remember the exact name, but that's also one we should do some extra effort to keep compatible.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer.java
```diff
@@ -914,14 +915,23 @@ private void rename(String newName) {
      * @param oldName the old name of the Gerrit server
      */
     private void changeSelectedServerInJobs(String oldName) {
-        for (AbstractProject job : PluginImpl.getConfiguredJobs_(oldName)) {
-            GerritTrigger trigger = (GerritTrigger)job.getTrigger(GerritTrigger.class);
+        for (Job job : PluginImpl.getConfiguredJobs_(oldName)) {
+
+            if (!(job instanceof AbstractProject)) {
+                logger.info("Unable to modify Gerrit Trigger configurations for job [" + job.getName()
```
rsandell: this should be ```warning``` instead of ```info```

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer.java
```diff
@@ -934,11 +944,19 @@ private void changeSelectedServerInJobs(String oldName) {
      * Remove "Gerrit event" as a trigger in all jobs selecting this server.
      */
     private void removeGerritTriggerInJobs() {
-        for (AbstractProject job : getConfiguredJobs()) {
-            GerritTrigger trigger = (GerritTrigger)job.getTrigger(GerritTrigger.class);
+        for (Job job : getConfiguredJobs()) {
+
+            if (!(job instanceof AbstractProject)) {
+                logger.info("Unable to remove Gerrit Trigger ffrom job [" + job.getName() + "]. "
```
rsandell: warning

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/PluginImpl.java
```diff
@@ -436,10 +437,10 @@ public static GerritHandler getHandler_() {
      * @param serverName the name of the Gerrit server.
      * @return the list of jobs configured with this server.
      */
-    public List<AbstractProject> getConfiguredJobs(String serverName) {
-        LinkedList<AbstractProject> configuredJobs = new LinkedList<AbstractProject>();
-        for (AbstractProject<?, ?> project : Hudson.getInstance().getItems(AbstractProject.class)) { //get the jobs
-            GerritTrigger gerritTrigger = project.getTrigger(GerritTrigger.class);
+    public List<Job> getConfiguredJobs(String serverName) {
+        LinkedList<Job> configuredJobs = new LinkedList<Job>();
+        for (Job<?, ?> project : Hudson.getInstance().getItems(Job.class)) { //get the jobs
```
rsandell: Inherital sin: change ```Hudson``` to ```Jenkins```

rsandell: Inherital sin: change ```Hudson``` to ```Jenkins```

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/GerritMessageProvider.java
```diff
@@ -71,10 +115,29 @@ public String getBuildCompletedMessage(AbstractBuild build) {
...
+     * @param build the build to complain about
+     * @return the file comments, default is an empty list.
+     * @deprecated Use {@link #getFileComments(hudson.model.Run)}
+     */
+    @Deprecated
+    public Collection<CommentedFile> getFileComments(AbstractBuild build) {
+        if (thisOverrides("getFileComments", Run.class)) {
+            return getFileComments((Run)build);
+        }
+        return null;
```
rsandell: IIRC the previous contract was to return an empty list and not null?, see method above

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ToGerritRunListener.java
```diff
@@ -428,15 +429,19 @@ private String obtainFailureMessage(@Nullable GerritTriggeredEvent event,
                 // The filename may contain environment variables
                 filepath = envVars.expand(filepath);
 
-                // Check for ANT-style file path
-                FilePath[] matches = this.getMatchingWorkspaceFiles(build.getWorkspace(), filepath);
-                logger.debug("Found matching workspace files: {}", matches);
+                if (build instanceof AbstractBuild) {
```
rsandell: So is there still no way to solve this for a workflow? If so then the config page should reflect that.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ToGerritRunListener.java
```diff
@@ -428,15 +429,19 @@ private String obtainFailureMessage(@Nullable GerritTriggeredEvent event,
...
-                    content = this.getExpandedContent(path, envVars);
-                    logger.info("Obtained failure message from file: {}", content);
+                    if (matches.length > 0) {
+                        // Use the first match
+                        FilePath path = matches[0];
+                        content = this.getExpandedContent(path, envVars);
+                        logger.info("Obtained failure message from file: {}", content);
+                    }
+                } else {
+                    logger.info("Unable to find matching workspace files for job type {}", build.getClass().getName());
```
rsandell: should be log level ```warning```

rsandell: And the job name should be logged as well.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/job/ssh/BuildStartedCommandJob.java
```diff
@@ -59,9 +59,9 @@
      * @param taskListener a listener.
      * @param event        the event.
      * @param stats        the stats.
-     * @see GerritNotifier#buildStarted(AbstractBuild, TaskListener, GerritTriggeredEvent, BuildsStartedStats)
+     * @see GerritNotifier#buildStarted(Run, TaskListener, GerritTriggeredEvent, BuildsStartedStats)
      */
-    public BuildStartedCommandJob(IGerritHudsonTriggerConfig config, AbstractBuild build,
+    public BuildStartedCommandJob(IGerritHudsonTriggerConfig config, Run build,
```
rsandell: This might need to keep its old Constructor signature just in case for backwards compat issues.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -548,24 +564,51 @@ protected void schedule(GerritCause cause, GerritTriggeredEvent event) {
...
+     * Added here to facilitate unit testing.
+     *
+     * @param theJob The job.
+     * @param quitePeriod Quite period.
+     * @param cause Build cause.
+     * @param badgeAction build badge action.
+     * @param parameters Build parameters.
+     * @return Scheduled build future.
+     */
+    protected Future schedule(final Job theJob, int quitePeriod, GerritCause cause, BadgeAction badgeAction,
```
rsandell: This does not belong in this class, should be in the listener.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/TriggeredItemEntity.java
```diff
@@ -123,9 +127,10 @@ public void setBuild(AbstractBuild build) {
          * If this object is newly deserialized, the project will be looked up from {@link #getProjectId() }
          * @return the project.
          */
-        public AbstractProject getProject() {
+        @WithBridgeMethods(AbstractProject.class)
+        public Job getProject() {
             if (project == null) {
-                project = Hudson.getInstance().getItemByFullName(projectId, AbstractProject.class);
+                project = Hudson.getInstance().getItemByFullName(projectId, Job.class);
```
rsandell: inherital sin.

rsandell:
:bug: Found mostly just small nits.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ToGerritRunListener.java
tfennelly: I's not possible and I'm not sure what would be involved doing that for workflow. I think we can leave that for now and come back to it as a separate task.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/job/ssh/BuildStartedCommandJob.java
tfennelly: Where's the compatibility issue? `AbstractBuild` is a `Run`, so there shouldn't be an issue.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
tfennelly: Right, but putting it anywhere else makes unit testing a total PITA. I was around the house on this one trying to make it fit into the listener, but doing so was a total pain. [See this comment](https://github.com/jenkinsci/gerrit-trigger-plugin/pull/240/files#diff-462a2e225475f498620d2751fa464ca0R599).

tfennelly:
@rsandell I fixed what I reasonably could from your review. I don't think the others should hold up this PR so hopefully you can take back that bug please ;)

tfennelly:
@rsandell and if you can point me to the `SomethingApi` that you are talking about then we can review that. I went through all changes and fixed backward compatibility wherever it was relevant, so there's nothing outstanding in this PR afaik.

tfennelly:
@rsandell oh and I purposely didn't touch/fix things like the use of `Hudson` instead of `Jenkins`. I fixed the ones you pointed to, but in general I do not consider that to be part of what this PR is about.

jglick:
:bee:

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ToGerritRunListener.java
rsandell:
```
<j:if test="${it instanceof AbstractBuild}">
  <f:entry .../>
</j:if>
```

rsandell:
https://github.com/jenkinsci/gerrit-trigger-plugin/blob/master/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/api/GerritTriggerApi.java

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ToGerritRunListener.java
tfennelly: So now I have a little bit better an idea what you are talking about :smile:. Next question: where specifically is this reflected in the UI i.e. in which Jelly files and on which config entries? Maybe `buildFailureMessage` in `src/main/resources/com/.../hudsontrigger/GerritTrigger/config.jelly?`

tfennelly:
@rsandell `GerritTriggerApi` didn't change.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ToGerritRunListener.java
rsandell: It looks like ```buildUnsuccessfulFilepath```.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/job/ssh/BuildStartedCommandJob.java
rsandell: ok

rsandell:
@tfennelly on the use of ```Hudson``` instead of ```Jenkins```; This plugin contains too many old/bad practices for any one person to fix them all in a simple pull request. So we use the principle of _inhetital sin_ i.e. if you touch a line you inherit all the sins of that line and gets asked to fix it, because they are now your sins :) That way we can hopefully inch our way to a better code base instead of giving up and never do anything about it.

That does not apply to for example bad architecture that spans over multiple lines and files though. I did not ask you for example to convert the rebuild actions to use a transient factory even though you've touched lines in relation to that ;)

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
rsandell: Hmm, I still don't like it. I put a lot of effort into moving all scheduling logic out of this class not long ago. I need to think about this a bit, either to convince myself that it's OK or find a better way. I think I understand your argument though and I'm all for making the code more testable.

tfennelly:
@rsandell While I'm a person of zero religion/faith or supernatural beliefs of any kind now, I would have been born, baptised and educated within the Roman Catholic religion (as with most Irish people). So, I'm very familiar with sick ideas such as that of [Ancestral/Original sin](https://en.wikipedia.org/wiki/Ancestral_sin) i.e. inheriting someone elses sins.

So while I always thought I had been "cleansed" at baptism, I now find my sole has been tarnished with your bad deed. Is that how it works? Grrrrrrrr :smile_cat:

tfennelly:
@rsandell anyway ... back to the main point ... this PR needs to move forward if we can :smile:

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTriggerTest.java
```diff
@@ -1149,14 +1154,6 @@ public void testGerritEventManualEvent() {
         trigger.createListener().gerritEvent(event);
 
         verify(listener).onTriggered(same(project), same(event));
-
-        verify(project).scheduleBuild2(
```
rsandell: you are removing a verification step here, so the test isn't testing the same thing as before anymore.

rsandell: You are no longer verifying that the schedule2 method is actually called, the proxy above will only fail if it gets scheduled with bad parameters, but will pass the test if its not scheduled at all.

rsandell:
Repent and thou shall be merged!

Sorry couldn't help myself :smiling_imp: 

Moving forward: I found some motivation to still not like the schedule thingie. :bug: for removing the verify of schedule2. But I've gone through the rest and I it looks good.

tfennelly:
@rsandell fair point on the non-verify of the `schedule2` call. Will fix that now. Thanks.

tfennelly:
@rsandell ok, that should be ok now (assuming CI tests pass + over-picky checkstyles).

rsandell:
Sorry @tfennelly I was a bit lazy in my last comment. 
* I'm still not liking the schedule method in the trigger class.
* the verification that the job is actually scheduled still has the potential of getting borked with future updates, i.e. changes could be made in the plugin that unknowingly works around the test verification, maybe you could spy on ```Jenkins.queue.schedule```?
* and the if here https://github.com/jenkinsci/gerrit-trigger-plugin/blob/master/src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger/config.jelly#L171

tfennelly:
@rsandell added the `<if>`. Hopefully the tests all still pass coz not sure that `instanceof` check will work in the jelly.

As for the `schedule`, there may be a way of doing it differently by spying on the `Queue`, but I'd really not like to start into another round of dev on this PR for something that's not actually "broken" (i.e. we just dislike the current approach). This PR has now been sitting around for nearly a month (waiting mostly on you :wink:). Is there any chance we can move these changes along and revisit the `schedule` issue to explore other ways of doing it.

The `schedule` method is `protected`, but I can also mark it with that `@InternalUseOnly` annotation (forget the exact name).

rsandell:
But you did break the tests, they don't test the same thing any more.

tfennelly:
@rsandell specifically where are they not testing the same thing? You were saying that it was no longer verifying that the `schedule` method was bing called. That's fixed now. Where is the difference now?

rsandell:
I fixed the build failure and cleaned up the tests in 136d5114a38f0d3fc0da8321af70ddfdf7fdb783 after the merge.

rsandell:
https://wiki.jenkins-ci.org/display/JENKINS/Gerrit+Trigger#GerritTrigger-Version2.15.0beta1%28releasedAug172015%29

tfennelly:
@rsandell thanks for that Bobby. I'll take it for a drive and test it again.

tfennelly:
@reviewbybees done

PR #241, opened 2015-08-31
[WiP] Upgrade gerrit-events to 2.9.0

dpursehouse:
- Add support for the merge-failed event
- Add support for the reviewer-added event
- Add support for the topic-changed event
- Fix up comments in GerritEventType enum
- Don't use deprecated DefaultHttpClient
- Don't use deprecated junit.framework.Assert.*

Change-Id: I1f44719e6e597a59dd57a815c0357baf37197d91

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -63,7 +63,7 @@
         <dependency>
             <groupId>com.sonymobile.tools.gerrit</groupId>
             <artifactId>gerrit-events</artifactId>
-            <version>2.7.1</version>
+            <version>2.9.0</version>
             <!-- New source is here: https://github.com/sonyxperiadev/gerrit-events -->
         </dependency>
         <dependency>


rsandell:
I don't see any actual use of the 2.9.0 features, so I don't really see a need to upgrade.

Also, as noted in the pr for ```Don't use deprecated DefaultHttpClient``` please test that it works on a standalone Jenkins (not only with hpi:run)

dpursehouse:
> I don't see any actual use of the 2.9.0 features, so I don't really see a need to upgrade.

I intend to add follow-up commits to add support for the new events.

> Also, as noted in the pr for Don't use deprecated DefaultHttpClient please test that it works on a standalone Jenkins (not only with hpi:run)

Will do.

rsandell:
2.9.1 was added in #249 so this can be closed

PR #242, opened 2015-09-01
Fix eclipse warnings

dpursehouse:


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -172,8 +172,6 @@
     private String serverName;
     private String gerritSlaveId;
     private List<PluginGerritEvent> triggerOnEvents;
-    @Deprecated
-    private transient boolean allowTriggeringUnreviewedPatches;
```
rsandell: :-1: This will break loading of old configurations.

dpursehouse: Fixed in the follow up commit cb3c5b565773c9c0ab8032b86c9e68a801d1211f.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -302,7 +302,7 @@ public GerritTrigger(
         this.triggerConfigURL = triggerConfigURL;
         this.gerritTriggerTimerTask = null;
         triggerInformationAction = new GerritTriggerInformationAction();
-        this.allowTriggeringUnreviewedPatches = allowTriggeringUnreviewedPatches;
+        this.allowTriggeringUnreviewedPatches = false;
```
rsandell: just remove this line completely and put ```@SuppressWarnings("unused")``` on the declaration instead.

dpursehouse: Done with 32eb41144fd757743a3656325e56f47a3d402187

PR #243, opened 2015-09-01
Do not use deprecated junit.framework.* classes

dpursehouse:
Replace them with org.junit.Assert.*

Also replace net.sf.ezmorph.test.ArrayAssertions.assertEquals
with org.junit.Assert.assertEquals.

Change-Id: I788d15761c3c2a7b4935e669bcc51d8f318f3602


src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsFunctionalTest.java
```diff
@@ -53,10 +53,10 @@
 import static com.github.tomakehurst.wiremock.client.WireMock.get;
 import static com.github.tomakehurst.wiremock.client.WireMock.urlMatching;
 import static com.sonymobile.tools.gerrit.gerritevents.mock.SshdServerMock.GERRIT_STREAM_EVENTS;
-import static junit.framework.TestCase.assertNotNull;
-import static junit.framework.TestCase.assertSame;
-import static junit.framework.TestCase.fail;
-import static net.sf.ezmorph.test.ArrayAssertions.assertEquals;
+import static org.junit.Assert.assertEquals;
```
rsandell: ```assertEquals(String message, Object[] expecteds, Object[] actuals)``` is deprecated use ```assertArrayEquals```

dpursehouse: The only usage of assertEquals is this one at line 165:

```
assertEquals(1, project.getLastCompletedBuild().getNumber());
```

which is actually `assertEquals(int, int)`.

So the import here of `net.sf.ezmorph.test.ArrayAssertions.assertEquals` was wrong.  I'm surprised it wasn't causing other compilation warnings or errors.

PR #244, opened 2015-09-01
Findbugs cleanup

dpursehouse:


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritDynamicUrlProcessor.java
```diff
@@ -40,6 +40,7 @@
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLConnection;
+import java.nio.charset.StandardCharsets;
```
rsandell: This class is Java 1.7, but this plugin is still based on a Jenkins core version that supports Java 1.6. So until we go ```>=1.612``` of Jenkins core we shouldn't use Java 1.7 features.

rsandell:
The build failure was because of ```BackCompat252HudsonTest``` being flaky and hanging as usual.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritDynamicUrlProcessor.java
dpursehouse: Done in 0b44bfc9f50dc74a0e14f6860b3625e03843025d

PR #247, opened 2015-09-02
Fix usage of deprecated Jenkins test classes

dpursehouse:


rsandell:
It could be a reason for using ```new WebClient()``` instead of ```createWebClient()``` I don't remember.

dpursehouse:
I wonder if it's case sensitive?  If I understand the code correctly, the accounts are being created with the names `Bobby` and `Admin`.

Or maybe the new version is returning different page content and the element is named something else?

I'll look into it some more tomorrow.  It's already late here.

dpursehouse:
> I wonder if it's case sensitive? If I understand the code correctly, the accounts are being created with the names `Bobby` and `Admin`.

This is wrong.  The ones I was referring to here are used in different tests.

> Or maybe the new version is returning different page content and the element is named something else?

This is the problem.  Previously the `href` was named `/gerrit_manual_trigger`, but now it has a `/jenkins` prefix.

I have a fix.  Will push it later today when I've finished testing.

dpursehouse:
This is strange.  The tests are passing when I run them in Eclipse, but not when run on the command line with Maven.

rsandell:
Different JDK on command line vs eclipse? You should run 1.7.0_68 at least but not 1.8

PR #248, opened 2015-09-07
Add a .mailmap file

dpursehouse:
There are some contributors who have made commits with different
email addresses and/or different formatting of their name.

Adding a .mailmap with entries for these contributors allows to
collate all their commits together in the shortlog, i.e.:

  git shortlog -n -c -s

Note that there are also several contributors whose commits are done
with only a github username rather than a full name. This commit does
not add mailmap entries for those users.

Change-Id: I3ec5769df76642ecdbd0cb58506563a465b4a3a7


rsandell:
So what would be the effect of this (except for ```git shortlog```)?

Because normally I intentionally commit under different e-mail to distinguish what I do for work and what I do privately.

dpursehouse:
As far as I am aware it's only used for the [presentation of authors in the shortlog](https://www.kernel.org/pub/software/scm/git/docs/git-shortlog.html#_mapping_authors). 

The output of `git log` etc is not changed.

PR #251, opened 2015-09-29
Permit use of Forbidden files without File Paths

scoheb:
When a job is configured to be triggered with the use of a Forbidden path,
this path is only inspected if the configuration ALSO contains a File path.

This patch corrects this requirement.

[JENKINS-29663]

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
@@ -818,8 +818,12 @@ public class GerritTrigger extends Trigger<Job> {
                          && p.isInteresting(changeBasedEvent.getChange().getProject(),
                                             changeBasedEvent.getChange().getBranch(),
                                             changeBasedEvent.getChange().getTopic())) {
-                        if (isFileTriggerEnabled() && p.getFilePaths() != null
-                                && p.getFilePaths().size() > 0) {
+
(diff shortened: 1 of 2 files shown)


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -818,8 +818,7 @@ public boolean equals(Object obj) {
                          && p.isInteresting(changeBasedEvent.getChange().getProject(),
                                             changeBasedEvent.getChange().getBranch(),
                                             changeBasedEvent.getChange().getTopic())) {
-                        if (isFileTriggerEnabled() && p.getFilePaths() != null
-                                && p.getFilePaths().size() > 0) {
+                        if (isFileTriggerEnabled()) {
```
rsandell: This scares me a little, because it means that there will be a callback to query files for each change-based event that enters Jenkins instead of if there is interest in files. (Indicated by the extra ssh commands you needed to mock in the tests).
I think you can get the same effect if you just also check ```p.getForbiddenFilePaths()```.

rsandell:
Just tickled the gh pull request builder to try again by closing and reopening.

rsandell:
The build failure seems to be a remoting issue in Jenkins and not due to this change, the maven build itself succeeded.

scoheb:
Good point. I will rework this one.

scoheb:
Same remoting issue causing failure?

rsandell:
Yes, same issue it seems.

PR #252, opened 2015-10-01
Option to permit disable Strict Forbidden files

scoheb:
Add option to gerrit project to permit the disable of the default strict forbidden file verification of an event.

Enabling this option will allow an event to trigger a build if the event contains BOTH one or more wanted file paths AND one or more forbidden file paths.

In other words, with this option, the build will not get triggered if the change contains only forbidden files, otherwise it will get triggered.

[FIXED JENKINS-30620]

Change-Id: Icb84111c82a38db722ec09bf7538441bc1b9424a

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/GerritProject.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/GerritProject.java
@@ -54,6 +54,7 @@ public class GerritProject implements Describable<GerritProject> {
     private List<FilePath> filePaths;
     private List<Topic> topics;
     private List<FilePath> forbiddenFilePaths;
+    private boolean disableStrictForbiddenFileVerification;
 
     /**
      * Default empty constructor.
@@ -70,6 +71,7 @@ public class GerritProject implements Describable<GerritProject> {
      * @param topics the topic-rules
      * @param filePaths the file-path rules.
      * @param forbiddenFilePaths the forbidden file-path rules.
+     * @param disableStrictForbiddenFileVerification whether to be strict or not.
      */
     @DataBoundConstructor
     public GerritProject(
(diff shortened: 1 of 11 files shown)


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/GerritProject.java
```diff
@@ -195,16 +215,33 @@ public void setForbiddenFilePaths(List<FilePath> forbiddenFilePaths) {
...
                     if (isInterestingTopic(topic) && isInterestingFile(files)) {
-                        return true;
+                        foundInterestingTopicOrFile = true;
+                    }
+                    if (disableStrictForbiddenFileVerification) {
+                        // Here we want to be able to trigger a build if the event contains
+                        // wanted topics or file paths even though there may be a forbidden file
+                        return foundInterestingTopicOrFile;
+                    } else {
+                        if (foundInterestingForbidden && foundInterestingTopicOrFile) {
```
rsandell: IIUC

```java
if (foundInterestingForbidden) {
    // we have a forbidden file and a wanted file path.
    return false;
} else if (foundInterestingTopicOrFile) {
    // we DO not have a forbidden file and but we have a wanted file path.
    return true;
}
```
would have the same logic but perhaps easier to read?

scoheb: Yep. You are right. Too many hours in front of a screen :) Tx

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/GerritProject.java
```diff
@@ -78,7 +80,8 @@ public GerritProject(
             List<Branch> branches,
             List<Topic> topics,
             List<FilePath> filePaths,
-            List<FilePath> forbiddenFilePaths) {
+            List<FilePath> forbiddenFilePaths,
+            boolean disableStrictForbiddenFileVerification) {
```
rsandell: Note to future selves; we should stop putting things into the constructor and start using ```@DataboundSetter```s instead.

rsandell:
Are you in a hurry to get this released or can I wait for something else to drop in?
Version numbers are cheap, so I have no problems with releasing this anyway, but three new versions in September is quite an awesome set of releases ;)

PR #254, opened 2015-10-07
Add Replication status to queued item and failures

scoheb:
This change adds the replication status to the reason for the fact
that an item is in the queue.

For example, if a replication is failing towards a mirror, its status will added as shown below:

(pending—Replication to sekimirror (is failing) not completed.)

In addition, if the replication has timed-out, the status will be displayed in the console log
as shown below:

ERROR: Waiting for replication of refs/changes/55/899655/1 to sekimirror timed out.

Replication Status:
-------------------

** Replication to sekimirror is failing!

Change-Id: I254d006629ddae375a3b95ad9a1b137c4a8006be

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/GerritSlave.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/GerritSlave.java
@@ -51,6 +51,9 @@ public class GerritSlave extends AbstractDescribableImpl<GerritSlave> {
     private String host;
     private int timeoutInSeconds;
 
+    //Hold the replication status
+    private transient ReplicationStatus replicationStatus;
+
     /**
      * Standard Constructor.
      * @param id the ID or {@code null} to generate a new one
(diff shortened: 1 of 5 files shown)


scoheb:
I may have some additional changes for this...can we put this on-hold temporarily?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/GerritSlave.java
```diff
@@ -112,6 +115,55 @@ public int getTimeoutInSeconds() {
...
+
+    /**
+     * Possible replication status.
+     */
+    public enum ReplicationStatus {
+
+        /**
+         * Replication is in progress.
+         **/
+        INPROGRESS("in progress"),
```
rsandell: Try to make these texts I18n

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/replication/ReplicationFailedAction.java
```diff
@@ -47,6 +52,19 @@ public ReplicationFailedAction(String reason) {
      * @return the reason
      */
     public String getReason() {
-        return reason;
+        StringBuffer slavesReason = new StringBuffer("\n\nReplication Status:");
+        slavesReason.append("\n-------------------\n\n");
+
+        if (blockedItem.getSlavesWaitingFor() != null && blockedItem.getSlavesWaitingFor().size() > 0) {
+            Iterator<String> it = blockedItem.getSlavesWaitingFor().keySet().iterator();
+            while (it.hasNext()) {
```
rsandell: seems like you could just as easily do a for each loop over ```blockedItem.getSlavesWaitingFor().values()```

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/replication/ReplicationFailedAction.java
```diff
@@ -32,14 +37,14 @@
  */
 public class ReplicationFailedAction extends InvisibleAction {
 
-    private String reason;
+    private BlockedItem blockedItem;
```
rsandell: You should probably make BlockedItem serializable just in case, if nothing else just to mark that it will/could be serialized at some point.

rsandell:
Sure, I can wait :)

I just took a quick glanse over the code and provided some suggestions ;)

scoheb:
not needed right now. I will re-visit this at a later time.

PR #255, opened 2015-10-22
Allow override of code-review/verified value from job

scoheb:
Currently, it is NOT possible to have one Jenkins job setting only the Verified label to +1 or -1
and another Jenkins job setting only code-review label to +1 or -1 for the same patch-set in Gerrit.

This change will now allow:

- Setting a code-review or verified value to empty for the Server default value
- Setting a job specific value for a code-review or verified value
- Only the jobs that have overridden the value in their configuration will contribute to the code-review or verified value.

[JENKINS-30367] [JENKINS-30393]


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ParameterExpander.java
```diff
@@ -272,6 +272,11 @@ private String expandParameters(String gerritCommand, Run r, TaskListener taskLi
         for (Map.Entry<String, String> param : parameters.entrySet()) {
             gerritCommand = gerritCommand.replace("<" + param.getKey() + ">", param.getValue());
         }
+        //replace null and Integer.MAX_VALUE code review value
```
rsandell: nasty, but understandable ;)

scoheb: I totally agree...I am been cringing since I submitted this PR ;)

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ParameterExpander.java
```diff
@@ -365,7 +370,7 @@ protected int getVerifiedValue(Result res, GerritTrigger trigger) {
      * @return the lowest verified value.
      */
     public int getMinimumVerifiedValue(MemoryImprint memoryImprint, boolean onlyBuilt) {
```
rsandell: missed changing to ```Integer``` here

rsandell: If possible, to be more correct I think this method should return ```null``` if no minimum value is found at all. It would be more correct than returning ```Integer.MAX_VALUE``` that is definitely not a number that was found.

The method should also be annotated with ```CheckForNull``` to help findbugs, and the javadoc should be updated to note the new behaviour.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ParameterExpander.java
```diff
@@ -398,8 +402,8 @@ public int getMinimumVerifiedValue(MemoryImprint memoryImprint, boolean onlyBuil
      * @param onlyBuilt only count builds that completed (no NOT_BUILT builds)
      * @return the lowest code review value.
      */
-    public int getMinimumCodeReviewValue(MemoryImprint memoryImprint, boolean onlyBuilt) {
-        int codeReview = Integer.MAX_VALUE;
+    public Integer getMinimumCodeReviewValue(MemoryImprint memoryImprint, boolean onlyBuilt) {
```
rsandell: same here if possible.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/config/Config.java
```diff
@@ -305,36 +281,25 @@ public void setValues(JSONObject formData) {
...
-        gerritBuildUnstableCodeReviewValue = formData.optInt(
-                "gerritBuildUnstableCodeReviewValue",
-                DEFAULT_GERRIT_BUILD_UNSTABLE_CODE_REVIEW_VALUE);
-        gerritBuildNotBuiltVerifiedValue = formData.optInt(
-                "gerritBuildNotBuiltVerifiedValue",
-                DEFAULT_GERRIT_BUILD_NOT_BUILT_VERIFIED_VALUE);
-        gerritBuildNotBuiltCodeReviewValue = formData.optInt(
-                "gerritBuildNotBuiltCodeReviewValue",
-                DEFAULT_GERRIT_BUILD_NOT_BUILT_CODE_REVIEW_VALUE);
+        if (formData.isEmpty()) {
```
rsandell: Why the difference in handling verified vs code review values?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ParameterExpander.java
scoheb: ok will do.

src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer/index.jelly
```diff
@@ -113,35 +113,35 @@
                         <f:textbox name="gerritBuildStartedVerifiedValue"
                                    value="${it.config.gerritBuildStartedVerifiedValue}"
                                    default="${com.sonyericsson.gerrithudsontrigger.config.Config.DEFAULT_GERRIT_BUILD_STARTED_VERIFIED_VALUE}"
```
rsandell: I don't think all of these default values are possible any more.

I'm going to add some inherital sin on you here and ask you if you can clean the default values up a bit, because they have probably never really worked. I doubt any of them are useful and maybe could just be removed since a new server config should be instantiated with default values from the beginning.

src/main/webapp/help-GerritBuildFailedCodeReview.html
```diff
@@ -1 +1,3 @@
 The default Code Review vote that should be set when a build fails.
+<br><br>
+If this field is left as an empty string, then any job's configured value will have precedence.
```
rsandell: Maybe should also add what the result is if no job has it configured as well.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/config/Config.java
scoheb: I guess our view was that the new default/standard for Code Review values should be that the Server values should start off as empty for a new server. 

Most users associate build health with +1/-1 for Verified and not Code Review. In other words, sending Code Review values should be disabled by default. 

Does that sounds reasonable?

src/main/webapp/help-GerritBuildFailedCodeReview.html
scoheb: will do.

src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer/index.jelly
scoheb: you mean the textbox default values? Yes. I will clean it up.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/config/Config.java
rsandell: Well we could look at it another way that gerrit does not come configured with the Verified label out of the box, so having nothing set for verified and only sensible defaults for Code Review should be the proper way to go for new users :)

rsandell: Then we would remove the obstacle/bug of not being able to run the plugin on a brand spanking new Gerrit server :)

src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer/index.jelly
rsandell: Yes the textbox defaults. You should maybe also look over the defaults that are set when a new server config is created.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/config/Config.java
scoheb: How about we compromise? We keep the defaults for both Verified and CodeReview?

This way it will work for new Gerrit server installs, but it will not break the environments (company starts with an 'E' ;) ) that expect default Verified values.

rsandell:
Now that was code I feel I haven't seen in a long time :)

Great work!

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/config/Config.java
rsandell: I don't think we would need to compromise if the variables are treated equally.
```formData.isEmpty``` should only happen on the first creation of a new default server. If that only creates sensible default values for Code Review I don't think it would matter to you as you only need to configure that once, any other server could then just be copied from a previous server config with the correct values for your setup.
Any existing configurations should "just work" with the previous values when you start up with the new version.

Or am I misinterpreting this section of code completely?

Comment (scoheb) [inline src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/config/Config.java]: I think we are saying the same thing...

We make use of the `forData.isEmpty` construct for BOTH Verified and CodeReview cases.

**Comment by scoheb** [comment]
@rsandell ... Can you please take a look now?

**Comment by rsandell** [comment]
Will do

File: src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/config/Config.java (hunk 4 above)
Comment (rsandell) [inline src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/config/Config.java]: I was more thinking that the default values for all verified should be `null` so they would work better with a default Gerrit setup.

File: src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ParameterExpanderSkipVoteParameterTest.java [hunk 7]
```diff
@@ -70,8 +70,12 @@ public ParameterExpanderSkipVoteParameterTest(TestParameter parameter) {
     public void testCodeReview() {
         IGerritHudsonTriggerConfig config = Setup.createConfig();
         ParameterExpander instance = new ParameterExpander(config);
-        int result = instance.getMinimumCodeReviewValue(parameter.memoryImprint, true);
-        assertEquals(parameter.expectedCodeReview, result);
+        Integer result = instance.getMinimumCodeReviewValue(parameter.memoryImprint, true);
+        if (parameter.expectedCodeReview == null) {
+            assertEquals(null, result);
```
rsandell: `assertNull(result);` would be more descriptive I guess.

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ParameterExpanderSkipVoteParameterTest.java
```diff
@@ -82,8 +86,12 @@ public void testCodeReview() {
     public void testVerified() {
         IGerritHudsonTriggerConfig config = Setup.createConfig();
         ParameterExpander instance = new ParameterExpander(config);
-        int result = instance.getMinimumVerifiedValue(parameter.memoryImprint, true);
-        assertEquals(parameter.expectedVerified, result);
+        Integer result = instance.getMinimumVerifiedValue(parameter.memoryImprint, true);
+        if (parameter.expectedVerified == null) {
+            assertEquals(null, result);
```
rsandell: `assertNull(result);` would be more descriptive I guess.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/config/Config.java
scoheb: Do you think we could keep the setup for default values as it was prior to this PR?

I can create a new PR that could address correcting the default value setup to be similar to Gerrit.

rsandell: Ok

scoheb:
I fixed the assertNulls. Please review/merge at your convenience.

rsandell:
OK, I'll merge as soon as Jenkins says the build is OK.

Should I wait with release until #254 and the "new defaults" are also merged?

scoheb:
I just closed #254 .

I would prefer a release BEFORE the "new defaults".

rsandell:
ok, releasing asap ;)

PR #258, opened 2015-11-24
Refactored the form posting

rsandell:
Replaced most values with DataBoundSetter to make future expansions easier.

Replaced noNameAndEmailParameters and readableMessage with ParameterMode settings so users can choose between Plain, Encoded or No parameter for those parameters instead.

@reviewbybees


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -238,39 +250,18 @@
...
+            Integer gerritBuildStartedCodeReviewValue, Integer gerritBuildSuccessfulVerifiedValue,
+            Integer gerritBuildSuccessfulCodeReviewValue, Integer gerritBuildFailedVerifiedValue,
+            Integer gerritBuildFailedCodeReviewValue, Integer gerritBuildUnstableVerifiedValue,
+            Integer gerritBuildUnstableCodeReviewValue, Integer gerritBuildNotBuiltVerifiedValue,
+            Integer gerritBuildNotBuiltCodeReviewValue, boolean silentMode, boolean silentStartMode,
+            boolean escapeQuotes, boolean noNameAndEmailParameters, boolean readableMessage, String dependencyJobsNames,
+            String buildStartMessage, String buildSuccessfulMessage, String buildUnstableMessage,
+            String buildFailureMessage, String buildNotBuiltMessage, String buildUnsuccessfulFilepath, String customUrl,
+            String serverName, String gerritSlaveId, List<PluginGerritEvent> triggerOnEvents,
+            boolean dynamicTriggerConfiguration, String triggerConfigURL, String notificationLevel) {
```
jglick: That is the kind of diff hunk I love to see!

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -238,39 +250,18 @@
      * @param triggerConfigURL               Where to fetch the configuration file from
      * @param notificationLevel              Whom to notify.
      */
-    @DataBoundConstructor
-    public GerritTrigger(
-            List<GerritProject> gerritProjects,
-            SkipVote skipVote,
```
jglick: Remember to set default values for these fields.

rsandell:
The change to enum.jelly is in a pull request to core: https://github.com/jenkinsci/jenkins/pull/1926

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -172,18 +173,53 @@
...
+            DescriptorImpl descriptor = (DescriptorImpl)getDescriptor();
+            if (descriptor != null) {
+                ListBoxModel options = descriptor.doFillNotificationLevelItems(this.serverName);
+                if (!options.isEmpty()) {
+                    this.notificationLevel = options.get(0).value;
+                }
+            }
+            //CS IGNORE EmptyBlock FOR NEXT 1 LINES. REASON: Handled one row below
+        } catch (NullPointerException ignored) { /*Could happen during testing*/ }
+        if (this.notificationLevel == null) {
```
jtnord: would it not make more sense to set the default values in the field definition rather than having to put them here?

rsandell: They are also set in the field definition, this is just some extra security to get sensible defaults in case the object is created outside of form posting.

rsandell:
Just tickling the pull request builder

PR #259, opened 2015-11-25
Correct Missed Events Playback manager init

scoheb:
It is possible for the Missed Events playback manager to not be correctly initialized
once the REST API has been enabled AFTER the connection has been started.

The same issue applies if the Gerrit events-log plugin was installed AFTER the
Gerrit connection was started.

This ensures that the listener is active so that playback can execute when required.

[FIXED JENKINS-31439]


rsandell:
Should I release this fix directly or can it wait for #258 ?

scoheb:
Since it is a bug fix, it would be nice if it could be released now.

Tx

PR #260, opened 2015-11-27
Project list is not updating in desired interval

scoheb:
Fix bug whereby project list updating was not active upon connection startup.

Only applies for Gerrit servers < 2.12.

[FIXED JENKINS-31473]

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritProjectListUpdater.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritProjectListUpdater.java
@@ -37,6 +37,7 @@ import java.io.Reader;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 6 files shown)


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritProjectListUpdater.java
```diff
@@ -140,6 +138,10 @@ public synchronized void shutdown() {
 
     @Override
     public void run() {
+        //Ensure connected
+        while (!connected) {
```
rsandell: It might be safer to make connected an AtomicBoolean

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/mock/MockConfigForProjectListTest.java
```diff
@@ -0,0 +1,480 @@
...
+
+import java.io.File;
+import java.util.LinkedList;
+import java.util.List;
+
+import net.sf.json.JSONObject;
+
+import org.apache.http.auth.Credentials;
+
+/**
```
rsandell: Would it be possible to just extend whatever class you copied this from and override the particular values that you need? It seems a bit ugly/unnecessary to bloat the tests like this.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritProjectListUpdater.java
scoheb: will do

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/mock/MockConfigForProjectListTest.java
scoheb: will do

scoheb:
Seems like the build is hanging on a test. Could you please abort it? I will debug it.

rsandell:
It didn't hang on the standard backcompat tests at least :)

scoheb:
I have another PR on its way for another bug. So this PR can be rolled into your next planned release at your convenience!

PR #261, opened 2015-11-27
Protect PlaybackManager from null EventCreatedOn

scoheb:
It is possible that Gerrit might return a null eventCreated attribute for some events.

This protects the persistence of events from this situation.

[FIXED JENKINS-30975]


rsandell:
:+1: Just waiting for the tests to pass.

PR #269, opened 2015-12-02
Rename category deprecated default names CRVW and VRIF

engycz:
Rename category deprecated default names CRVW and VRIF
removed in Gerrit 2.6 (2 years old) to current
label names Code-Review and Verified

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -8,7 +8,7 @@
 
     <groupId>com.sonyericsson.hudson.plugins.gerrit</groupId>
     <artifactId>gerrit-trigger</artifactId>
-    <version>2.18.0</version>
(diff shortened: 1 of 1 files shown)


scoheb:
Can we not just add the "new" ones and keep the legacy ones? I guess it may be possible that someone is still using Gerrit 2.6?

engycz:
Default values concerns only to newly created servers in gerrit trigger. Copied servers also copies verdict categories from the parent. So this change is mainly for new Jenkins users with >= 2.6 Gerrit server.
CRVW and VRIF were last time in Gerrit 2.5.6

Merge as you wish, it is not a bug.

scoheb:
+1 for me

PR #273, opened 2015-12-22
Fix projectListRefreshInterval zero value after upgrade from 2.13.0 to 2.14.0

engycz:
Fix projectListRefreshInterval zero value after upgrade from 2.13.0 to 2.14.0

[FIXED JENKINS-31800]

### DIFF
--- a/src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer/index.jelly
+++ b/src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer/index.jelly
@@ -251,7 +251,7 @@
                                      <f:textbox name="projectListFetchDelay"
                                         value="${it.config.projectListFetchDelay}"
                                         default="${com.sonyericsson.gerrithudsontrigger.config.Config.DEFAULT_PROJECT_LIST_FETCH_DELAY}"
-                                        checkUrl="'${rootURL}/${serverURL}/projectListFetchDelayCheck?value='+escape(this.value)"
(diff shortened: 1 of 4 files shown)


src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer/index.jelly
```diff
@@ -263,7 +263,7 @@
                                      <f:textbox name="projectListRefreshInterval"
                                         value="${it.config.projectListRefreshInterval}"
                                         default="${com.sonyericsson.gerrithudsontrigger.config.Config.DEFAULT_PROJECT_LIST_REFRESH_INTERVAL}"
-                                        checkUrl="'${rootURL}/${serverURL}/projectListFetchDelayCheck?value='+escape(this.value)"
+                                        checkUrl="'${rootURL}/${serverURL}/positiveIntegerCheck?value='+escape(this.value)"
```
rsandell: `GerritServer.DescriptorImpl.doProjectListFetchDelayCheck()` should be removed if we don't plan to use it any more which it seems like.

rsandell:
Back from vacation, sorry for the delay in review.

src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer/index.jelly
engycz: Done

PR #276, opened 2016-03-08
Don't used ReflectiveOperationException

dpursehouse:
Using ReflectiveOperationException breaks Java 6 runtime compatibility.

[FIXED JENKINS-30857]

Change-Id: Ic8c88102cf9ef3a3286f5311d1b3ed661ac2d77c


dpursehouse:
@rsandell it seems I broke Java6 compatibility with one of my patches last year, and I only found out about it today :(

rsandell:
@dpursehouse Yes, maybe we should add an animal-sniffer config in the pom just as I did in gerrit-events to avoid this in the future?

dpursehouse:
@rsandell done, see PR #277.

PR #277, opened 2016-03-08
Add animal-sniffer-maven-plugin to check for Java 1.6 compatibility

dpursehouse:
Change-Id: I5d664b12e856b5d0b01a740697295abe684670a2

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -336,7 +336,7 @@
             <plugin>
                 <groupId>org.codehaus.mojo</groupId>
                 <artifactId>animal-sniffer-maven-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


rsandell:
I have a variant configuration prepared as well that might not break the build.

rsandell:
For some reason it isn't picking up the 1.6 signature but uses 1.5 instead. I managed to fix that in my config.

rsandell:
Superseded by 24cb235d3c9504667387edf4274d3e00e30bfbf4

rsandell:
@dpursehouse thanks for the kick :D

PR #278, opened 2016-03-30
Don't crash because JENKINS-33467 has changed expected behaviour

rsandell:
@reviewbybees esp @jglick


jtnord:
:bee:

rsandell:
Failed because of a build timeout, that happens sometimes and not related to this pr.

jglick:
Not sure why `cleanUpGerritCauses` exists at all. For _any_ kind of cause, there could be multiple triggers before a build starts. (And not limited to quiet period—the queue may simply be saturated.) Would seem better to simply delete this method.

rsandell:
@jglick The method is a safety net, a long time ago in a Jenkins version far far away we noticed jobs being triggerd multiple times for the same event (something that shouldn't happen) And this method was a workaround for there to not be multiples of the same cause for the same event on the build page.

PR #281, opened 2016-04-28
Adapt to new plugin parent pom

rsandell:
This makes it easier to test on newer core versions.

Besides the pom mostly javadoc and jelly fixes but usage of HtmlUnit needed to be tweaked as well.

When playing around to find a fix for the failing configRoundTrip test on workflow I needed to change the access of Queue.Item.id that indicated a potential hidden bug, so I decided to bump dependency to 1.609.3 and fix it.

@reviewbybees 
@TWestling


rsandell:
Hmm, it is a bit disturbing to me that the tests now timeout and runs much slower. They pass on my machine, but runs slower, and only on jdk8. On jdk7 they run out of perm gen.
Seems like the ci builds are having a hard time even when running on an m1.large.

rsandell:
Hmm again, the timeouts could be due to `com.jcraft.jsch.JSchException: Algorithm negotiation fail`, but this is java running an ssh command to the same java process. So I guess there are no available algorithms at all? When did this happen?

rsandell:
Ok, so as it stands now, running with jdk8 fails due to the above algorithm exchange error. And jdk 7 fails due to not enough memory. Will see if I can turn off reuse forks et.al. to get he memory consumption down.

pom.xml
```diff
@@ -102,15 +124,21 @@
             <version>2.0</version>
         </dependency>
         <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>junit</artifactId>
+            <version>1.12</version>
+            <scope>test</scope>
```
jglick: Why? All plugins get this dep anyway.

pom.xml
```diff
@@ -102,15 +124,21 @@
...
             <artifactId>workflow-aggregator</artifactId>
-            <version>1.4</version>
+            <version>${workflow.version.test}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-step-api</artifactId>
-            <version>1.4</version>
+            <version>${workflow.version}</version>
```
jglick: A little dangerous, but OK for now.

pom.xml
```diff
@@ -298,25 +290,6 @@
                 </executions>
```
jglick: `build-helper-maven-plugin` configuration looks like it might be unnecessary now, or could be moved to parent POM.

src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritManagement/index.jelly
```diff
@@ -1,3 +1,4 @@
+<?jelly escape-by-default='true'?>
```
jglick: Remember that this change can cause regressions. If the original file had

```xml
${it.someUntaintedDataWhichMightContainHTML}
```

you need to switch it to

```xml
<j:out value="${it.someUntaintedDataWhichMightContainHTML}"/>
```

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/mock/Setup.java
```diff
@@ -817,10 +817,8 @@ public static RefReplicated createRefReplicatedEvent(String project, String ref,
     public static void lockDown(JenkinsRule j) throws Exception {
         SecurityRealm securityRealm = j.createDummySecurityRealm();
         j.getInstance().setSecurityRealm(securityRealm);
-
-        GlobalMatrixAuthorizationStrategy authorizationStrategy = new GlobalMatrixAuthorizationStrategy();
-        authorizationStrategy.add(Hudson.READ, "authenticated");
-        j.getInstance().setAuthorizationStrategy(authorizationStrategy);
+        j.getInstance().setAuthorizationStrategy(
+                new MockAuthorizationStrategy().grant(Hudson.READ).everywhere().toAuthenticated());
```
jglick: Glad to see this get used!

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spec/DuplicateGerritListenersPreloadedProjectHudsonTestCase.java
```diff
@@ -152,7 +152,7 @@ public void testReconfigureUsingRestApi() throws Exception {
         Document document = xmlPage.getXmlDocument();
         String xml = changeConfigXml(gerritProjectPattern, document);
         URL url = UrlUtils.toUrlUnsafe(j.getURL().toExternalForm() + testProj.getUrl() + "config.xml");
-        WebRequestSettings request = new WebRequestSettings(url, HttpMethod.POST);
+        WebRequest request = new WebRequest(url, HttpMethod.POST);
         request.setRequestBody(xml);
         j.jenkins.setCrumbIssuer(null);
```
jglick: :ant: Better to use `getCrumbedUrl`.

jglick:
:bee:

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/mock/Setup.java
rsandell: I remembered it from a code review some time back, so I knew it was available :)

src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritManagement/index.jelly
rsandell: Yes, I took a quick scan through the files that I was changing, the only worry I have is how the data binding works for form values, e.g. if a textfield value contains `"<"` or `">"`?

pom.xml
rsandell: Doesn't seem so, the plugin wasn't loaded when the unit tests ran, so I had to add it explicitly.

src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritManagement/index.jelly
jglick: That is fine. You only need to look for cases where the view _intends_ to display HTML content bound to some model object attribute (`something.detailMessage`, etc.), where the content is known be untainted by user input. Adding `escape-by-default='true'` without switching that usage to `j:out` will cause the content to be rendered as plain text rather than markup, breaking the UI.

pom.xml
jglick: Ah, `org.jenkins-ci.plugins:junit` not `junit:junit`. Ignore me.

rsandell:
So the tests are working again locally for me, and hopefully surefire is tweaked enough so that it can run on an old m1.large, waiting on Jenkins to see if it is so.

rsandell:
Yay! All the tests are passing on CI :dancer:

rsandell:
@TWestling do you want to take a look first or can I merge without mercy?

rsandell:
Ping @TWestling again for testing's sake

PR #283, opened 2016-05-16
Set new environment variable for a change-merged event.

Jamie-sas:
JIRA JENKINS-23873

This change can only be applied after pull request [#54](https://github.com/sonyxperiadev/gerrit-events/pull/54) on gerrit-events project is applied, otherwise an error about a missing method will be seen.


Jamie-sas:
In case someone needs a bit more info...

This change makes the merged SHA-1 available, as well as the plain patchset SHA-1 for a change that caused a merge.

We were running into problems where a Jenkins validation build of a Gerrit submit would work, but a build after the patch was merged would fail. This was due to the patch itself being fine, but not working with another patch it merged with.

In order to combat the problem, we wanted to do the build based on the merged source, but that value was unavailable to Jenkins due to this jar not surfacing the revision... it was in the event stream from Gerrit, but the jar didn't have any code to parse if from the JSON and supply it upstream.

So, I fixed that, and made the Gerrit Trigger plugin create a new env var based on this value, named "GERRIT_MERGED_REVISION". If no merge occurs (such as a plain fast-forward push), the value will be identical to GERRIT_PATCHSET_REVISION, but in the event of a merge, it is a different value - the SHA-1 of the merged commit.

This allows you to set up Jenkins to build the patchset itself, the merged set, or both, as needed.

Jamie-sas:
GERRIT_MERGED_REVISION could as easily been named GERRIT_NEWREV, to match the JSON value, but there's already a GERRIT_NEWREV which is the patchset revision.  All RefUpdate events have both an oldRev and a newRev, but newRev is the revision of the patchset itself.  The merged revision is a 3rd revision, hence the more explicit name.

PR #284, opened 2016-05-17
Provide GERRIT_MERGED_REVISION after change-merged event

polygon:
The newRev field of a change-merged event was not parsed yet. Since this field holds the hash of a merge commit it can be important in certain build conditions.

This patch exposes this field as a GERRIT_MERGED_REVISION variable to Jenkins builds. It requires the following PR in gerrit-events to compile successfully: https://github.com/sonyxperiadev/gerrit-events/pull/55 (which is the reason why Jenkins reports a failure below)

It fixes https://issues.jenkins-ci.org/browse/JENKINS-23873

Some changes to pom.xml to update the gerrit-events version might be required. Since I'm not normally a Java person I leave that to the maintainer.

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTriggerParameters.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTriggerParameters.java
@@ -37,6 +37,7 @@ import com.sonymobile.tools.gerrit.gerritevents.dto.events.ChangeBasedEvent;
 import com.sonymobile.tools.gerrit.gerritevents.dto.events.ChangeRestored;
 import com.sonymobile.tools.gerrit.gerritevents.dto.events.GerritTriggeredEvent;
 import com.sonymobile.tools.gerrit.gerritevents.dto.events.RefUpdated;
+import com.sonymobile.tools.gerrit.gerritevents.dto.events.ChangeMerged;
 import hudson.model.Job;
 import hudson.model.ParameterValue;
 import hudson.model.StringParameterValue;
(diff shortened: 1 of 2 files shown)


rsandell:
I've updated the pom on master.
You need to update the help file for the trigger as well that lists all the parameters.

rsandell:
Since we already have the parameter `GERRIT_NEWREV` for `RefUpdated` events, I think it's better to reuse that.

polygon:
I like GERRIT_MERGED_REVISION more since it's more clear what it means. However I also see the point of re-using GERRIT_NEWREV due to the naming in the Gerrit event stream.

I added some help text, changed the parameter name and rebased on top of the current master.

polygon:
I hope I'm not pushing on this too much but I wonder when there is going to be a new release including this code.

rsandell:
I wanted the fix for the regression on security 170 merged before I made the next release. And that was merged and released yesterday :).

clabu:
@rsandell Does this mean that if we want to build the merged result of the changeMerged() event, we need to check out the commit specified by GERRIT_NEWREV (or just take HEAD)? We currently are not, and then it looks like the implicit checkout of a Pipeline job is taking the commit of the change before it was merged. Or are we configuring our Pipeline job wrong?

rsandell:
@clabu Yes, if you use the gerrit build chooser ("choose what revision to build" I believe it says in the UI) it will pick the revision from the active patchset for all change based events, and for `RefUpdated` it picks the newRev.
So for change merged events, if you want to use the merge commit you will need a different scm configuration.

rsandell:
OTOH it also wouldn't be too hard to add an optional setting to [GerritTriggerBuildChooser](https://github.com/jenkinsci/gerrit-trigger-plugin/blob/master/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTriggerBuildChooser.java) to make it pick the merge rev for change-merged instead of the generic rev. ;)

PR #288, opened 2016-07-05
JENKINS-36139: JGit4 RevWalk is now a Closeable so use RevWalk.close()

bfh-reseiwe:

PR #291, opened 2016-07-06
JENKINS-36139: Use try-with-resource for RevWalk in JGit4

bfh-reseiwe:
Supersedes pull request #288.

This resolves but JENKINS-36139.

JGit4 is targeted towards java 7, and RevWalk is now a Closeable, hence the release() method is renamed to close(). Use the new try-with-resource syntax for RevWalk.

@rsandell

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -52,9 +52,9 @@
     </developers>
 
     <properties>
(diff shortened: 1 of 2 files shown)


rsandell:
:+1: just waiting for Jenkins to provide the build result before merging.

PR #294, opened 2016-07-13
operator '^' for dynamic trigger must be escaped in regex

mbudnick:
If you use "Dynamic Trigger Configuration" with line "b^**" there will be an exception "cannot parse" in GerritDynamicUrlProcessor line 154.
Operator "^" must be escaped if used in regex pattern.

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritDynamicUrlProcessor.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritDynamicUrlProcessor.java
@@ -95,7 +95,7 @@ public final class GerritDynamicUrlProcessor {
         if (!firstoperator) {
           operators.append("|");
         }
-        operators.append(type.getOperator());
+        operators.append(regexEscapted(type.getOperator()));
         firstoperator = false;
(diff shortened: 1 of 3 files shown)


rsandell:
Could you provide a unit test?

mbudnick:
I never wrote tests with mockito and powermock, I use jmockit. So I don't know how to mock url.openConnection() or test private methods.
There are also errors in test-classes and pom.xml when I import the project in my Eclipse and tests don't run.
So I can't provide a unit test.

rsandell:
You don't need to do any of that mocking or url connection stuff. Just copy and slightly modify this one https://github.com/jenkinsci/gerrit-trigger-plugin/blob/master/src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spec/SpecGerritTriggerHudsonTest.java#L217
And you can run `mvn test -Dtest=SpecGerritTriggerHudsonTest` on the commandline to verify.

mbudnick:
I wrote a unit test for "b^**" (which was my problem). How can I add this to this pull request? Should I amend my first commit to include the test?

rsandell:
Just push a new commit to `mbudnick:master` no need to ammend

rsandell:
Thanks!

PR #295, opened 2016-07-13
[JENKINS-30821] Add comment-added comment as job parameter

teemumurtola:
Expose the comment posted to Gerrit on a comment-added event as a job parameter named GERRIT_CHANGE_COMMENT_TEXT. Use the option for commit message encoding also for this parameter for now. If desired, I can look into introducing a separate option for this.

Disclaimer: no tests done due to lack of a suitable environment, but the change should be mostly trivial and most other functionality in GerritTriggerParameters does not seem to be covered by tests, either.


rsandell:
It should be fairly trivial to add a separate setting for it, so let's do it.

There are plenty of tests for various parameters that you can piggyback on, just do a usage search on `GerritTriggerParameters`, for example `ParameterModeJenkinsTest`.

You should also update the [help file](https://github.com/jenkinsci/gerrit-trigger-plugin/blob/master/src/main/webapp/help-whatIsGerritTrigger.html) to mention the new parameter as well.

teemumurtola:
Added a separate option (tested that it appears correctly on the configuration page), updated the help text, and added tests similar to what existed for the commit message parameter.

rsandell:
Thanks!

PR #298, opened 2016-09-10
Add a silent level job setting

zaro0508:
Add a silent level parameter to allow Jenkins to keep silent on votes
only, on comments only or keep silent on all feeback.

This resolves issue JENKINS-38000

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/config/SilentLevel.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/config/SilentLevel.java
@@ -1,0 +1,61 @@
+/*
+ *  The MIT License
+ *
+ *  Copyright 2014 Christian Georgi
+ *
+ *  Permission is hereby granted, free of charge, to any person obtaining a copy
+ *  of this software and associated documentation files (the "Software"), to deal
+ *  in the Software without restriction, including without limitation the rights
+ *  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ *  copies of the Software, and to permit persons to whom the Software is
+ *  furnished to do so, subject to the following conditions:
+ *
+ *  The above copyright notice and this permission notice shall be included in
+ *  all copies or substantial portions of the Software.
+ *
+ *  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ *  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ *  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ *  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ *  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ *  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ *  THE SOFTWARE.
+ */
+package com.sonyericsson.hudson.plugins.gerrit.trigger.config;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Values for Silent Mode.
+ */
+public enum SilentLevel {
+
+    /**
+     * Don't disclose anything.
+     */
+    ALL,
+
+    /**
+     * Keep comments to yourself.
+     */
+    COMMENTS,
+
+    /**
+     * Keep votes to yourself.
+     */
+    VOTES;
+
+
+    /**
+     * Method to get the list of silent levels.
+     * @return A list of silent levels.
+     */
+    public static List<String> getLevels() {
+      List<String> levels = new ArrayList();
+      levels.add(ALL.toString());
+      levels.add(COMMENTS.toString());
+      levels.add(VOTES.toString());
+      return levels;
+    }
+}
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ToGerritRunListener.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ToGerritRunListener.java
@@ -24,6 +24,7 @@
 package com.sonyericsson.hudson.plugins.gerrit.trigger.gerritnotifier;
 
 import com.sonymobile.tools.gerrit.gerritevents.dto.events.GerritTriggeredEvent;
+import com.sonyericsson.hudson.plugins.gerrit.trigger.config.SilentLevel;
 import com.sonyericsson.hudson.plugins.gerrit.trigger.events.lifecycle.GerritEventLifecycle;
 import com.sonyericsson.hudson.plugins.gerrit.trigger.gerritnotifier.model.BuildMemory;
 import com.sonyericsson.hudson.plugins.gerrit.trigger.gerritnotifier.model.BuildsStartedStats;
@@ -129,7 +130,7 @@ public final class ToGerritRunListener extends RunListener<Run> {
             if (event instanceof GerritEventLifecycle) {
                 ((GerritEventLifecycle)event).fireBuildCompleted(r);
             }
-            if (!cause.isSilentMode()) {
+            if (!cause.isSilentMode() || (cause.isSilentMode() && trigger.getSilentLevel() != SilentLevel.ALL)) {
                 memory.completed(event, r);
 
                 Result result = r.getResult();
@@ -153,7 +154,7 @@ public final class ToGerritRunListener extends RunListener<Run> {
                 }
 
                 updateTriggerContexts(r);
-                allBuildsCompleted(event, cause, listener);
+                allBuildsCompleted(r, event, cause, listener);
             }
         }
     }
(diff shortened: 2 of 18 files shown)


src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer/index_ja.properties
```diff
@@ -88,6 +88,8 @@ Gerrit\ Connection\ Setting=\
     Gerrit\u63a5\u7d9a\u8a2d\u5b9a
 Notification\ Level=\
     \u901a\u77e5\u30ec\u30d9\u30eb
+Silent\ Level=\
```
rsandell: No need to add default translations, if a key is missing for the current lang, the default lang will be used. Better to not have them here, so a translator can find what is missing.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -208,6 +210,20 @@ public GerritTrigger(List<GerritProject> gerritProjects) {
...
+          if (descriptor != null) {
+              ListBoxModel options = descriptor.doFillSilentLevelItems();
+              if (!options.isEmpty()) {
+                  this.silentLevel = SilentLevel.valueOf(options.get(0).value);
+              }
+          }
+          //CS IGNORE EmptyBlock FOR NEXT 1 LINES. REASON: Handled one row below
+        } catch (NullPointerException ignored) { /*Could happen during testing*/ }
+        if (this.silentLevel == null) {
+            this.silentLevel = SilentLevel.ALL;
```
rsandell: I'm not liking this default, the default should be that it behaves like it always has.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -2100,6 +2162,19 @@ public ListBoxModel doFillNotificationLevelItems(@QueryParameter("serverName") f
         }
 
         /**
+         * Fill the dropdown for silent levels.
+         * @return the values.
+         */
+        public ListBoxModel doFillSilentLevelItems() {
```
rsandell: I don't think this is needed if the databinding is performed correctly, use `f:enum`.

src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/Messages_ja.properties
```diff
@@ -166,3 +166,9 @@ NotificationLevel_OWNER_REVIEWERS=\
  Owner and reviewers
 NotificationLevel_ALL=\
  All
+SilentLevel_COMMENTS=\
```
rsandell: Dito

src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger/config.jelly
```diff
@@ -194,8 +194,17 @@
         <f:entry title="${%Silent Mode}"
                  field="silentMode"
                  help="/plugin/gerrit-trigger/trigger/help-SilentMode.html">
-           <f:checkbox name="silentMode"
-                       checked="${it.silentMode}"/>
+            <table width="100%">
+                <f:optionalBlock title="" inline="true" name="silentMode"
```
rsandell: OK, I think I see what you are going for here, but then you need to handle `null` as a valid value in the code below as well and not default to something.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -334,6 +352,7 @@ public GerritTrigger(List<GerritProject> gerritProjects, SkipVote skipVote, Inte
         this.gerritTriggerTimerTask = null;
         this.triggerInformationAction = new GerritTriggerInformationAction();
         this.notificationLevel = notificationLevel;
+        this.silentLevel = silentLevel;
```
rsandell: This constructor is deprecated and should not be added to. use `@DataboudSetter`s instead.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -1450,6 +1475,33 @@ public String getNotificationLevel() {
...
+        return silentLevel;
+    }
+
+    /**
+     * Should comments be kept silent
+     *
+     * @return whether comments are kept silent
+     */
+    public boolean isSilentComments() {
+        return (silentMode && (silentLevel == SilentLevel.ALL || silentLevel == SilentLevel.COMMENTS));
```
rsandell: Instead of adding even more code to this class I think this logic is better put in static utility methods in the enum class.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -208,6 +210,20 @@ public GerritTrigger(List<GerritProject> gerritProjects) {
             this.notificationLevel = "";
         }
```
rsandell: `silentMode` needs to be deprecated and migrated to `silentLevel`. E.g./I.e. Deprecate `isSilendMode` and replace the implementation with `return silentLevel != null && silentLevel == SilentLevel.ALL` add a `readResolve` and deprecate and set `silentMode` to be `transient` etc.
https://wiki.jenkins-ci.org/display/JENKINS/Hint+on+retaining+backward+compatibility

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/job/rest/BuildCompletedRestCommandJob.java
```diff
@@ -58,31 +61,36 @@
...
     }
 
     @Override
     protected ReviewInput createReview() {
         SecurityContext old = ACL.impersonate(ACL.SYSTEM);
         try {
-            String message = parameterExpander.getBuildCompletedMessage(memoryImprint, listener);
             Collection<ReviewLabel> scoredLabels = new ArrayList<ReviewLabel>();
-            if (memoryImprint.getEvent().isScorable()) {
+            GerritTrigger trigger = GerritTrigger.getTrigger(build.getParent());
```
rsandell: This diff looks a bit scary, reminder to myself to look at it more.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ToGerritRunListener.java
```diff
@@ -215,10 +218,10 @@ public synchronized void onStarted(Run r, TaskListener listener) {
                     ((GerritEventLifecycle)cause.getEvent()).fireBuildStarted(r);
                 }
             }
-            if (!cause.isSilentMode()) {
+            GerritTrigger trigger = GerritTrigger.getTrigger(r.getParent());
+            if (!cause.isSilentMode() || (cause.isSilentMode() && !trigger.isSilentComments())) {
```
rsandell: there needs to be a better way to check this than just listing all possibilities, or at least a utility function.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ToGerritRunListener.java
```diff
@@ -162,18 +163,20 @@ public synchronized void onCompleted(@Nonnull Run r, @Nonnull TaskListener liste
      * Manages the end of a Gerrit Event. Should be called after each build related to an event completes if that build
      * should report back to Gerrit.
      *
+     * @param build   the current build.
      * @param event   the Gerrit Event which may need to be completed.
      * @param cause   the Gerrit Cause which triggered the build initially.
      * @param listener   the Jenkins listener.
      */
-    public synchronized void allBuildsCompleted(GerritTriggeredEvent event, GerritCause cause, TaskListener listener) {
+    public synchronized void allBuildsCompleted(Run build, GerritTriggeredEvent event,
```
rsandell: I'm sceptical to changing public signatures like this.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/config/SilentLevel.java
```diff
@@ -0,0 +1,61 @@
...
+    VOTES;
+
+
+    /**
+     * Method to get the list of silent levels.
+     * @return A list of silent levels.
+     */
+    public static List<String> getLevels() {
+      List<String> levels = new ArrayList();
+      levels.add(ALL.toString());
```
rsandell: I'm guessing you are using this for UI diaply, in that case it should be the I18n versions of the names.

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/mock/TestUtils.java
```diff
@@ -238,6 +239,7 @@ public static void waitForBuilds(Job project, int number, int timeoutMs) {
         private List<GerritProject> projects = new ArrayList<GerritProject>();
         private String serverName = PluginImpl.DEFAULT_SERVER_NAME;
         private boolean silentMode = false;
+        private SilentLevel silentLevel = SilentLevel.ALL;
```
rsandell: default should be `null` no?

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spec/SpecGerritTriggerHudsonTest.java
```diff
@@ -552,6 +553,7 @@ public void testBuildLatestPatchsetOnlyNonRelatedChangeDifferentBranchCannotAbor
 
         GerritTrigger trigger = project.getTrigger(GerritTrigger.class);
         trigger.setSilentMode(true);
+        trigger.setSilentLevel(SilentLevel.ALL);
```
rsandell: `setSilentMode` should be no more and replaced by `trigger.setSilentLevel(SilentLevel.ALL)` If I've understood your logic correctly.

quite:
Was this ever merged?

PR #299, opened 2016-10-05
[JENKINS-38675] use refspec instead of branch name for the revision

jacob-keller:
In the GerritTriggerBuildChooser, we have a custom method for
determining which revision to build. However, this revision is not
actually a branch but is instead a Gerrit refspec of the built event.
Instead of reporting the branch which is almost certainly meaningless to
the Git summary in our case, report the refspec string instead. We'll
use the branch as a fallback when running a build which was not built
due to a Gerrit event.

The result of this change is that we should see something like:

Revision: <sha1id>
  refs/changes/x/xxxx/x

Instead of

Revision: <sha1id>
  master

This produces a much more useful summary since it does not help to tell
the user "we built master" when instead we actually built one of the
Gerrit revisions that has not yet been committed.

Signed-off-by: Jacob Keller <jacob.e.keller@intel.com>

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTriggerBuildChooser.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTriggerBuildChooser.java
@@ -114,10 +114,15 @@ public class GerritTriggerBuildChooser extends BuildChooser {
                 rev = "FETCH_HEAD";
             }
 
+            String refspec = context.actOnBuild(new GetGerritEventRefspec());
+            if (refspec == null) {
+                refspec = singleBranch;
+            }
+
             ObjectId sha1 = git.revParse(rev);
 
             Revision revision = new Revision(sha1);
-            revision.getBranches().add(new Branch(singleBranch, sha1));
+            revision.getBranches().add(new Branch(refspec, sha1));
 
             return Collections.singletonList(revision);
         } catch (GitException e) {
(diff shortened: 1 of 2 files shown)


jacob-keller:
Yea, I agree. I have been thinking about how exactly to test this, thanks for the suggestions.

jacob-keller:
@rsandell I added a test. I don't think it's idea but I didn't manage to go down the GitSampleRepoRule path just yet. The current test relies heavily on mocking objects. Let me know if you have better suggestions.

I also found a bug in the previous revision as I had confused revisions and refname. The new version has a separate context for grabbing the gerrit event. It might be worth refactoring that more to avoid using context twice? Can we get context to return the actual gerritEvent and then let the main section determine what to do here? I wasn't sure how that worked.

jacob-keller:
It should be noted that GitSampleRepoRule is not in any released version of the git plugin, except for beta 3.0.0-beta1, so I don't think we can depend on it for the test easily :(

I think the mocked test is ok enough at the moment but it does seem pretty fragile.

jacob-keller:
I added a second follow on patch which shows my suggested alternative above. I do not think I know which I prefer at this point, and I really don't like the way the test is implemented with a copied BuildChooserContextImpl but I can't find a better way.

rsandell:
Well I was hoping for a more integration type test, using `JenkinsRule` that triggered a build with the `GerritTriggerBuildChooser` set and verified that the correct rev got checkedout in the workspace and that the changelog showed correctly. But this is at least a lot better than nothing :)

jacob-keller:
I couldn't figure out how to get that to work. We can always add an
additional test later though.

PR #300, opened 2016-11-08
no-test-jar=false

jglick:
Downstream of https://github.com/jenkinsci/plugin-pom/pull/38.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.18</version>
+        <version>2.19</version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -55,7 +56,7 @@
         <jenkins.version>1.609.3</jenkins.version>
         <!--<jenkins.version>2.5-SNAPSHOT</jenkins.version>-->
         <java.level>6</java.level>
-        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
```
jglick: Already inherited.

jglick:
Needs fix, do not merge.

jglick:
Now should be OK.

PR #302, opened 2016-11-28
[JENKINS-40059] CommentAdded trigger firing on every comment

tyroneabdy:
The CommentAdded trigger was firing on any comment after
the necessary value of an approval had been met.
This was happening due to that in previous versions of gerrit
before 2.13.0 approval information only appeared once
something had changed.
This changed to use the oldValue that would optionally
be included on state change. The logic added
to the gerrit-trigger-plugin was based on an earlier
patchset that had an ever present variable for state
change.
This led to a fall through to the old style of checking
for whether to trigger based on a comment.

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
@@ -1036,14 +1036,18 @@ public class GerritTrigger extends Trigger<Job> {
                      * Gerrit stream events changed to append approval info to
                      * every comment-added event.
                      **/
-                    if (approval.isUpdated()
-                            && approval.getType().equals(
-                                commentAdded.getVerdictCategory())
-                            && (approval.getValue().equals(
-                                commentAdded.getCommentAddedTriggerApprovalValue())
-                            || ("+" + approval.getValue()).equals(
-                                commentAdded.getCommentAddedTriggerApprovalValue()))) {
-                            return true;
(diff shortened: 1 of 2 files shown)


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -1036,14 +1036,18 @@ private boolean isServerInteresting(GerritTriggeredEvent event) {
...
-                            && approval.getType().equals(
-                                commentAdded.getVerdictCategory())
-                            && (approval.getValue().equals(
-                                commentAdded.getCommentAddedTriggerApprovalValue())
-                            || ("+" + approval.getValue()).equals(
-                                commentAdded.getCommentAddedTriggerApprovalValue()))) {
-                            return true;
+                    GerritServer server = PluginImpl.getServer_(serverName);
+                    if (server.getGerritVersion() != null
+                            && server.getGerritVersion().compareTo("2.13.0") >= 0) {
```
rsandell: You should implement this using real major.minor.micro version checking.
See https://github.com/jenkinsci/gerrit-trigger-plugin/blob/master/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/version/GerritVersionChecker.java

scoheb:
Hi @rsandell !

Any chance we can get this released? Is there anything else to address?

rsandell:
Just tickled the pr builder.

scoheb:
Test failures fixed here: https://github.com/jenkinsci/gerrit-trigger-plugin/issues/309

PR #303, opened 2017-01-12
Change Nr. is a link to review detail

jbesta:
On search results page, user may open list of parameters about a change. One of this parameters is `GERRIT_CHANGE_URL` which is a URL to Gerrit review server and user may click on `(link)` to navigate to Gerrit review detail. This change makes the same link from `Change Nr` so that user may navigate to Gerrit review detail directly.


jbesta:
Closing PR in favor of #305

PR #304, opened 2017-01-12
Prevent double HTML escaping of subject in search results

jbesta:
Characters like `"` in commit message are not properly displayed on search page - `&#34;` is rendered Instead of `"`.


rsandell:
Yepp, seems like something we missed when adding `<?jelly escape-by-default='true'?>` to index.jelly

PR #305, opened 2017-01-12
Change Nr. is a link to review detail

jbesta:
On search results page, user may open list of parameters about a change. One of this parameters is `GERRIT_CHANGE_URL` which is a URL to Gerrit review server and user may click on `(link)` to navigate to Gerrit review detail. This change makes the same link from `Change Nr` so that user may navigate to Gerrit review detail directly.

### DIFF
--- a/src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/actions/manual/ManualTriggerAction/index.jelly
+++ b/src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/actions/manual/ManualTriggerAction/index.jelly
@@ -132,8 +132,19 @@
                                                                    id="check${theId}"
                                                                    onClick="rowSelected('${theId}')"/>
                                                         </td>
-                                                        <td onClick="activateRow('${theId}')">
-                                                            ${res.getString('number')}
+                                                        <j:set var="selectedServer" value="${request.session.getAttribute('selectedServer')}"/>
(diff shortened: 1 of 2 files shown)


scoheb:
Looks good. What about a test?

Something like GerritTriggerProjectHudsonTest#testPopulateDropDown where you verify that the HtmlElement exists?

src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/actions/manual/ManualTriggerAction/index.jelly
```diff
@@ -132,8 +132,19 @@
                                                                    id="check${theId}"
                                                                    onClick="rowSelected('${theId}')"/>
                                                         </td>
-                                                        <td onClick="activateRow('${theId}')">
```
rsandell: I feel the number is the most common place you'd want to click for activating the row, so why remove it??

jbesta:
Initially, I removed it, because it activated row even if you clicked on the link to navigate to Gerrit.
I added the functionality back in following commit and now `click` event is consumed after clicking on the link, so it no longer happens.

src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/actions/manual/ManualTriggerAction/index.jelly
jbesta: Initially, I removed it, because it activated row even if you clicked on the link to navigate to Gerrit.
I added the functionality back in following commit and now click event is consumed after clicking on the link, so it no longer happens.

PR #306, opened 2017-01-18
Display full commit message in tooltip

jbesta:
On *Gerrit Manual Trigger* page, "only" commit message header
(first 65 characters of commit message) is displayed in `Subject`
column.
This change modifies it such that when user hovers over change
subject, native tooltip with full commit message is displayed.
This should provide additional information about a change.


jbesta:
I can't find in gerrit changelogs exact version which introduced this field, however, the feature was implemented on Gerrit server already in 10/2011 according to this commit https://github.com/gerrit-review/gerrit/commit/0943d6e7fcbfbf10a55a67cd6cb84a8503d1addb.

Anyway, `gerrit query` command documentation page (https://review.openstack.org/Documentation/cmd-query.html) states:

> Note that any field may be missing in the JSON messages, so consumers of this JSON stream should deal with that appropriately.

so I think we should add additional existence check for `commitMessage` field in JSON.

rsandell:
`opt` or `optString` could also work.

PR #308, opened 2017-02-10
Support specification of variable patterns

rburgst:
- this change allows the use of system environment patterns for
  project and branch specification such as

  ${PROJECT_NAME}

- Example:
  the gerrit host has a configured environment variable defined as

  export PROJECT_NAME=myproject

  now when we push a review commit for "myproject" then the gerrit
  trigger will automatically react

- one use case for this is to be able to specfiy a generic seed
  job for a jenkins instance which will automatically trigger
  a build on the variable based project

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/CompareType.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/CompareType.java
@@ -46,7 +46,13 @@ public enum CompareType {
     /**
      * Regular expression comparison.
      */
-    REG_EXP(new RegExpCompareUtil());
+    REG_EXP(new RegExpCompareUtil()),
+
+    /**
+     * Plain comparison however, the pattern will be expanded with the system environment.
+     * This allows comparisons such as "${PROJECTNAME}".
+     */
+    PLAIN_VAR(new CompareUtil.PlainVarCompareUtil());
 
     /**
      * Gets a list of all CompareType's displayNames.
@@ -88,7 +94,7 @@ public enum CompareType {
         return PLAIN;
     }
 
-    private CompareUtil util;
+    CompareUtil util;
 
     /**
      * Private Constructor.
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/CompareUtil.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/CompareUtil.java
@@ -23,11 +23,14 @@
  */
 package com.sonyericsson.hudson.plugins.gerrit.trigger.hudsontrigger.data;
 
-import java.io.File;
+import hudson.EnvVars;
 import org.apache.tools.ant.types.selectors.SelectorUtils;
+
+import java.io.File;
 
 /**
  * Base interface for the compare-algorithms.
+ *
  * @author Robert Sandell &lt;robert.sandell@sonyericsson.com&gt;
  */
 public interface CompareUtil {
@@ -56,7 +59,7 @@ public interface CompareUtil {
      * Compares based on Ant-style paths.
      * like <code>my/&#042;&#042;/something&#042;.git</code>
      */
-    static class AntCompareUtil implements CompareUtil {
+    class AntCompareUtil implements CompareUtil {
 
         @Override
         public boolean matches(String pattern, String str) {
@@ -80,9 +83,9 @@ public interface CompareUtil {
     }
 
     /**
-     * Compares with pattern.equals(str).
+     * Compares with pattern.equalsIgnoreCase(str).
      */
-    static class PlainCompareUtil implements CompareUtil {
+    class PlainCompareUtil implements CompareUtil {
 
         @Override
         public boolean matches(String pattern, String str) {
@@ -105,7 +108,7 @@ public interface CompareUtil {
      * string.matches(pattern)
      * @see java.util.regex.Pattern
      */
-    static class RegExpCompareUtil implements CompareUtil {
+    class RegExpCompareUtil implements CompareUtil {
 
         @Override
         public boolean matches(String pattern, String str) {
@@ -120,6 +123,43 @@ public interface CompareUtil {
         @Override
         public char getOperator() {
             return '~';
+        }
+    }
+
+    /**
+     * Compares just like plain comparison, however, patterns are expanded using the
+     * given system environment settings.
+     */
+    class PlainVarCompareUtil implements CompareUtil {
+        private final EnvVars systemEnvVars;
+
+        public PlainVarCompareUtil() {
+            systemEnvVars = new EnvVars(EnvVars.masterEnvVars);
+        }
+
+        /**
+         * Test only method to allow injecting test properties into our environment.
+         * @param key the key to inject into the env vars.
+         * @param value the value associated with the key.
+         */
+        void inject(String key, String value) {
+            systemEnvVars.put(key, value);
+        }
+
+        @Override
+        public boolean matches(String pattern, String str) {
+            String expandedPattern = systemEnvVars.expand(pattern);
+            return expandedPattern.equalsIgnoreCase(str);
+        }
+
+        @Override
+        public String getName() {
+            return "PlainVar";
+        }
+
+        @Override
+        public char getOperator() {
+            return ':';
         }
     }
 }
--- a/src/main/webapp/trigger/help-GerritTriggerConfiguration.html
+++ b/src/main/webapp/trigger/help-GerritTriggerConfiguration.html
@@ -8,7 +8,9 @@
     You can specify the name pattern in three different ways, as provided by the &quot;Type&quot; drop-down menu.
 </p>
 <ul>
-    <li><strong>Plain:</strong> The exact name in Gerrit, case sensitive equality.</li>
+    <li><strong>Plain:</strong> The exact name in Gerrit, case insensitive equality.</li>
+    <li><strong>PlainVar:</strong> The exact name in Gerrit, case insensitive equality, allows usage of system environment
+        variables defined on the gerrit host (e.g. <code>${ENVIRONMENT}</code>).</li>
     <li><strong>Path:</strong> ANT style pattern. <i>Ex: &quot;**/base/*&quot;</i></li>
     <li><strong>RegExp:</strong> Regular expression.</li>
 </ul>
--- a/src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/GerritProjectWithFilesInterestingTest.java
+++ b/src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/GerritProjectWithFilesInterestingTest.java
@@ -23,6 +23,7 @@
  */
 package com.sonyericsson.hudson.plugins.gerrit.trigger.hudsontrigger.data;
 
+import org.junit.BeforeClass;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
@@ -58,6 +59,13 @@ public class GerritProjectWithFilesInterestingTest {
     public void testInteresting() {
         assertEquals(scenarioWithFiles.expected, scenarioWithFiles.config.isInteresting(
                 scenarioWithFiles.project, scenarioWithFiles.branch, scenarioWithFiles.topic, scenarioWithFiles.files));
+    }
+
+    @BeforeClass
+    public static void setupEnv() {
+        // this is a somewhat hacky way to deal with the fact that the compare util was already set up before
+        // we get the chance to update the system environment
+        ((CompareUtil.PlainVarCompareUtil) CompareType.PLAIN_VAR.util).inject("PROJECTNAME", "myproject");
     }
 
     /**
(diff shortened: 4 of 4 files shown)


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/CompareUtil.java
```diff
@@ -56,7 +59,7 @@
      * Compares based on Ant-style paths.
      * like <code>my/&#042;&#042;/something&#042;.git</code>
      */
-    static class AntCompareUtil implements CompareUtil {
```
rsandell: why remove the `static` declaration?

rsandell: why remove the `static` declaration?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/CompareUtil.java
```diff
@@ -122,4 +125,41 @@ public char getOperator() {
             return '~';
         }
     }
+
+    /**
+     * Compares just like plain comparison, however, patterns are expanded using the
+     * given system environment settings.
+     */
+    class PlainVarCompareUtil implements CompareUtil {
```
rsandell: I think it would be better to skip this one and add the functionality to the existing Plain comparer, deprecate the old `matches` method and introduce a new one that also takes an EnvVars as parameter.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/CompareType.java
```diff
@@ -88,7 +94,7 @@ public static CompareType findByOperator(char operator) {
         return PLAIN;
     }
 
-    private CompareUtil util;
+    CompareUtil util;
```
rsandell: should be no need for this.

src/main/webapp/trigger/help-GerritTriggerConfiguration.html
```diff
@@ -8,7 +8,9 @@
     You can specify the name pattern in three different ways, as provided by the &quot;Type&quot; drop-down menu.
 </p>
 <ul>
-    <li><strong>Plain:</strong> The exact name in Gerrit, case sensitive equality.</li>
+    <li><strong>Plain:</strong> The exact name in Gerrit, case insensitive equality.</li>
+    <li><strong>PlainVar:</strong> The exact name in Gerrit, case insensitive equality, allows usage of system environment
```
rsandell: I doubt it is on the gerrit host? You me4an the Jenkins master right? Hard to say since this PR seems incomplete.

rsandell:
Just tickling the PR builder

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/CompareUtil.java
rburgst: IntelliJ improvement suggestion. A static inner class in an interface has no effect.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/CompareUtil.java
rburgst: I didn't want to mess too much with the existing code, but this would definitely be a good improvement. I can do that.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/CompareType.java
rburgst: with the existing setup it was necessary for the test. With your suggestion it won't be any more. Correct.

src/main/webapp/trigger/help-GerritTriggerConfiguration.html
rburgst: Yes, the variables have to exist on the Jenkins host.

rburgst:
My git workspace is clean, what files do you think are missing?

rsandell:
Well, there should be a call to `inject` other than the test, where else would the environment come from?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/CompareUtil.java
```diff
@@ -122,4 +125,41 @@ public char getOperator() {
...
+        private final EnvVars systemEnvVars;
+
+        public PlainVarCompareUtil() {
+            systemEnvVars = new EnvVars(EnvVars.masterEnvVars);
```
rsandell: Ah, I had missed the use of `EnvVars.masterEnvVars`. I don't think it's a good choice to put it here since you would then only pick up whatever is set when Jenkins starts up and not what is changed during runtime.
We need to find a way to provide it as close to the call to `matches` as possible, without having to create a new instance and copy over all values from `EnvVars.masterEnvVars` for every job check, since that could cause some performance issues.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/CompareUtil.java
```diff
@@ -39,9 +39,10 @@
      * Tells if the given pattern matches the string according to the implemented comparer/algorithm.
      * @param pattern the pattern to use.
      * @param str the string to match on.
+     * @param envVars the environment variables exisiting on the jenkins host.
      * @return true if the string matches the pattern.
      */
-    boolean matches(String pattern, String str);
+    boolean matches(String pattern, String str, EnvVars envVars);
```
rsandell: Darn it, I didn't see that this was an interface when I recommended the change to the signatures, but I did say deprecate the old signature and add a new one.
Since it is a public interface we can't easily tell if there are any implementations of this in other plugins and we need to try to keep binary backwards compatibility.

The simplest way I can think of at the moment would be to
 * Add back `boolean matches(String pattern, String str);` and deprecate it.
 * Add a new class: AbstractCompareUtil or something and make it an abstract class, implementing the old method signature that only calls the new method `boolean matches(String pattern, String str, EnvVars envVars);` with `null` for `envVars`.
 * Make `AntCompareUtil` et.al. extend `AbstractCompareUtil` instead of `implements CompareUtil`. They also needs to stay in this class so they keep their old names.
 * Then make all references to `CompareUtil` in this plugin reference `AbstractCompareUtil` instead.
 * Add a new file with the interface

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/CompareUtil.java
```diff
@@ -111,8 +115,9 @@ public char getOperator() {
     class RegExpCompareUtil implements CompareUtil {
 
         @Override
-        public boolean matches(String pattern, String str) {
-            return str.matches(pattern);
+        public boolean matches(String pattern, String str, EnvVars envVars) {
+            String expandedPattern = envVars.expand(pattern);
```
rsandell: needs null check on `envVars`

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/GerritProject.java
```diff
@@ -210,23 +212,24 @@ public void setForbiddenFilePaths(List<FilePath> forbiddenFilePaths) {
      * @param branch the branch.
      * @param topic the topic.
      * @param files the files.
+     * @param envVars the environment variables exisiting on the jenkins host.
      * @return true is the rules match.
      */
-    public boolean isInteresting(String project, String branch, String topic, List<String> files) {
-        if (compareType.matches(pattern, project)) {
+    public boolean isInteresting(String project, String branch, String topic, List<String> files, EnvVars envVars) {
```
rsandell: Same here. This is a public api and needs to keep backwards compatibility as much as possible.

Keep the old signature, deprecate it and have it call the new signature with `null` to the new parameter.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -991,21 +1019,39 @@ public boolean equals(Object obj) {
...
+     * Checks if the current system env vars are still the same and updates our local copy if necessary
+     */
+    private void updateEnvVarsIfNecessary() {
+        EnvironmentVariablesNodeProperty envVarsProp =
+                Jenkins.getActiveInstance().getGlobalNodeProperties().get(EnvironmentVariablesNodeProperty.class);
+        EnvVars globalEnvVars = envVarsProp.getEnvVars();
+        int newSystemEnvVarsHash = EnvVars.masterEnvVars.hashCode();
+        int newGlobalEnvVarsHash = globalEnvVars.hashCode();
+        if (newSystemEnvVarsHash != systemEnvVarsHash || newGlobalEnvVarsHash != globalHashVarsHash) {
+            this.envVars.clear();
```
rburgst: very good point. It extends `TreeMap` so unfortunately its not. Is there a performance penalty to consider or is it enough to make it `synchronized`?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/CompareUtil.java
rburgst: I didn't realise that this code might be used by other plugins, I guess there isn't really a way to figure this out?

rsandell: Yes, sorry. The "darn it" was on me not on you :)
This plugin has been around for quite a long time so it's really hard to tell what public methods can be changed and not, not just other plugins but whatever groovy scripts that admins has laying around etc. So we try to keep binary compatibility as much as possible and only make breaking changes when absolutely nessesary.

rburgst:
please let me know if the general direction of the patch is fine,  then I will proceed and see about the build failures.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/CompareType.java
```diff
@@ -102,10 +104,11 @@ private CompareType(CompareUtil util) {
      * Tells if the given string matches the given pattern based on the algorithm of this CompareType instance.
      * @param pattern the pattern
      * @param str the string
+     * @param envVars the environment variables exisiting on the jenkins host.
      * @return true if the string matches the pattern.
      */
-    public boolean matches(String pattern, String str) {
```
rsandell: The old signature needs to be kept and deprecated for backwards binary compatibility

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/CompareUtil.java
```diff
@@ -52,20 +65,42 @@
...
+            }
+        }
+    }
+
     /**
      * Compares based on Ant-style paths.
      * like <code>my/&#042;&#042;/something&#042;.git</code>
      */
-    static class AntCompareUtil implements CompareUtil {
+    class AntCompareUtil extends AbstractCompareUtil {
```
rsandell: `static class`

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/CompareUtil.java
```diff
@@ -37,10 +40,20 @@
      * @param pattern the pattern to use.
      * @param str the string to match on.
      * @return true if the string matches the pattern.
+     * @deprecated use {@link #matches(String, String, EnvVars)} instead.
      */
     boolean matches(String pattern, String str);
```
rsandell: Add a `@Deprecated` annotation as well so tools has an easier time figuring it out.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/CompareUtil.java
```diff
@@ -52,20 +65,42 @@
...
+    class AntCompareUtil extends AbstractCompareUtil {
 
         @Override
-        public boolean matches(String pattern, String str) {
+        public boolean matches(String pattern, String str, EnvVars envVars) {
             // Replace the Git directory separator character (always '/')
             // with the platform specific directory separator before
             // invoking Ant's platform specific path matching.
             String safePattern = pattern.replace('/', File.separatorChar);
+            String expandedPattern = expandWithEnvVarsIfPossible(safePattern, envVars);
```
rsandell: I think it is better to do this before the `\` to `/` replacement in case any of the used env vars contains platform specific paths.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/GerritProject.java
rsandell: My old comment is still valid. Old method signatures needs to be kept and deprecated.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -178,6 +183,9 @@
     private List<PluginGerritEvent> triggerOnEvents;
     private boolean dynamicTriggerConfiguration;
     private String triggerConfigURL;
+    private final EnvVars envVars;
```
rsandell: This is no good, when the project is saved these fields will be serialized with the trigger and take up unnecessary disk space.

I think a `ThreadLocal` `LoadingCache` or something is better.

PR #309, opened 2017-04-10
Fix Tests for Comment Added for Gerrit 2.13+

scoheb:
Adds to https://github.com/jenkinsci/gerrit-trigger-plugin/pull/302


rsandell:
OK, I merged both locally to be able to test that it worked. But it seems to be working.

PR #310, opened 2017-04-18
Fix CommentAdded event for Gerrit snapshot version < 2.13

hugares:
0a8e1c246 adapted CommentAdded event firing for Gerrit version >= 2.13
and backward compatibility is kept for Gerrit version < 2.13.

The version check done to maintain backward compatibility was not
excluding snapshot so snapshot versions that are < 2.13 were handling
CommentAdded event the way they are supposed to be for Gerrit >= 2.13.


hugares:
@rsandell When do you plan to release 2.23.2? I am asking because we do not centrally managed our Jenkins instances so people can install 2.23.1 anytime and hit the regression that this PR is fixing.

rsandell:
@hugares The release finished at: 2017-04-19T13:08:26+02:00 during my lunch break :)

hugares:
Thanks a lot!

PR #312, opened 2017-05-10
[JENKINS-39010] CommentAddedRegex: Input is a multiline message

bpedersen2:
Interpret the incoming comment message as multi-line text, as gerrit
prefixes some heading to the input.

Add debug logging as well.

Fixes: [JENKINS-39010](https://issues.jenkins-ci.org/browse/JENKINS-39010)


rsandell:
Thanks!
Please provide unit tests next time.

bpedersen2:
Thanks for merging. I actually looked at providin ga test, but as I found now suitable test to derive from , I ahd to give up (I normally just do python programming, so writing a java test from scratch is quite time consuming for me)

rsandell:
Here is one although it is using a lot of mocks so maybe not the best one to copy https://github.com/jenkinsci/gerrit-trigger-plugin/blob/master/src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/EventListenerTest.java#L120

PR #315, opened 2017-05-24
GerritDynamicUrlProcessor: fix slf4j notation

alonbl:
Signed-off-by: Alon Bar-Lev <alon.barlev@gmail.com>

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritDynamicUrlProcessor.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritDynamicUrlProcessor.java
@@ -180,7 +180,7 @@ public final class GerritDynamicUrlProcessor {
         char operChar = oper.charAt(0);
         CompareType type = CompareType.findByOperator(operChar);
 
-        logger.trace("==> item:({0}) oper:({1}) text:({2})", new Object[]{item, oper, text});
+        logger.trace("==> item:({}) oper:({}) text:({})", new Object[]{item, oper, text});
 
         if (SHORTNAME_PROJECT.equals(item)) { // Project
           // stash previous project to the list


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritDynamicUrlProcessor.java
```diff
@@ -180,7 +180,7 @@ private static String regexEscapted(char symbol) {
         char operChar = oper.charAt(0);
         CompareType type = CompareType.findByOperator(operChar);
 
-        logger.trace("==> item:({0}) oper:({1}) text:({2})", new Object[]{item, oper, text});
+        logger.trace("==> item:({}) oper:({}) text:({})", new Object[]{item, oper, text});
```
hugares: I think you can remove the object array as well:
logger.trace("==> item:({}) oper:({}) text:({})", item, oper, text);

alonbl: Correct! Fixed.

rsandell: I don't see the fix

alonbl: we are using too old slf4j, I had to revert this as the slf4j does not support ...

the important fix is not to use {n}.

rsandell: Yea, I see that now. Strange, we should perhaps look into bumping slf4j to a newer version but that might put us in a bit of trouble since core is also bundling a version of slf4j that might conflict.

PR #316, opened 2017-05-28
Support sending multiline comments

lorenzph:
While Gerrit's API supports multiline review comments the Gerrit Trigger
UI limits the user to single line inputs. Improve this by using a
textarea widget instead.


lorenzph:
@rsandell: I have added a test that verifies that the ParameterExpander correctly handles newlines within the command template. I am not sure how else this UI only change can be tested (essentially the same thing can be accomplished by editing the plugin configuration file but editing the config via the UI ends up removing the newlines as the browser discards the newlines from the textboxes).

Let me know if you have any further suggestions on how to test this.

rsandell:
Well I was hoping for to have a test that actually sends a message with a new line in it, so that we know the default command line works with a newline in the message. It does not only affect the UI since it will now be able to send back message strings to the command. And that is what I need to know works.

PR #317, opened 2017-06-15
Set tag for review comments on REST

orgads:
This allows to filter out these messages on Gerrit 2.14 UI

[JENKINS-43904]

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/job/rest/BuildCompletedRestCommandJob.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/job/rest/BuildCompletedRestCommandJob.java
@@ -116,7 +116,8 @@ public class BuildCompletedRestCommandJob extends AbstractRestCommandJob {
                 }
             }
 
-            return new ReviewInput(message, scoredLabels, commentedFiles).setNotify(notificationLevel);
+            return new ReviewInput(message, scoredLabels, commentedFiles).setNotify(notificationLevel)
+                .setTag(Constants.TAG_VALUE);
         } finally {
             SecurityContextHolder.setContext(old);
         }
--- a/pom.xml
+++ b/pom.xml
@@ -70,7 +70,7 @@
         <dependency>
             <groupId>com.sonymobile.tools.gerrit</groupId>
             <artifactId>gerrit-events</artifactId>
-            <version>2.10.2</version>
+            <version>2.11.0</version>
             <!-- New source is here: https://github.com/sonyxperiadev/gerrit-events -->
         </dependency>
         <dependency>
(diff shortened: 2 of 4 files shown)


orgads:
Do you prefer to have it configurable? I'm not sure there's a reason...

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/job/rest/BuildCompletedRestCommandJob.java
```diff
@@ -116,7 +116,8 @@ protected ReviewInput createReview() {
                 }
             }
 
-            return new ReviewInput(message, scoredLabels, commentedFiles).setNotify(notificationLevel);
+            return new ReviewInput(message, scoredLabels, commentedFiles).setNotify(notificationLevel)
+                .setTag("jenkins-gerrit-trigger");
```
rsandell: I would prefer it is a constant somewhere instead of a literal.

rsandell:
We can start by having it as a constant, and if someone wants it configurable they are welcome to contribute that feature ;)

Have you tried this on an older Gerrit that doesn't support tag to see if it still works?
If it doesn't we will nee to add some version checks. see https://github.com/jenkinsci/gerrit-trigger-plugin/blob/master/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/version/GerritVersionChecker.java#L44

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/job/rest/BuildCompletedRestCommandJob.java
orgads: Can you suggest where?

orgads:
[**@rsandell**](https://github.com/rsandell) commented on [Jun 15, 2017, 12:25 PM GMT+3](https://github.com/jenkinsci/gerrit-trigger-plugin/pull/317#issuecomment-308678454 "2017-06-15T09:25:08Z - Replied by Github Reply Comments"):
> Have you tried this on an older Gerrit that doesn't support tag to see if it still works?  

Tried on https://codereview.qt-project.org which has Gerrit 2.7. It works fine.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/job/rest/BuildCompletedRestCommandJob.java
orgads: Done

orgads:
Can you please release it?

rsandell:
Sorry, got lost in my pile of TODOs

Released 2.24.0 a few minutes ago

PR #318, opened 2017-06-24
[JENKINS-32898] Fix Disable Strict Forbidden File Verification option

kars7e:
When Strict Forbidden File Verification option is disabled, job can
still be triggered if file paths are mixed, i.e. there are some files
that match forbidden file paths, and some that only match included file
paths. However, when ALL files match forbidden file paths, job should
not be triggered.

This patch ensures this by checking whether all files match forbidden
file paths if Strict Forbidden File Verification Option is disabled.


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/GerritProject.java
```diff
@@ -213,37 +213,32 @@ public void setForbiddenFilePaths(List<FilePath> forbiddenFilePaths) {
...
                     }
-                    if (disableStrictForbiddenFileVerification) {
-                        // Here we want to be able to trigger a build if the event contains
-                        // wanted topics or file paths even though there may be a forbidden file
-                        return foundInterestingTopicOrFile;
-                    } else {
-                        if (foundInterestingForbidden) {
-                            // we have a forbidden file and a wanted file path.
+                } else {
+                    for (FilePath ffp : forbiddenFilePaths) {
```
rsandell: probably needs a `null` check as you've done in `isForbiddenFile`

PR #319, opened 2017-07-25
Provide the data about parent builds as environment variable to the dependent build

Jimilian:
Idea is quite simple: we have build and test jobs and we want to get rid of intermediate steps. So, test job has build job as dependency and know how to download artifacts.
Other possible use cases:
* build some shared code once and re-use in dependent builds
* re-trigger only small pieces
* provide more granular feedback

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/dependency/DependencyQueueTaskDispatcher.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/dependency/DependencyQueueTaskDispatcher.java
@@ -182,15 +182,40 @@ public final class DependencyQueueTaskDispatcher extends QueueTaskDispatcher
         }
 
 
-        List<Job> blockingProjects = getBlockingDependencyProjects(dependencies, event);
+        Job blockingProject = getBlockingDependencyProjects(dependencies, event);
 
-        if (blockingProjects.size() > 0) {
-            return new BecauseDependentBuildIsBuilding(blockingProjects);
+        if (blockingProject != null) {
+            return new BecauseDependentBuildIsBuilding(blockingProject);
         } else {
             logger.info("No active dependencies on project: {} , it will now build", p);
+
+            ToGerritRunListener toGerritRunListener = ToGerritRunListener.getInstance();
+            if (toGerritRunListener != null) {
+                List<Run> parentRuns = toGerritRunListener.getRuns(event);
+
+                if (parentRuns == null) {
+                    logger.info("All dependencies on project: {}, are triggered in silent mode. "
+                            + "Can not get list of actual dependencies", p);
+                    return null;
+                }
+
+                List<Run> actualDependencies = new ArrayList<Run>(dependencies.size());
+                for (Run run : parentRuns) {
+                    if (dependencies.contains(run.getParent())) {
+                        actualDependencies.add(run);
+                    }
+                }
+
+                // TODO: returning `null` from a QueueTaskDispatcher does not mean the build will start immediately.
+                // So, this line can be called multiple times. In case of performance/stability issues it makes sense to
+                // mode to TransientActionFactory or maybe a QueueListener would be better.
+                item.replaceAction(new GerritDependencyAction(actualDependencies));
+            }
+
             return null;
         }
     }
+
 
     /**
      * Gets the subset of projects which have a building element needing to complete for the same event.
@@ -198,19 +223,17 @@ public final class DependencyQueueTaskDispatcher extends QueueTaskDispatcher
      * @param event The event should have also caused the blocking builds.
      * @return the sublist of dependencies which need to be completed before this event is resolved.
      */
-    protected List<Job> getBlockingDependencyProjects(List<Job> dependencies,
-            GerritTriggeredEvent event) {
-        List<Job> blockingProjects = new ArrayList<Job>();
+    private Job getBlockingDependencyProjects(List<Job> dependencies, GerritTriggeredEvent event) {
         ToGerritRunListener toGerritRunListener = ToGerritRunListener.getInstance();
         if (toGerritRunListener != null) {
             for (Job dependency : dependencies) {
                 if (toGerritRunListener.isTriggered(dependency, event)
                         && toGerritRunListener.isBuilding(dependency, event)) {
-                    blockingProjects.add(dependency);
+                    return dependency;
                 }
             }
         }
-        return blockingProjects;
+        return null;
     }
 
     /**
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/dependency/GerritDependencyAction.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/dependency/GerritDependencyAction.java
@@ -1,0 +1,74 @@
+package com.sonyericsson.hudson.plugins.gerrit.trigger.dependency;
+
+import hudson.EnvVars;
+import hudson.model.AbstractBuild;
+import hudson.model.EnvironmentContributingAction;
+import hudson.model.InvisibleAction;
+import hudson.model.Run;
+
+import javax.annotation.Nonnull;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Adds Action that stores the data about dependency jobs.
+ */
+public class GerritDependencyAction extends InvisibleAction implements EnvironmentContributingAction {
+    @Nonnull
+    private final List<String> deps;
+
+    /**
+     * Saves the important information about the parent runs to be persistent.
+     * Stores string to keep the data even if original execution of parent job would be removed from history.
+     * @param runs List of runs it depend on
+     */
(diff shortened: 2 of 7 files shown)


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/dependency/DependencyQueueTaskDispatcher.java
```diff
@@ -188,18 +191,50 @@ public CauseOfBlockage canRun(Queue.Item item) {
...
+                List<ParameterValue> parameters = new ArrayList<ParameterValue>();
+                StringBuilder actualDependencies = new StringBuilder();
+                for (Run run : toGerritRunListener.getRuns(event)) {
+                    if (dependencies.contains(run.getParent())) {
+                        String originalName = run.getParent().getFullName();
+                        String keyName = originalName.replaceAll("[^a-zA-Z0-9]+", "_");
+                        String prefix = "DEP_" + keyName;
+                        String number = String.valueOf(run.getNumber());
+                        String result = run.getResult().toString();
+                        parameters.add(new StringParameterValue(prefix + "_BUILD_NAME", originalName));
```
rsandell: This won't work after the introduction of SECURITY-170. Core will reject these parameters.

Add an [EnvironmentContributingAction](http://javadoc.jenkins-ci.org/hudson/model/EnvironmentContributingAction.html) to the build instead of build parameters.

The build parameters was only kept in this plugin due to backwards compatibility. Any new stuff not related to the event itself we should try to provide the more "correct way".

Jimilian: Thx, will do

Jimilian: @rsandell, I did it... and only after that I faced the problem. This Action doesn't support Pipeline jobs... https://issues.jenkins-ci.org/browse/JENKINS-29537

Jimilian:
Filled https://github.com/jenkinsci/jenkins/pull/2975 as well.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/dependency/GerritDependencyAction.java
```diff
@@ -0,0 +1,44 @@
...
+import hudson.model.Run;
+
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Adds Action that stores the data about dependency jobs.
+ */
+public class GerritDependencyAction extends InvisibleAction implements EnvironmentContributingAction {
+    private List<Run> runs = Collections.emptyList();
```
jglick: If this action is ever serialized to disk anywhere, which I suppose it is (seems to be added to `queue.xml` and `EnvironmentContributingAction` is added to `build.xml), then this field is illegal. It will in fact explode on you: https://github.com/jenkinsci/jenkins/pull/2957

Maybe you just want a `List<String>` of `id`s?

Jimilian:
@rsandell, @jglick, please, review

Jimilian:
@rsandell, @jglick, could you take a look, please?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/dependency/GerritDependencyAction.java
```diff
@@ -0,0 +1,74 @@
...
+
+            env.put(prefix + "_BUILD_NAME", originalName);
+            env.put(prefix + "_BUILD_NUMBER", number);
+            env.put(prefix + "_BUILD_RESULT", result);
+
+            depKeys.append(keyName);
+            depKeys.append(" ");
+            }
+
+        env.put("DEP_KEYS", depKeys.toString().trim());
```
rsandell: is `DEP_KEYS` really a good name?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/dependency/DependencyQueueTaskDispatcher.java
```diff
@@ -188,18 +188,40 @@ public CauseOfBlockage canRun(Queue.Item item) {
             return new BecauseDependentBuildIsBuilding(blockingProjects);
         } else {
             logger.info("No active dependencies on project: {} , it will now build", p);
+
+            ToGerritRunListener toGerritRunListener = ToGerritRunListener.getInstance();
```
rsandell: The above logging statement is a bit misleading, probably leading to this.
But returning `null` from a `QueueTaskDispatcher` does not necessarily mean that the build will run. It only means that **this** dispatcher thinks it's OK for the build to start. There could be other dispatchers later in the chain that will block it still.
And that means that this code can be called several times for an item in the queue and you'll end up adding multiple `GerritDependencyAction`s to the build.
Using either a `TransientActionFactory` or maybe a `QueueListener` would be better. You could also potentially add the one action to it when the trigger actually schedules the build and let it calculate itself when needed.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/dependency/GerritDependencyAction.java
```diff
@@ -0,0 +1,74 @@
...
+
+    /**
+     * Saves the important information about the parent runs to be persistent.
+     * Stores string to keep the data even if original execution of parent job would be removed from history.
+     * @param runs List of runs it depend on
+     */
+    public GerritDependencyAction(List<Run> runs) {
+        deps = new ArrayList<String>(runs.size());
+        for (Run run : runs) {
+            deps.add(run.getParent().getFullName() + "#" + run.getNumber() + "#" + run.getResult());
```
rsandell: See if you can use [getExternalizableId](http://javadoc.jenkins-ci.org/hudson/model/Run.html#getExternalizableId--)

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/dependency/GerritDependencyAction.java
Jimilian: I did not want to put `GERRIT_` prefix in front to distinguish plugin "features" and data that comes from Gerrit. `DEP_KEYS` works fine for me, but we can rename it to something better. Any suggestions?

rsandell: Yea, that's a good distinction to have; that `GERRIT_*` is data about the event from Gerrit. I'll need to remember that :)
At the same time I think it need to be a bit more distinguishable from any other env var in the mix.

I'm not sure but maybe I just don't like the unnecessary abbreviation of dependency.

Perhaps somewhat mimicing what the [Parameterized Trigger Plugin](https://wiki.jenkins.io/display/JENKINS/Parameterized+Trigger+Plugin#ParameterizedTriggerPlugin-UsageasaBuildstep) provides would be OK? But perhaps not exactly.

`TRIGGER_DEPENDENCY_KEYS` ?

rsandell: And perhaps prefixing the others with `TRIGGER_` as well?

Jimilian: I like this idea. Will switch to `TRIGGER_`.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/dependency/DependencyQueueTaskDispatcher.java
Jimilian: Can I use `actions.replace` for now? I think the list of actions for Queue item should be relatively small (did not measure, btw). So, it will not be very expensive. In this case the rest of code will be the same.
Also it looks like in case of multiple actions we always pickup the first one, so, it means if we have the dependency A-B, we can have the race:
* A passed. This dispatcher added action.
* B was blocked because of other `QueueTaskDispatcher`
* A was re-triggered
* A passed, another dispatcher said ok
* Looks like we should see the first build of A - not a last one.

rsandell: `Actionable` has an [addOrReplace](http://javadoc.jenkins-ci.org/hudson/model/Actionable.html#addOrReplaceAction-hudson.model.Action-) that could work for now. But please add a TODO as well that this needs to be fixed.

Jimilian: It's added only in `2.29`, but `replaceAction` called it inside anyway.

Jimilian:
🕺

PR #320, opened 2017-07-25
fix [JENKINS-45095]

mawinter69:
implement Queuelistener that will cleanup the memory when queue entry is
cancelled.


rsandell:
Thanks for the fix!

PR #326, opened 2017-09-26
Abort ongoing builds of topic if new change to the same topic was submitted

Jimilian:
I'm not sure that this feature would be useful to other teams, but in our case we need to abort on-going builds if new patch-set was added to the topic or the topic itself was changed.

- [x] test is added
- [x] help is added


Jimilian:
@rsandell we found a bug in this implementation, re-trigger does not work for non-topic changes if feature is enabled. will submit fix soon.

shushu2017:
how to re-trigger? multi-checkin as one checkin, verified by jenkins, then show verify +1 on gerrit for each checkin, can it?

PR #328, opened 2017-09-29
Fix NPE during re-triggering in case of "abort same topic" feature is enabled

Jimilian:
Fix NPE that I introduced in https://github.com/jenkinsci/gerrit-trigger-plugin/pull/326.


Jimilian:
@rsandell Sorry for inconvenience, but this one better to merge before releasing next version :) Unfortunately, we didn't mention this bug in our field testing.

Jimilian:
@jenkinsci/hacktoberfest

shushu2017:
git review -t to trigger jenkins, how can I get the code jenkins is checking about?$WORKSPACE only code pushed to git last time.

PR #329, opened 2017-10-10
Fix link to the job in memory report

Jimilian:
Now memory report gives the incorrect url to the job.


Jimilian:
@jenkinsci/hacktoberfest

PR #332, opened 2017-10-25
Do not query Jenkins Core API if it's possible

Jimilian:
`Entity.getProject` uses Jenkins Core API to get the instance of Job. As result
it leads to big memory footprint because `getItemByFullName` needs to pass
ACL checks.

screen shot 2017-10-25 at 14 49 33


Jimilian:
@rsandell thanks for unbelievable speed :)

Unfortunately, matrix plugin does not have this short cut. I will feel the PR for that tomorrow, probably. Plus, I'm not sure that it's always SYSTEM user. At least it's not forced through context.

PR #333, opened 2017-10-25
Fix the deadlock in RunningJobs

Jimilian:
One more deadlock... :)

[threadDump.txt](https://github.com/jenkinsci/gerrit-trigger-plugin/files/1460826/threadDump.txt)

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
@@ -101,7 +101,7 @@ import java.text.MessageFormat;
 import java.text.ParseException;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.HashMap;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedList;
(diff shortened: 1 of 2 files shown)


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -769,7 +769,7 @@ public ListBoxModel doFillVerdictCategoryItems() {
      *
      * @return the store of running jobs.
      */
-    /*package*/ synchronized RunningJobs getRunningJobs() {
+    /*package*/ RunningJobs getRunningJobs() {
         if (runningJobs == null) {
             runningJobs = new RunningJobs();
```
rsandell: Removing the synchronized here means there is risk of dirty write.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -2205,7 +2205,7 @@ public String getHelpFile() {
          * @param parameters the parameters for the new build, used to find it later.
          * @param projectName the name of the current project for better logging.
          */
-        public synchronized void scheduled(ChangeBasedEvent event, ParametersAction parameters, String projectName) {
+        public void scheduled(ChangeBasedEvent event, ParametersAction parameters, String projectName) {
```
rsandell: Removing the `synchronized` means risk of `ConcurrentModificationException` below when iterating over `runningJobs`

rsandell: Well maybe not `ConcurrentModificationException` but there is no guard against the map being manipulated while iterating it iiuc.

Jimilian: I will take a look, this code is in our production for 3 weeks already and everything was fine (at least we did not face deadlocks or CME). But I agree that some guarding can be missed.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
Jimilian: `synchronized` was not needed, `runningJobs` was supposed to be initialised during class creation, but it was not final. So, I made it final and removed the check.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
Jimilian: You were right, we observed non-deterministic behavior during big load (then we start to receive tons of events from here and there). I submitted a new patch, but (!) I think it would be better to wait at least 1 week before merging it :)

Jimilian: @rsandell So far looks good.

Jimilian:
@rsandell can you take a look please?

PR #335, opened 2017-11-20
JENKINS-39498 Fix invalid data

scoheb:


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsPlaybackManager.java
```diff
@@ -155,6 +158,38 @@ protected synchronized Date getDateFromTimestamp() {
         return new Date();
     }
 
+    @Override
+    public void run() {
```
rsandell: Probably better implemented as an [AsyncPeriodicWork](http://javadoc.jenkins.io/hudson/model/AsyncPeriodicWork.html) to save on thread count.

scoheb:
@rsandell Sergii's tests were successful. Would you still like me to refactor to use AsyncPeriodicWork before we merge?

rsandell:
@scoheb It's up to you. I can merge as is, and we can get that nit fixed sometime down the line.

scoheb:
@rsandell refactoring complete. Take a look at your convenience.

rsandell:
Thank you very much @scoheb for looking into this!

PR #336, opened 2017-12-04
JENKINS-48155: gerrit-trigger-plugin does not repsect dynamicTriggerConfiguration flag

Jimilian:
If user does not want to use dynamic configuration anymore, let's assign
the list of projects to empty list to avoid possible race conditions.

Also instead of firing update with empty url, better to check it.

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
@@ -1388,6 +1388,10 @@ public class GerritTrigger extends Trigger<Job> {
      */
     @DataBoundSetter
     public void setDynamicTriggerConfiguration(boolean dynamicTriggerConfiguration) {
+        if (!dynamicTriggerConfiguration) {
+            dynamicGerritProjects = Collections.emptyList();
+        }
+
         this.dynamicTriggerConfiguration = dynamicTriggerConfiguration;
     }
 
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTriggerTimerTask.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTriggerTimerTask.java
@@ -29,6 +29,7 @@ import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 
 import jenkins.model.Jenkins;
+import org.apache.commons.lang.StringUtils;
 
 /**
  * TimerTasks that are created from a GerritTrigger and periodically calls
@@ -57,6 +58,9 @@ public class GerritTriggerTimerTask extends TimerTask {
     public void run() {
         GerritTrigger trigger = getGerritTrigger();
         if (trigger == null) {
+            return;
+        }
+        if (StringUtils.isEmpty(trigger.getTriggerConfigURL())) {
             return;
         }
         trigger.updateTriggerConfigURL();


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritDynamicUrlProcessor.java
```diff
@@ -246,13 +245,6 @@ private static String regexEscapted(char symbol) {
     public static List<GerritProject> fetch(String gerritTriggerConfigUrl, String serverName)
             throws IOException, ParseException {
 
-        if (gerritTriggerConfigUrl == null) {
```
rsandell: I see no reason to remove this, it will just cause another `MalformedURLException` to be thrown from the URL constructor with a more cryptic message.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -1731,14 +1733,6 @@ public void updateTriggerConfigURL() {
             String triggerInformationMessage = MessageFormat.format(
                     "ParseException when fetching dynamic trigger url: {0}", pe.getMessage());
             triggerInformationAction.setErrorMessage(triggerInformationMessage);
-        } catch (MalformedURLException mue) {
-            String logErrorMessage = MessageFormat.format(
-                    "MalformedURLException for project: {0} and URL: {1} Message: {2}",
```
rsandell: Why is this removed? The user could still specify a malformed url in the config screen, and now the error will be harder to find.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -1720,8 +1723,7 @@ public void updateTriggerConfigURL() {
                 triggerInformationAction.setErrorMessage("Dynamic trigger configuration needs "
                         + "a specific configured server");
             } else {
-                List<GerritProject> fetchedProjects = GerritDynamicUrlProcessor.fetch(triggerConfigURL, serverName);
-                dynamicGerritProjects = fetchedProjects;
+                dynamicGerritProjects = GerritDynamicUrlProcessor.fetch(triggerConfigURL, serverName);
```
rsandell: Why this change?
If there is an error now you will be stuck with not triggering on anything instead of at least triggering on what was parsed in the last fetch.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTriggerTimerTask.java
```diff
@@ -59,6 +59,9 @@ public void run() {
         if (trigger == null) {
             return;
         }
+        if (trigger.getTriggerConfigURL() == null || trigger.getTriggerConfigURL().isEmpty()) {
```
rsandell: use `StringUtils.isEmpty()`

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritDynamicUrlProcessor.java
Jimilian: Because updateTriggerConfigURL was trigged only from once place and now we have a check directly in this place.
Should I keep it to double check?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
Jimilian: if `fetch` will raise a exception - assignment will not happen. Do not see any difference - just use one variable less.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
Jimilian: Agree, will restore.

Jimilian:
@rsandell Now it's the most conservative PR ever ;)

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritDynamicUrlProcessor.java
rsandell: Well it's a public method so it's best to keep it I think :)

PR #337, opened 2017-12-05
Reduce memory usage

Jimilian:
Taking the full name of project from Jenkins leads to string duplication
problem as well as involves recursion. So, it's better to limit the
number of requests to Jenkins Core API as much as possible.

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/model/BuildMemory.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/model/BuildMemory.java
@@ -341,8 +341,9 @@ public class BuildMemory {
         if (pb == null) {
             return false;
         } else {
+            String fullName = project.getFullName();
             for (Entry entry : pb.getEntries()) {
-                if (entry.isProject(project)) {
+                if (entry.isProject(fullName)) {
                     return true;
                 }
             }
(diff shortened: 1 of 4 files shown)


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/model/BuildMemory.java
```diff
@@ -1057,15 +1061,15 @@ public String toString() {
...
              *
-             * @param other the other project to check
+             * @param otherName the other project name to check
              * @return true if so.
              * @see #getProject()
              */
-            public boolean isProject(Job other) {
-                if (this.project != null && other != null) {
-                    return this.project.equals(other.getFullName());
+            public boolean isProject(String otherName) {
```
rsandell: the original signature can still be kept and deprecated, don't want to break any scripts or other unknown integrations if we can avoid it.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/TriggeredItemEntity.java
```diff
@@ -201,25 +201,24 @@ public boolean equals(Object obj) {
...
 
         /**
-         * If this object represents the same build as aBuild.
-         * @param aBuild the build.
+         * If this object represents the same build as specified by parameters.
+         * @param otherBuildNumber build number
+         * @param otherParentName project full name
          * @return true if it is so.
          */
-        public boolean equals(Run aBuild) {
```
rsandell: please keep original signature and deprecate

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/TriggeredItemEntity.java
```diff
@@ -201,25 +201,24 @@ public boolean equals(Object obj) {
...
+                    && projectId.equals(otherParentName);
         }
 
         /**
          * If this object represents the same project as aProject.
-         * @param aProject the project to compare.
+         * @param aProjectFullname the project to compare.
          * @return true if it is so.
          */
-        public boolean equals(Job aProject) {
```
rsandell: dito

Jimilian:
@rsandell Done

PR #338, opened 2017-12-05
JENKINS-30557: Jenkins Gerrit Trigger should use relative URL

Jimilian:
Time to time we raise Jenkins master on the host with existing master (for testing or release updates) and keep root url the same. In this setup it becomes a really tricky to configure Gerrit :)


Jimilian:
No way to do it in reliable manner :(

PR #339, opened 2017-12-05
Fix loose forbidden file

Typz:
When `disableStrictForbiddenFileVerification` is set and the 
contains only forbidden files, the build should not be triggered. This
can happen when setting forbidden files but not allowed files, i.e.
marking all files as interesting.

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/GerritProject.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/GerritProject.java
@@ -34,6 +34,7 @@ import hudson.util.ComboBoxModel;
 
 import java.util.Collection;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 
 import org.kohsuke.stapler.DataBoundConstructor;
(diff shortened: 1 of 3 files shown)


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/GerritProject.java
```diff
@@ -219,29 +220,22 @@ public boolean isInteresting(String project, String branch, String topic, List<S
                 boolean foundInterestingTopicOrFile = false;
```
rsandell: I don't think these are used any more?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/GerritProject.java
```diff
@@ -219,29 +220,22 @@ public boolean isInteresting(String project, String branch, String topic, List<S
...
-                    } else {
-                        if (foundInterestingForbidden) {
-                            // we have a forbidden file and a wanted file path.
-                            return false;
-                        } else if (foundInterestingTopicOrFile) {
-                            // we DO not have a forbidden file and but we have a wanted file path.
-                            return true;
-                        }
-                    }
+                    return isInterestingTopic(topic) && !files.isEmpty() && isInterestingFile(files);
```
rsandell: I'm not sure we want to return `false` if there are no changed files. Only if there are files we are interested in and changed files is empty.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/GerritProject.java
Typz: Done

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/GerritProject.java
Typz: The condition you indicate is handled in isInterestingFile().

This test is here because we should return `false` when all files are forbidden, i.e. there are no more after filtering.

Typz: Done

Typz:
@rsandell : I have updated the patch, can you check it again? Thanks,

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/GerritProject.java
```diff
@@ -216,32 +217,29 @@ public boolean isInteresting(String project, String branch, String topic, List<S
...
+                        Iterator<String> i = files.iterator();
+                        while (i.hasNext()) {
+                            String file = i.next();
+                            for (FilePath ffp : forbiddenFilePaths) {
+                                if (ffp.isInteresting(file)) {
+                                    if (!disableStrictForbiddenFileVerification) {
+                                        return false;
+                                    } else {
+                                        foundInterestingForbidden = true;
+                                        i.remove();
```
Jimilian: Potentially this line creates bug. Because https://github.com/sonyxperiadev/gerrit-events/blob/7f3547c6d55946e25e99a847b5160d69e59994ba/src/main/java/com/sonymobile/tools/gerrit/gerritevents/dto/events/ChangeBasedEvent.java caches the list of files and checks it against null. So, one trigger modifies list of files visible for another trigger.

PR #343, opened 2018-01-15
[JENKINS-48943] - Whitelist java.util.Collections entries required by the plugin

oleg-nenashev:
Fix in https://github.com/jenkinsci/gerrit-trigger-plugin/commit/93a771948044fab506b5666ad6778b9636cd9f2c is not complete.

https://issues.jenkins-ci.org/browse/JENKINS-48943

@reviewbybees @rsandell

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,6 @@
 buildPlugin(
     platforms: ['linux'],
+    jenkinsVersions: [null, '2.102']
     jdkVersions: [8],
     findbugs: [run: true, archive: true],
     checkstyle: [archive: true, unstableTotalAll: 0]
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,6 +1,6 @@
 buildPlugin(
     platforms: ['linux'],
-    jenkinsVersions: [null, '2.102']
+    jenkinsVersions: [null, '2.102'],
     jdkVersions: [8],
     findbugs: [run: true, archive: true],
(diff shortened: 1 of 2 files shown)


oleg-nenashev:
@rsandell I cannot setup CI which would verify 2.102, but I confirm that the plugin appears to be fixed after this patch. Would be great to ship it

Jenkinsfile
```diff
@@ -1,5 +1,6 @@
 buildPlugin(
     platforms: ['linux'],
+    jenkinsVersions: [null],
```
jglick: Why? That is the default.

src/main/resources/META-INF/hudson.remoting.ClassFilter
```diff
@@ -0,0 +1,3 @@
+# Remove once https://github.com/jenkinsci/jenkins/pull/3234 is integrated?
```
jglick: link to this PR for convenience

oleg-nenashev:
Should be replaced once https://github.com/jenkinsci/jenkins/pull/3234 is merged

Jenkinsfile
oleg-nenashev: It is a part of my fiasco with Jenkinsfiling tests against 2.102. Maybe we need to update the blogpost, it does not really work in several plugins due to Maven Enforcer. The blogpost should suggest a flag for disabling it

PR #346, opened 2018-01-24
[JENKINS-39498] Stop panicking about "eventCreatedOn" and losing my position

tekkamanendless:
Premise:

When a lot of changes come in within a few seconds of each other, our Jenkins instance would "miss" some of those events.  As far as I can tell, Jenkins somehow received an event with no "eventCreatedOn" timestamp (which is strange because Gerrit never sent such an event, but which may be caused if the event processor receives a null event), which then caused this plugin to switch to "unsupported" mode, deleting the XML flatfile where it tracks which events have been handled.  Almost immediately afterward, the plugin would surprisingly determine that "eventCreatedOn" was being sent, switch to "supported" mode, and be on its merry way.  However, those few events in that window were lost, and my people would send me messages about their change requests not being verified.

This patch:

As far as I can tell, the original code was overly concerned with ever possibly processing an event without "eventCreatedOn", so much so that the _event processing code_ would switch the plugin to "unsupported" mode and then reject the event.

These changes simply remove mode-change ability from the event processing code.  That code should process events, not make configuration changes to the plugin's state.  If an event happens to be received that is missing "eventCreatedOn" (or if an event is passed in that's null for whatever reason), then that event is simply skipped.

In addition, I updated the function that checks for the "events-log" plugin in Gerrit.  Instead of returning true or false, it can now return null when it doesn't know.  This is helpful for situations where Jenkins loses connectivity to Gerrit.  Instead of changing the state of the plugin, this will now merely persist the existing state if a new state cannot be determined.

Edit:
Here's the Jira ticket that was referenced: https://issues.jenkins-ci.org/browse/JENKINS-39498

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsPlaybackManager.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsPlaybackManager.java
@@ -152,8 +152,13 @@ public class GerritMissedEventsPlaybackManager implements ConnectionListener, Na
     public void checkIfEventsLogPluginSupported() {
         GerritServer server = PluginImpl.getServer_(serverName);
         if (server != null && server.getConfig() != null) {
-            isSupported = GerritPluginChecker.isPluginEnabled(
-                    server.getConfig(), EVENTS_LOG_PLUGIN_NAME, true);
+            Boolean newValue = GerritPluginChecker.isPluginEnabled(server.getConfig(), EVENTS_LOG_PLUGIN_NAME, true);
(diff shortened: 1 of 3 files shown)


rsandell:
@scoheb since you are intimate with this part some review help would be much appreciated :)

scoheb:
I have been following the comments on the JIRA issue and I have seen this PR. I will find some time to review it soon.
Thanks @tekkamanendless !

rsandell:
If there is a Jira issue connected to this then please add it to the title and link it in the description.

Ex: '[JENKINS-XXXX] Stop panicking about "eventCreatedOn" and losing my position'

scoheb:
@rsandell Looks good to me

PR #349, opened 2018-02-05
[JENKINS-48622] reduce logging spam from GerritMissedEventsPlaybackEnabledChecker

scoheb:


Jimilian:
btw, we can add somewhere in docs that setting `com.sonyericsson.hudson.plugins.gerrit.trigger.playback.checkEnabledPeriod=600000` before running Jenkins (10 minutes in ms) can mitigate this problem :)

PR #350, opened 2018-02-09
Start the trigger in a known state when using a dynamic configuration URL

tekkamanendless:
In my Jenkins instance, I seemingly randomly have these instances where two patchset-created events, sent seconds apart, have a problem: one of them will be processed normally, and the other will not.

After a whole bunch of logging and debugging, I have traced the problem to the `dynamicGerritProjects` list.  On the event that works, the list will be the proper size; however, on the event that doesn't, it'll be empty.  This is ultimately due to the fact that dynamically configured triggers start with a bad initial state.

This request solves this problem with:
* Trigger restarts should not have a bad initial state.

There are also some logging updates to:
* Generally add some context to otherwise context-free log statements.

----

On trigger restarts...

The `start` action of `GerritTrigger` doesn't actually fetch the dynamic project configuration.  Instead, it starts a thread to do it.  This means that there is a window of time between when `start` returns and when the thread completes its first task that the dynamic project list is empty.

For classic use, this was probably not a huge issue.  However, using imperative Jenkinsfile jobs, this is a major problem because after every job is run, the configuration "changes" (if you use the `properties` command in the job).

In my case, when multiple Gerrit events were submitted in rapid succession, some would be "lost" due to the trigger-restart time window where there were no dynamic projects listed.  In the case of those "lost" events, the trigger would run, and the `isInteresting` method would return false, stating that the event was not relevant to any project or branch in the configuration (logging shows that the length of `dynamicGerritProjects` is zero in those cases).

This commit ensures that the dynamic project list is updated as part of the `start` action so that the trigger is always valid.

----

On logging...

Some of the logs showed some potentially helpful debugging information, but without context.  For example, when determining that an event would be ignored by the trigger, this was printed: `Nothing interesting here, move along folks!`.  This message did not state the event, so looking through my logs to see what went wrong did not provide very much help (I have a lot of jobs triggered with this plugin).  The new statement looks like `Event is not interesting; event: {}`.  Most of the log changes are along this line: adding context such as the event or the job.


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -147,6 +147,8 @@
     private transient RunningJobs runningJobs = new RunningJobs();
     private List<GerritProject> gerritProjects;
     private List<GerritProject> dynamicGerritProjects;
+    //! Lock to control `dynamicGerritProjects`
+    private static final Object DYNAMIC_GERRIT_PROJECTS_LOCK = new Object();
```
rsandell: Are you sure `private final Object DYNAMIC_GERRIT_PROJECTS_LOCK = new Object();` didn't work?
This concerns me as now all Gerrit jobs on the master will need to wait for a single lock.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -1725,7 +1746,22 @@ public void updateTriggerConfigURL() {
                         + "a specific configured server");
             } else {
                 List<GerritProject> fetchedProjects = GerritDynamicUrlProcessor.fetch(triggerConfigURL, serverName);
-                dynamicGerritProjects = fetchedProjects;
+                // Since other threads may be accessing the dynamic project list, we need to make sure
+                // that we update its contents in a thread-safe way.
+                synchronized (DYNAMIC_GERRIT_PROJECTS_LOCK) {
+                    if (dynamicGerritProjects == null) {
+                        logger.debug("Project number changed from null to "
```
rsandell: String concatenation for a debug log is bad. Use `{}` formating instead.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -614,6 +616,12 @@ public void start(Job project, boolean newInstance) {
 
         // Create a new timer task if there is a URL
         if (dynamicTriggerConfiguration) {
+            // We *must* update list of projects immediately *before* we can be labeled as "started".
+            // When using imperative Jenkinsfiles, every run will "update" the job's configuration.
+            // When the job's configuration is updated, all triggers are stopped and restarted.
+            // Thus, if we do *not* update the project list, then Gerrit events in quick succession
+            // will be missed due to the start/stop cycle.
+            updateTriggerConfigURL();
```
rsandell: This will severely increase Jenkins startup and job save times.
We need to find a different approach at solving this problem.
For example adding a flag that `dynamicGerritProjects` is being updated and have `isInteresting` wait for that flag to clear.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
tekkamanendless: Yes; you get a NullPointerException in `isInteresting` every time.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
tekkamanendless: I did, but I got a compile-time error having to do with not being able to use a string as `Marker` or _vice versa_ (I forget which); I suspect that it has to do with the `debug(Marker marker, String format, Object arg1, Object arg2)` signature.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
tekkamanendless: I don't see why it's a problem to have a job's trigger start only when it's ready to start.  I expect the trigger to only be ready when it has loaded the dynamic configuration; I don't think that anyone looking to use this plugin would think that it makes sense for the trigger to be basically useless and unpredictable with Jenkinsfile imperative pipelines.

I suppose that we could have a global cache that maps the URL to its data (or that maps the job to its data) so that the information would persist across trigger restarts.

If you don't want to do that, the only other quick solution that I can think of would be to block within `isInteresting` until the dynamic project list has been loaded (also with a static global lock unless you can figure out how to make Jenkins actually instantiate the local one for `isInteresting`).  I don't know what kind of effects that would have (that is, if `isInteresting` takes seconds to run instead of practically instantly).

rsandell: It's Jenkins startup time that I'm most concerned with. Potentially thousands of jobs wanting to contact external services and retrieve data, that's why it was put in a separate thread in the first place.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
rsandell: Maybe a static Map with job names for keys and `Lock` objects as values?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
Jimilian: > I suppose that we could have a global cache that maps the URL to its data (or that maps the job to its data) so that the information would persist across trigger restarts.

I have this functionality already in my repo. It works perfectly, let me fill PR.
Also look like task that polls dynamic configuration is never stopped, but trigger update fills new task.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -1725,7 +1746,22 @@ public void updateTriggerConfigURL() {
                         + "a specific configured server");
             } else {
                 List<GerritProject> fetchedProjects = GerritDynamicUrlProcessor.fetch(triggerConfigURL, serverName);
-                dynamicGerritProjects = fetchedProjects;
+                // Since other threads may be accessing the dynamic project list, we need to make sure
+                // that we update its contents in a thread-safe way.
```
Jimilian: Sorry, can't get why it was not thread-safe before?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
Jimilian: if you need more than two args you can do `new Object[]{arg1, arg2, arg3}`

Jimilian:
Sorry for feeling my PRs too late.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
rsandell: Having `isInteresting` take a bit of time to run from time to time is ok, it is designed for that so if throughput gets slow you can increase the number of threads serving incoming events. Though in this case it would only perform tiny bit better since in the end all would wait for that one block on the same jobs waiting for the dynamic projects to load. But when that's done it should continue to flow.

rsandell: @tekkamanendless are you OK with using/working around the cache in #353 ? If I should merge that first for you to use?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
tekkamanendless: That would work once the jobs are in the map; are we guaranteed anything about one-thread-per-job or anything like that?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
tekkamanendless: You're right; I had originally tried to lock on `dynamicGerritProjects`, but since it could be null, that was a bad idea.  Since the address of `dyanmicGerritProjects` isn't used for any reason (for example, locking), we can probably go back to the switcheroo here.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
tekkamanendless: Okay, as long as we never set `dynamicGerritProjects` to null once it's already set to something, then we're okay.  I've removed all the lock stuff.

Note that if we ever do set it to null after it was already a list, then we'll run into NPE problems.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
tekkamanendless: This has been reverted.

rsandell:
#353 is now merged, so you can go ahead and adapt it.

tekkamanendless:
I've merged in all the latest changes (the cache proxy); we're down to a line change in "start" and some updated logging lines.

tekkamanendless:
Any guidance on where we want to go from here?  My use case absolutely requires that all jobs be properly configured with their triggers before they run, but I'm willing to accept that I'm a minority in this case.  Would you be happy with a setting that toggles this on somehow?

rsandell:
You need to fix it so that `updateTriggerConfigURL()` isn't called on the main thread during Jenkins startup and slows down startup times. The cache helps but I fear it won't be good enough.
During "normal" operations I can live with having the update run directly in the same thread as whatever called `start` but not during Jenkins startup.

tekkamanendless:
Okay, I think that I understand what you guys are going for.

Here's what I'm thinking would do the trick:

1. The trigger part has to be set up immediately.  From the moment that the job is registered, it needs to be listening for Gerrit events, otherwise it will lose them (it can't stop and start listening).
2. The EventListener should be stripped down to just write every event to a queue (in GerritTrigger).  It won't do any processing; it'll just write them down and move on.
3. A queue-reading thread should attempt to pull items off of the queue as soon as they come in and process them the way that EventListener used to ("isInteresting" and all that).  However, that queue-reading thread will only be active if the dynamic configuration has been loaded (for non-dynamic configurations, it will always be active).

Thus, when the job initially starts up, the queue will be open for business, but the queue-reading thread will not be pulling anything from the queue yet.  Once a configuration has been loaded (non-dynamic configurations are trivially considered "loaded" immediately), the queue-reading thread will go through them, one at a time, and figure out if they should trigger the job.

When jobs are "restarted" (which is upon each and every run if you are running a pipeline job that calls "properties(...)" up front), there will be no discontinuity since no events will be lost.  We will pause queue-reading thread upon "stop" and resume it on "start".

Upon initial startup, there will be no additional delays, since we can now safely afford to have the dynamic-configuration loading happen out-of-band whenever it gets around to it.  The moment that the dynamic configuration is known, all of the events that happened from the moment that "start" was initially called will be processed.

This is the behavior that I am expecting (after a job is present in Jenkins, all events that happen are captured by it), and it meets your startup thread requirements.

If that works for you, I'll get that written and submitted tomorrow.

tekkamanendless:
I have a few weeks of in-house testing to go before I'll be ready with this, but this is generally what I think meets your requirements.

rsandell:
Sorry again for not responding timely.
Putting the events in a queue will be very risky as it will change the triggering behaviour severely especially around dependant triggered jobs. So that is a no go in my mind.

The events are already handled on a thread pool, with each thread handling one event and checking all the jobs, so pausing one of those threads while the dynamic trigger is updating should be safer.
You could potentially achieve that for example by the update thread flipping a semaphore when the update is happening, and when start has been called and then flipping it back when it's done writing the dynamic info. Then you just wait for the semaphore to be open before calling `isInteresting`.

And as I commented before, it will probably be fine to call the update directly in the start method as long as you don't do it during Jenkins startup ('Jenkins.getInstance().getInitLevel() != InitMilestone.COMPLETED'). and maybe do the semaphore only then to simplify.

tekkamanendless:
Okay, so this time I think I have what you want.

I added a latch that the EventListener waits on before calling `isInteresting`.  For static projects, the latch is set to `0` immediately upon start, so there is effectively no waiting.  For dynamic projects, the latch is initialized to `1` upon start and is decreased to `0` when the trigger URL has been fetched.

There's a little bit of logic in there for reusing the same latch between `stop` and `start` calls if the URL didn't change.  For a pipeline job that calls `properties`, it'll `stop` and `start` itself after every run, so it's important to not hold everyone up if the trigger URL part of the job didn't change between runs.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -1441,6 +1479,24 @@ public String getTriggerConfigURL() {
...
+        // was started.  Just make a note of that here; we'll deal with it in "start()".
+        if (triggerConfigURL == null) {
+            if (this.triggerConfigURL != null) {
+                // We need to start fresh when the URL has changed.
+                this.dynamicConfigurationChangedFromLastStart = true;
+            }
+        } else {
+            if (!triggerConfigURL.equals(this.triggerConfigURL)) {
+                // We need to start fresh when the URL has changed.
+                this.dynamicConfigurationChangedFromLastStart = true;
```
rsandell: Have you tested that this actually works? IIRC when configuring the trigger a new instance is created, both via the UI and via pipeline script, so it's not the same instance that gets the `stop`  and `start` calls.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -148,6 +149,14 @@
...
     private List<GerritProject> gerritProjects;
     private List<GerritProject> dynamicGerritProjects;
+    //! This latch will be used to signal that the project list is ready for use.
+    //! For static configuration, this will be ready immediately.
+    //! For dynamic configuration, this will be ready after the first time that
+    //! the project list has been fetched.
+    //!
+    //! Default the latch to the non-waiting zero state, which corresponds to
+    //! static project configurations.
+    private CountDownLatch projectListIsReady = new CountDownLatch(0);
```
rsandell: This might need to be `volatile` and `transient` and then created anew in `readResolve()`.

rsandell:
Yepp, your approach seems good.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
tekkamanendless: No, I haven't begun testing yet.  Once you're happy with the approach (and looks like you are), I'll actually deploy this to my test servers and work through my test cases.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
tekkamanendless: I went with `transient` and just made sure to initialize it in the constructors.

tekkamanendless:
First round of testing has gone well.  I haven't noticed any bad behaviors with a pipeline job calling `properties()` every time that it's run.  All of the events are being handled properly.

Jenkins question for you: am I guaranteed that _one_ of the defined `GerritTrigger` constructors will be called?  As long as a `GerritTrigger` instance cannot be made without calling one of those constructors (I'm mainly concerned with the XML state stuff), then we should be good with the latch.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
tekkamanendless: I had always assumed that it was the same instance.  If you have documented proof of that fact, then we can probably trim some of the new code a bit.

(I've struggled with the lack of up-to-date Jenkins plugin docs; mostly I've just been piecing together how things work from looking at existing plugins and hoping that they're right.)

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
tekkamanendless: Looks like there's still a situation where this can be `null` _and_ the EventListener thread is running.  Looking into where things are getting constructed to see why this is.

tekkamanendless: Okay, doing that in `readResolve` seems to have solved the problem.  It looks like the first run of the plugin is from `readResolve`, and neither of the constructors is called, since the deserializer will only call a zero-argument version of the constructor.

tekkamanendless:
Answer to my own question: either `GerritTrigger(List)`, `GerritTrigger(List, ...)`, or `readResolve()` will be called.  `readResolve()` is magically called when deserializing from XML.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
rsandell: https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/model/AbstractProject.java#L1804-L1808

rsandell:
So how are you progressing? Do you feel that you are done? I don't see any unit tests for example verifying that it works and won't regress in the future.

tekkamanendless:
My changes are ready as far as I'm concerned.

On the unit testing side, I just realized that there are currently no tests covering any cases where `setTriggerConfigURL` is called.  I'll see what I can do about putting together something that'll try to cover that case.  If you have any tips for how you'd like to see that done, I'll try to follow them.  Otherwise, I'll wing it and we can discuss the implementation here.

tekkamanendless:
Quick update: I can make some simple unit tests around making sure that the dynamic project list loads when it's supposed to load with some in-memory HTTP servers or temp files, but I have no idea how to spin up enough of the framework to ensure that the trigger is started and listening for Gerrit events to properly demonstrate that an event received after "start" but before the URL is fetched is processed properly after the URL is fetched.  I'm going to need some serious assistance from someone who has way more experience with unit-testing Jenkins plugins.

tekkamanendless:
I've submitted what I believe to be a halfway-decent unit test to cover this.  Take a look and let me know what else you want.

tekkamanendless:
For the record, if I run the test without my changes, it fails as expected, so I'm happy with it.

```
[ERROR] Failures: 
[ERROR]   GerritTriggerTest.testDynamicTriggerConfigurationTimeGap:1810 
Wanted but not invoked:
toGerritRunListener.onTriggered(
    same(abstractProject),
    same(PatchsetCreated: Change-Id for #1000: Iddaaddaa123456789 PatchSet: 1)
);
-> at com.sonyericsson.hudson.plugins.gerrit.trigger.hudsontrigger.GerritTriggerTest.testDynamicTriggerConfigurationTimeGap(GerritTriggerTest.java:1810)

However, there were other interactions with this mock:
toGerritRunListener.isProjectTriggeredAndIncomplete(
    abstractProject,
    PatchsetCreated: Change-Id for #1000: Iddaaddaa123456789 PatchSet: 1
);
-> at com.sonyericsson.hudson.plugins.gerrit.trigger.hudsontrigger.GerritTrigger.isInteresting(GerritTrigger.java:953)

toGerritRunListener.isTriggered(
    abstractProject,
    PatchsetCreated: Change-Id for #1000: Iddaaddaa123456789 PatchSet: 1
);
-> at com.sonyericsson.hudson.plugins.gerrit.trigger.hudsontrigger.GerritTrigger.isInteresting(GerritTrigger.java:956)
```

tekkamanendless:
Has anyone had a chance to look at the changes made in December (including the tests)?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
rsandell: Did you resolve this?

tekkamanendless: I don't think that I ever saw this comment (sorry).  Is that a confirmation that a trigger instance will _never_ be started again after it has been stopped?  If so, I'll remove all of the restart-related code.

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTriggerTest.java
```diff
@@ -1733,6 +1733,87 @@ public void testDependencyValidationThreeProjectsInvolved() {
...
+
+        // Set up the job within the trigger.
+        Whitebox.setInternalState(trigger, "job", project);
+
+        // We need to make sure that whenever anyone asks for the trigger for any job
+        // that it returns this one.
+        PowerMockito.mockStatic(GerritTrigger.class);
+        // TODO: Why doesn't this work?
+        // TODO: PowerMockito.stub(PowerMockito.method(GerritTrigger.class, "getTrigger")).toReturn(trigger);
+        when(GerritTrigger.getTrigger(any(Job.class))).thenReturn(trigger);
```
rsandell: It might not work because you perhaps need to use PowerMockito's when method?

rsandell:
Hmm it's the `gerritEvent` method that should call `onTriggered` and it seems like you are setting up the mock correctly, so why it isn't called is a bit of a mystery to me as well. Perhaps make sure the trigger object isn't returning `true` for `isSilentMode`?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
tekkamanendless: `dynamicConfigurationChangedFromLastStart` has been completely removed.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
tekkamanendless: This is now done in a separate thread from the `start` thread, and anything that needs to access the dynamic list of projects waits until the list is ready.

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTriggerTest.java
tekkamanendless: I detect a hint of sarcasm there.  From what I had read on `PowerMockito` last year, it was my understanding that the `stub` method was the proper way to handle a static method on a class, but I have no interest in revisiting this since `when` works.  It is what it is.

Comments have been removed.

tekkamanendless:
At this point, I believe that I have addressed everyone's concerns, and my test instance of Gerrit is still happy with everything.

tekkamanendless:
It looks like there was a problem with the Jenkins tooling; it couldn't spin up a node.  Is there a button that you could press to re-trigger the build?
```
[Pipeline] node

Still waiting to schedule task

???

Creating placeholder flownodes because failed loading originals.



GitHub has been notified of this commit’s build result



java.io.IOException: Tried to load head FlowNodes for execution Owner[Plugins/gerrit-trigger-plugin/PR-350/28:Plugins/gerrit-trigger-plugin/PR-350 #28] but FlowNode was not found in storage for head id:FlowNodeId 1:8
```

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
rsandell: yepp, that link confirms that a reconfiguration of a job will create a new trigger object.

rsandell:
Sometimes closing and reopening the pr will make Jenkins start a new build. But now always. Simplest is to push a bogus commit that only changes a comment or something simple like that.

tekkamanendless:
Okay, I removed a couple debug lines that weren't adding anything, and Jenkins actually ran the build this time.

PR #351, opened 2018-02-15
Add option to include matrix results in message

Vynce:
Add an option to matrix jobs that will include the build results of all
child builds in the final build result message that is sent to Gerrit.
The option can be found in Matrix Job -> Gerrit Trigger -> Advanced ->
Include Matrix Results. This option will probably produce too much
output with large matrices, but can be very useful with smaller ones.

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ParameterExpander.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ParameterExpander.java
@@ -37,6 +37,8 @@ import com.sonymobile.tools.gerrit.gerritevents.dto.events.GerritTriggeredEvent;
 import com.sonymobile.tools.gerrit.gerritevents.dto.rest.Notify;
 
 import edu.umd.cs.findbugs.annotations.CheckForNull;
+import hudson.matrix.MatrixBuild;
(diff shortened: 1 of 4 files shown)


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ParameterExpander.java
```diff
@@ -644,6 +646,22 @@ private String createBuildsStats(MemoryImprint memoryImprint, TaskListener liste
                             }
                         }
                     }
+
+                    if (build instanceof MatrixBuild && trigger.getIncludeMatrixResults()) {
```
rsandell: I was hoping to one day be able to yank out all direct dependencies on `MatrixBuild` into optional parts, so that users doesn't have to have the matrix plugin installed for this plugin to work.
But oh well, there are plenty of things in other places to fix for that anyway :)

Vynce: Thanks for taking a look! I'll add some tests.

Vynce:
Since there's no test infrastructure for matrix builds, this is going to take a while :-). I'll submit another PR when I get the tests implemented.

PR #352, opened 2018-02-15
Remove jobs from all memory entities

Jimilian:
Otherwise user will never receive feedback if job trigged by GerritTrigger was removed during validation.


Jimilian:
@rsandell I released that this guy is not merged only recently. Added tests :)

PR #353, opened 2018-02-15
Add proxy to avoid too much connections

Jimilian:
In our environment we have several hundreds jobs, but only few unique dynamic trigger options (10 or 20). This PR has two main goals:
- reduce memory consumption for establishing ssl connection / parsing dynamic trigger config list
- speed dynamic config update time for duplicated entities

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritDynamicUrlProcessor.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritDynamicUrlProcessor.java
@@ -30,6 +30,8 @@ import com.sonyericsson.hudson.plugins.gerrit.trigger.hudsontrigger.data.FilePat
 import com.sonyericsson.hudson.plugins.gerrit.trigger.hudsontrigger.data.GerritProject;
 import com.sonyericsson.hudson.plugins.gerrit.trigger.hudsontrigger.data.Topic;
 
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -53,6 +55,7 @@ import java.util.regex.Pattern;
  *
  * @author Fredrik Abrahamson &lt;fredrik.abrahamson@sonymobile.com&gt;
  */
+@Restricted(NoExternalUse.class)
 public final class GerritDynamicUrlProcessor {
 
     private static final Logger logger = LoggerFactory.getLogger(GerritDynamicUrlProcessor.class);
@@ -126,13 +129,12 @@ public final class GerritDynamicUrlProcessor {
      * Read and parse the dynamic trigger configuration.
      *
      * @param reader stream from which to read the config
-     * @param serverName the name of the Gerrit server configured in the project, could be null.
      *
      * @return List of Gerrit projects
      * @throws ParseException when the fetched content couldn't be parsed
      * @throws IOException for all other kinds of fetch errors
      */
-    private static List<GerritProject> readAndParseTriggerConfig(BufferedReader reader, String serverName)
+    private static List<GerritProject> readAndParseTriggerConfig(BufferedReader reader)
             throws IOException, ParseException {
       Pattern linePattern = buildLinePattern();
 
@@ -238,12 +240,11 @@ public final class GerritDynamicUrlProcessor {
      * since the last fetch, it returns null.
      *
      * @param gerritTriggerConfigUrl the URL to fetch
-     * @param serverName name of the Gerrit server.
      * @return a list of GerritProjects if successful, or null if no change
      * @throws ParseException when the fetched content couldn't be parsed
      * @throws IOException for all other kinds of fetch errors
      */
-    public static List<GerritProject> fetch(String gerritTriggerConfigUrl, String serverName)
+    public static List<GerritProject> fetch(String gerritTriggerConfigUrl)
             throws IOException, ParseException {
 
         if (gerritTriggerConfigUrl == null) {
@@ -264,13 +265,17 @@ public final class GerritDynamicUrlProcessor {
         try {
           instream = connection.getInputStream();
           reader = new BufferedReader(new InputStreamReader(instream, Charset.forName("UTF-8")));
-          return readAndParseTriggerConfig(reader, serverName);
+          return readAndParseTriggerConfig(reader);
         } finally {
-          if (reader != null) {
-            reader.close();
-          } else if (instream != null) {
-            instream.close();
-          }
+            try {
+                if (reader != null) {
+                    reader.close();
+                }
+            } finally {
+                if (instream != null) {
+                    instream.close();
+                }
+            }
         }
     }
 }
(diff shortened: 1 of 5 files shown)


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/CacheProxy.java
```diff
@@ -0,0 +1,79 @@
...
+        if (cache.containsKey(url) && !isExpired(url)) {
+            logger.debug("Get dynamic projects from cache for URL: " + url);
+            return cache.get(url);
+        }
+
+        List<GerritProject> gerritProjects = GerritDynamicUrlProcessor.fetch(url);
+        ttl.put(url, System.currentTimeMillis());
+        cache.put(url, gerritProjects);
+
+        logger.info("Get dynamic projects directly for URL: {}", url);
```
rsandell: Should probably be moved up to before the fetch is performed, or reworded.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritDynamicUrlProcessor.java
```diff
@@ -238,12 +237,11 @@ private static String regexEscapted(char symbol) {
      * since the last fetch, it returns null.
      *
      * @param gerritTriggerConfigUrl the URL to fetch
-     * @param serverName name of the Gerrit server.
      * @return a list of GerritProjects if successful, or null if no change
      * @throws ParseException when the fetched content couldn't be parsed
      * @throws IOException for all other kinds of fetch errors
      */
-    public static List<GerritProject> fetch(String gerritTriggerConfigUrl, String serverName)
+    public static List<GerritProject> fetch(String gerritTriggerConfigUrl)
```
rsandell: I get a strange twitch in my eye whenever a public method signature gets changed without a deprecated old signature left intact.
In this case I doubt anyone external uses it so it is most likely fine.
While you're in here could you add an `@Restricted(NoExternalUse.class)` on the class if that's possible in this version of core otherwise on each public method signature.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -1716,33 +1716,27 @@ public void updateTriggerConfigURL() {
         }
         triggerInformationAction.setErrorMessage("");
         try {
-            if (isAnyServer()) {
-                triggerInformationAction.setErrorMessage("Dynamic trigger configuration needs "
```
rsandell: I can't remember why it needed this? And I get paranoid about it since I spent months tweaking the multi server configurations to not break.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTriggerTimer.java
```diff
@@ -115,7 +115,7 @@ private long calculateDynamicConfigRefreshInterval(@Nonnull GerritTrigger trigge
      *
      * @return the average value.
      */
-    private long calculateAverageDynamicConfigRefreshInterval() {
+    public long calculateAverageDynamicConfigRefreshInterval() {
```
rsandell: `@Restricted(NoExternalUse.class)`

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/CacheProxy.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Keeps map between url and dynamic trigger configuration.
+ * It's used to speed up execution time during updating
+ * trigger jobs and reduce number of connections for the duplicated configs.
+ */
+final class CacheProxy {
```
rsandell: This needs a better name to indicate that it's a proxy for fetching dynamic projects.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
Jimilian: It was needed because very long time ago serverName in `GerritDynamicUrlProcessor.fetch` was really used.

Jimilian:
@rsandell tests were added, comments were addressed.

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/DynamicConfigurationCacheProxyTest.java
```diff
@@ -0,0 +1,106 @@
...
+    @Test
+    public void fetchThroughCache() throws Exception {
+        PowerMockito.mockStatic(GerritDynamicUrlProcessor.class);
+        List<GerritProject> gerritProjects1 = Collections.singletonList(mock(GerritProject.class));
+        List<GerritProject> gerritProjects2 = Collections.singletonList(mock(GerritProject.class));
+        when(GerritDynamicUrlProcessor.fetch(anyString())).thenReturn(gerritProjects1).thenReturn(gerritProjects2);
+
+        List<GerritProject> res1 = DynamicConfigurationCacheProxy.getInstance().fetchThroughCache("someUrl");
+        List<GerritProject> res2 = DynamicConfigurationCacheProxy.getInstance().fetchThroughCache("someUrl");
+
```
rsandell: Should verify that `GerritDynamicUrlProcessor.fetch(` was called on the first call but not the second If I've understood the test correctly.

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/DynamicConfigurationCacheProxyTest.java
```diff
@@ -0,0 +1,106 @@
...
+     */
+    @Test
+    public void fetchDirectlyWithCache() throws Exception {
+        List<GerritProject> gerritProjects1 = Collections.singletonList(mock(GerritProject.class));
+        List<GerritProject> gerritProjects2 = Collections.singletonList(mock(GerritProject.class));
+        when(GerritDynamicUrlProcessor.fetch(anyString())).thenReturn(gerritProjects1).thenReturn(gerritProjects2);
+        setRefreshInternal(FORCE_REFRESH_INTERVAL);
+
+        List<GerritProject> res1 = DynamicConfigurationCacheProxy.getInstance().fetchThroughCache("someUrl");
+        List<GerritProject> res2 = DynamicConfigurationCacheProxy.getInstance().fetchThroughCache("someUrl");
```
rsandell: Should verify that `GerritDynamicUrlProcessor.fetch(` was called on both calls If I've understood the test correctly.

Jimilian: Yes, this is done via `verifyStatic(times(2))`

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/DynamicConfigurationCacheProxyTest.java
Jimilian: Yes, this is done via `verifyStatic()` that is shortcut for `verifyStatic(times(1))`. I also found the name of this check a bit ... misleading.

Jimilian:
@rsandell, please, advise if my understanding of Mockito API is wrong.

rsandell:
@Jimilian you are probably correct, that's why I didn't go any further :)
the reason why I haven't merged yet was because I was waiting for a response from @tekkamanendless just to make sure we didn't break his day :)

Jimilian:
@rsandell thank you!

PR #355, opened 2018-02-15
Do not allocate new array to perform event processing

Jimilian:
Also both getters return unmodifiableList to prevent API misusing.

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
@@ -24,6 +24,7 @@
  */
 package com.sonyericsson.hudson.plugins.gerrit.trigger.hudsontrigger;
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -1346,6 +1347,27 @@ public void setTriggerOnEvents(List<PluginGerritEvent> triggerOnEvents) {
...
     }
 
+    /**
+     * Returns an iterator over the all gerrit projects configured for the trigger.
+     *
+     * @return an iterator over the all gerrit projects configured for the trigger.
+     */
+    private Iterator<GerritProject> getAllGerritProjectsIterator() {
+        if (gerritProjects != null && dynamicGerritProjects != null) {
+            return Iterators.concat(gerritProjects.iterator(), dynamicGerritProjects.iterator());
```
rsandell: I am normally opposed to using guava, but in this case it's using something existing from Jenkins core so it should be fine.

PR #356, opened 2018-03-01
[JENKINS-49813] RunningJobs: Remove extra shouldCancelPatchsetNumber check

aswild:
An extra check for !shouldCancelPatchsetNumber after the check for
(!abortBecauseOfTopic && !shouldCancelpatchsetNumber) breaks the "Abort
patch sets with same topic" feature.
It appears that these lines were added inadvertently during
merge conflict resolution in 7455d86c5d0eaf72743094d64e5faa23b1a5bf82.

Removing this extra if block allows for cancelling running jobs when
a new patchset arrives on the same topic.

Signed-off-by: Allen Wild <allenwild93@gmail.com>


Jimilian:
It was not my best commit :/ See #357 for more interesting things...

@rsandell does it make sense to keep this class as inner-private? maybe we can move RunningJobs to own file with tests, etc?

rsandell:
@Jimilian some refactoring in this area wouldn't hurt :)

PR #358, opened 2018-03-19
Added GERRIT_EVENT_FILE_BASED environment variable

venkrao:
GERRIT_EVENT_FILE_BASED is set to True if the
     build is triggered due to a file-based event.
     Otherwise, False.

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
@@ -143,6 +143,7 @@ public class GerritTrigger extends Trigger<Job> {
      * abort running jobs by setting this to 'false'.
      */
     public static final String JOB_ABORT = GerritTrigger.class.getName() + "_job_abort";
+    private transient GerritTriggerRuntimeParam triggerParamRuntime = new GerritTriggerRuntimeParam();
 
     //! Association between patches and the jobs that we're running for them
     private transient RunningJobs runningJobs = new RunningJobs();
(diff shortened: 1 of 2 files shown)


venkrao:
I can't understand the reason for the test failure. Any help is appreciated. https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fgerrit-trigger-plugin/detail/PR-358/1/pipeline are the logs.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -988,6 +989,10 @@ public boolean isInteresting(GerritTriggeredEvent event) {
                                                     changeBasedEvent.getFiles(
                                                         new GerritQueryHandler(getServerConfig(event))))) {
                                 logger.trace("According to {} the event is interesting.", p);
+                                if (triggerParamRuntime == null) {
+                                        triggerParamRuntime = new GerritTriggerRuntimeParam();
```
rsandell: This is bad as the `GerritTrigger` object is one instance per __job__ so adding a field to it means adding a field to the job not the build.

rsandell:
EVERE: Notify-error:
java.lang.NullPointerException
at com.sonyericsson.hudson.plugins.gerrit.trigger.hudsontrigger.GerritTrigger.getfileBasedMatch(GerritTrigger.java:1671)

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
venkrao: To be honest, I can't think of another way of doing it. Please give me some clues to which file is the right one for this purpose. 

As far as why I thought this was the *only* option: This is the only file that knows about list of files involved in a given Gerrit change(`changeBasedEvent.getFiles`), thus, able to check if the affected-files in a gerrit change match the filtered files of file-based-trigger configuration.

rsandell: You would need to carry the information over from somewhere around here to the build. data in builds are stored via actions, int this case the simplest would be to put the information into the `GerritCause` object that is added to the build when scheduled.
Also, here I don't know if you can be sure the event was interesting because of the file, but maybe :)

Jimilian:
What is FileBasedEvent?

PR #359, opened 2018-04-10
Query the documentation to check if plugin is enabled

deepy:
Instead of querying the plugin (which may give a redirect), query the plugin's Documentation (which does not require authentication).


rsandell:
Can we assume all plugins have documentation?

rsandell:
Is there an issue number for this?

deepy:
I've not created an issue for this, I fixed it locally and just sent the PR upstreams :-)

Based on memory I think it's fairly reliable to go on the documentation, but if not then maybe GerritPluginChecker should be refactored or renamed

rsandell:
I guess you could also make sure to login first? If the redirect is because Jenkins is not logged in iiuc.

PR #360, opened 2018-05-02
Avoid hash code collisions in build memory

Jimilian:
Time to time we observe very strange behaviour - when completely new patchset is triggered, it already has some finished builds in the history. We assume that it may be related due to TreeMap nature of BuildMemory.


Jimilian:
As expected https://github.com/jenkinsci/gerrit-trigger-plugin/pull/360/commits/5198d89516668ffad80800cf067d8a6b8f3aab16 failed https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fgerrit-trigger-plugin/detail/PR-360/1/tests

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/model/BuildMemoryReportContainerTest.java
```diff
@@ -0,0 +1,111 @@
...
+    public static Iterable<Object[]> permutations() {
+        ChangeBasedEvent event1 = new PatchsetCreatedWithPredefinedHash(777);
+        ChangeBasedEvent event2 = new PatchsetCreatedWithPredefinedHash(777);
+        return Arrays.asList(new Object[][]{
+                {null, null, 1},
+                {null, event1, 2},
+                {event1, null, 2},
+                {event1, event1, 1},
+                {event2, event2, 1},
+                {event1, event2, 2},
```
rsandell: Should be one, if they have the same hashcode, they essentially are the same event even though it is different object instances.
An event might be just one instance when a series of builds are triggered, but when those builds are restored from disk the events in those builds will be different instances, and the quick way to see that two instances respresents the same event is to use the `hashCode`.

PR #361, opened 2018-05-11
Abort topic once

Jimilian:
Otherwise same build can be aborted multiple times in case of topic
updates.

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
@@ -2334,18 +2334,17 @@ public class GerritTrigger extends Trigger<Job> {
                     return;
                 }
 
-                // Interrupt any currently running jobs.
-                Jenkins jenkins = Jenkins.getInstance();
-                assert jenkins != null;
+                // Interrupt any currently running builds of the job.
+                Jenkins jenkins = Jenkins.getActiveInstance();
                 for (Computer c : jenkins.getComputers()) {
-                    List<Executor> executors = new ArrayList<Executor>();
+                    List<Executor> executors = new ArrayList<Executor>(c.getExecutors());
                     executors.addAll(c.getOneOffExecutors());
-                    executors.addAll(c.getExecutors());
                     for (Executor e : executors) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -2334,18 +2333,17 @@ private void cancelJob(GerritTriggeredEvent event) {
...
-                    List<Executor> executors = new ArrayList<Executor>();
+                    List<Executor> executors = new ArrayList<Executor>(c.getExecutors());
                     executors.addAll(c.getOneOffExecutors());
-                    executors.addAll(c.getExecutors());
                     for (Executor e : executors) {
                         Queue.Executable currentExecutable = e.getCurrentExecutable();
                         if (currentExecutable != null && currentExecutable instanceof Run<?, ?>) {
                             Run<?, ?> run = (Run<?, ?>)currentExecutable;
-                            if (checkCausedByGerrit(event, run.getCauses())) {
+                            if (run.getParent().equals(job) && checkCausedByGerrit(event, run.getCauses())) {
```
rsandell: Might need to special case matrix builds here.

jadoro:
@Jimilian @rsandell Is there any update on this? We are facing resource leakage issue because exit handlers in Jenkins jobs might not complete due to multiple successive abort events:
If a new patch set is uploaded to a gerrit changeset that triggers *n* builds, each of these builds is aborted *n* times in rapid succession. While the first "abort"-event can be catched to trigger an exit handler, the second one will abort the exit handler itself.

jadoro:
@Jimilian @rsandell Just to keep you updated: We have been using a fork of `gerrit-trigger-plugin` including this PR for the last three months and it indeed fixes the issues we had before. We did not detect any negative side effects on our Jenkins instance.

It would be very nice if we could get this merged somehow; I'll be glad to help if you can direct me to a starting point.

rsandell (approved):

romanek-adam:
Hi @rsandell , I suspect my team is also affected by this issue. When do you expect the next plugin release that will include this fix?

rsandell:
@romanek-adam Sorry, I thought that I had made a release already, thanks for noticing my mistake :)

PR #313, opened 2017-05-22
Merge job started messages based on aggregation interval

Jimilian:
The main goal of this PR is to reduce number of notification user receives from Gerrit Trigger Plugin.
It allows users to specify "aggregation interval", so all jobs that were started in this interval would merge buildStarted message in one.
I.e. you do aggregation interval 10 seconds, after that 3 builds are started:
```
1. build1 at 00:00
2. build2 at 00:09
3. build3 at 00:11
```
PR will merge `build1` and `build2`, so, plugin will send
```
Build started:
build1_url
build2_url
```
After that `build3` will be sent separately:
```
Build started
build3_url
```

We already are using this feature in production (since previous week), looks good.

PR #362, opened 2018-05-15
Fix race condtion in ssh key generation

Jimilian:


Jimilian:
@rsandell, I moved it out of https://github.com/jenkinsci/gerrit-trigger-plugin/pull/313. Please, take a look.

PR #363, opened 2018-05-16
Do not use synchronised checks during event processing

Jimilian:
Using synchronised checks during event processing leads to contention on
GerritWorker threads. Especially if lock is done on ToGerritRunListener.
So, to speed up the processing we need to get rid of
`isProjectTriggeredAndIncomplete`.

`isProjectTriggeredAndIncomplete` can be replaced by `isTriggered` and `isBuilding`.
That transforms the previos version of check to `(isTriggered && isBuilding) || isTriggered`.
So, it's absolutely safe to keep `isTriggered` only.

After that plugin itself does not use `isProjectTriggeredAndIncomplete`.

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ToGerritRunListener.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ToGerritRunListener.java
@@ -206,7 +206,11 @@ public final class ToGerritRunListener extends RunListener<Run> {
      * @param event   the Gerrit Event which is being checked.
      * @param p   the Gerrit project being checked.
      * @return true if so.
+     *
+     * @deprecated Use {@link #isTriggered(Job, GerritTriggeredEvent)} and
+     *                 {@link #isBuilding(Job, GerritTriggeredEvent)}} instead
      */
+    @Deprecated
     public synchronized boolean isProjectTriggeredAndIncomplete(Job p, GerritTriggeredEvent event) {
         if (!memory.isTriggered(event, p)) {
             return false;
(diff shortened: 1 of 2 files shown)


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ToGerritRunListener.java
```diff
@@ -328,18 +332,13 @@ protected void cleanUpGerritCauses(GerritCause firstFound, Run build) {
...
      */
-    public synchronized void onTriggered(Job project, GerritTriggeredEvent event) {
+    public void onTriggered(Job project, GerritTriggeredEvent event) {
         //TODO stop builds for earlier patch-sets on same change.
         memory.triggered(event, project);
         if (event instanceof GerritEventLifecycle) {
             ((GerritEventLifecycle)event).fireProjectTriggered(project);
         }
-        //Logging
-        String name = null;
```
rsandell: Why did you remove this?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ToGerritRunListener.java
```diff
@@ -328,18 +332,13 @@ protected void cleanUpGerritCauses(GerritCause firstFound, Run build) {
      * @param project the project that will be built.
      * @param event   the event that caused the build to be scheduled.
      */
-    public synchronized void onTriggered(Job project, GerritTriggeredEvent event) {
+    public void onTriggered(Job project, GerritTriggeredEvent event) {
         //TODO stop builds for earlier patch-sets on same change.
         memory.triggered(event, project);
         if (event instanceof GerritEventLifecycle) {
```
rsandell: This is a change that might have outside repercussions. The lifecycle listeners might have assumptions about owning the monitor when called _(which is normal behavior in most listener cases that I've seen)_
But I think I can live with that risk if this change is really needed.

PR #365, opened 2018-05-23
Keep direct link to trigger in EventListener

Jimilian:
It would speed up event processing a lot. As scanning all jobs and looking for trigger needs significant amount of time.


rinrinne:
Can this patch avoid an issue "Job was triggered by previous configuration" in race condition? As far as I know, Trigger object is not updated but generated whenever job configuration is applied. (Even if no changes in job configuration)

I remember that a member of Jenkins core team said "trigger object should not be stored outside of job instance".

rsandell:
We used to have it similar to this, but that caused all kinds of problems when holding on to the live instance and changed it to always do a lookup. The lookup shouldn't be that big of a deal since the `ItemGroup`s has the jobs in a map, so it's not like we are iterating over all the jobs all the time. the only iteration done is over the triggers of that one job and that list should in practice always be in single digit length.

PR #372, opened 2018-08-09
Remove GerritMissedEventsPlaybackManager.getServerTimestamp method

hugares:
This getter for serverTimestamp variable was only used in the unit tests
and the variable is already accessed directly in other tests.


hugares:
Hey hey! Are you maintainer of this project now?

dpursehouse:
@rsandell what about this one?

dpursehouse:
Ping @rsandell

rsandell:
As you know I'm drilled into getting nervous about removing public and protected methods after they have been in a public release. So I've hesitantly glancing over this while trying to figure out what to think about it :)

rsandell:
The standard thing to do instead of removing it would be to annotate it with `@Restricted(NoExternalUse.class)` and deprecate it.
That way anyone updating to the newer version will get a compilation error but existing stuff will still work.
could also be event harder annotated with `@Restricted(DoNotUse.class)` and there would be compilation errors internally as well but still keep binary compatibility.

rsandell:
What is the harm in having it there?

PR #374, opened 2018-08-24
Improve Patchset trigger options

arthurbragaa:
This pull request adds the possibility to trigger a job when a draft patchset is created and to ignore otherwise.

To make the configuration a bit clear I removed the exclude drafts option and set the TriggerForPublishedPatchsets and TriggerForDrafts variables enabled by default, keeping the same behavior as before.

There are two things that I didn't manage to do:
1 - This pull request brakes backward compatibility with previous versions of the plugin.

 **Does Jenkins has some kind of migration API?**

2 - I couldn't make the tests in src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spec/BackCompat2173JenkinsTest.java green. 

 **Could someone point me to where are the test setup for the two blocks of tests listed below?**

- https://github.com/jenkinsci/gerrit-trigger-plugin/blob/master/src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spec/BackCompat2173JenkinsTest.java#L200

- https://github.com/jenkinsci/gerrit-trigger-plugin/blob/master/src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spec/BackCompat2173JenkinsTest.java#L128

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/events/PluginPatchsetCreatedEvent.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/events/PluginPatchsetCreatedEvent.java
@@ -32,6 +32,7 @@ import com.sonyericsson.hudson.plugins.gerrit.trigger.events.ManualPatchsetCreat
 import hudson.Extension;
 import hudson.model.Descriptor;
 import hudson.model.Hudson;
+import hudson.util.ListBoxModel;
 import org.jenkinsci.Symbol;
 import org.kohsuke.stapler.DataBoundConstructor;
 
@@ -45,11 +46,23 @@ public class PluginPatchsetCreatedEvent extends PluginGerritEvent implements Ser
     private static final long serialVersionUID = 970946986242309088L;
 
     /**
-     * Uncheck {@link #triggerForDrafts} to have the same effect.
+     * String value of TRIGGER_FOR_ALL.
+     */
+    public static final String TRIGGER_FOR_ALL = "all";
+    /**
+     * String value of TRIGGER_FOR_PUBLISHED.
+     */
+    public static final String TRIGGER_FOR_PUBLISHED = "published";
+    /**
+     * String value of TRIGGER_FOR_DRAFT.
+     */
+    public static final String TRIGGER_FOR_DRAFT = "draft";
+
+    /**
+     * Set {@link #triggerFor} = PluginPatchsetCreatedEvent.TRIGGER_FOR_PUBLISHED to have the same effect.
      */
     protected transient boolean excludeDrafts;
-    private boolean triggerForPublishedPatchsets;
-    private boolean triggerForDrafts;
+    private String triggerFor;
     private boolean excludeTrivialRebase;
     private boolean excludeNoCodeChange;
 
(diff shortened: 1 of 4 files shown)


gizmomogwai:
@arthurbragaa, could you add a test for draft only triggering? there are already several unit-tests that check the flags.

arthurbragaa:
@gizmomogwai It's done. I still didn't figure out how to fix the specs on https://github.com/jenkinsci/gerrit-trigger-plugin/blob/master/src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spec/BackCompat2173JenkinsTest.java#L200

rinrinne:
HI, I assume that BackCompat2173JenkinsTest is the test whether the past configuration annotated with @LocalData is correctly parsed into the latest one. If so, your inserted assertion has the gap with actual value parsed by readResolve().

Maybe you should not update any local data. Regarding @LocalData, please check.
https://wiki.jenkins.io/display/JENKINS/Unit+Test#UnitTest-@LocalData

arthurbragaa:
Hi @rinrinne now it should work. There was a problem with the readResolve method as you pointed out.
Thank you for your help.
@rinrinne @gizmomogwai, Is there something that can be improved on this pull request?

gizmomogwai:
the only thing that could come to mind is, to add tests for updating from the current version of gerrit-trigger to this new version.

gizmomogwai:
another thing that we couldenhance on is the ui part (e.g. a 3 state radiobutton) instead of two checkboxes that allow for one setting that never triggers.

rinrinne:
@arthurbragaa , maybe need to add new BackCompartTest with updated local data. Nothing help your PR, but is would help future PR that breaks BC like this.

arthurbragaa:
Ok, @rinrinne and @gizmomogwai. 
I'll have time to work on that tomorrow.
Thank you for your help

arthurbragaa:
@rinrinne Which tests would you suggest that I write for this change?
The way I understood, there is the following tests that cover this change:
https://github.com/jenkinsci/gerrit-trigger-plugin/blob/c147184b5b0feba85f2072a0969cb5feab44c3bc/src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/event/PluginPatchsetCreatedEventTest.java#L41

https://github.com/jenkinsci/gerrit-trigger-plugin/blob/c147184b5b0feba85f2072a0969cb5feab44c3bc/src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/event/PluginPatchsetCreatedEventTest.java#L58

https://github.com/jenkinsci/gerrit-trigger-plugin/blob/c147184b5b0feba85f2072a0969cb5feab44c3bc/src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/event/PluginPatchsetCreatedEventTest.java#L25

https://github.com/jenkinsci/gerrit-trigger-plugin/blob/c147184b5b0feba85f2072a0969cb5feab44c3bc/src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spec/BackCompat2173JenkinsTest.java#L203

I don't mind creating more tests, but I'm not sure on what to test. Could you give me some guidance on that?

Thank you

rinrinne:
BackCompat2173JenkinsTest can be passed by your implemented readResolve(). But local data tied to BackCompat2173JenkinsTest cannot be reproduced after your implementation is merged. Because no parameter "triggerForDrafts" or "triggerForPublishedPatchsets" exists in config.xml for new created job.

So all you should do is to create new job with your PR once to get generated config.xml. You can archive it as local data zip and add new backcompat test.

rsandell:
So what is happening with this PR? And is it still relevant?

arthurbragaa:
Hi @rsandell Happy new year!
This PR is still relevant for us but I was unable to create a backward compatibility test for it.
I'll try again this week.

PR #375, opened 2018-08-28
Assure the connection get closed after check

eryksz:
ClosableHttpResponse is not being closed if there is no Exception. 
Moved the corresponding code to the finally block.

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/utils/GerritPluginChecker.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/utils/GerritPluginChecker.java
@@ -168,6 +168,8 @@ public final class GerritPluginChecker {
         } catch (Exception e) {
             logger.warn(Messages.PluginHttpConnectionGeneralError(pluginName,
                     e.getMessage()), e);
+            return null;
+        } finally {
             if (execute != null) {
                 try {
                     execute.close();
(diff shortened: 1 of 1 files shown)


dpursehouse:
Hi @rsandell can you have a look at this?  We've found that there are a lot of unclosed connections against Gerrit and this might be one of the causes.

dpursehouse:
On Thu, 30 Aug 2018, 17:41 Robert Sandell, <notifications@github.com> wrote:

> *@rsandell* approved this pull request.
>
> Though if possible it should be converted to a try with resource
> expression instead now that we have java 7 as the language level.
>

Do we? As far as I could see when I was checking yesterday, the level is
still configured as 6, and I didn't see any other try with resource in the
code.

dpursehouse:
I checked again and the `animal-sniffer-maven-plugin` is configured as:

```
                <configuration>
                    <signature>
                        <groupId>org.codehaus.mojo.signature</groupId>
                        <artifactId>java16</artifactId>
                        <version>1.1</version>
                    </signature>
                </configuration>
```

and the README still says "Java 6 runtime compatible", so I don't think we can use `try-with-resource` as it's not Java 6 compatible.

Or is the README and pom.xml out of date?

dpursehouse:
> I thought I had bumped the core dependency to 2.1x

Isn't it about time to do this and move to Java 7 anyway?  Or better, to Java 8? :)

BTW when is the next release (containing this fix) likely to appear?

rsandell:
Yes, it is definitely time. Should take a look at the statistics to see what core version to bump to. IIRC anything newer than 2.11 would imply Java 8.

I'm going through the low hanging fruits for merge and then release this fix.

PR #380, opened 2018-09-06
skip updating the dynamic configuration if jobs are disabled

cashlalala:
There's no need to continuously inquire the dynamic config if job is disabled

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTriggerTimerTask.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTriggerTimerTask.java
@@ -63,6 +63,9 @@ public class GerritTriggerTimerTask extends TimerTask {
         if (StringUtils.isEmpty(trigger.getTriggerConfigURL())) {
             return;
         }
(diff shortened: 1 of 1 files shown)


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTriggerTimerTask.java
```diff
@@ -63,6 +63,9 @@ public void run() {
         if (StringUtils.isEmpty(trigger.getTriggerConfigURL())) {
             return;
         }
+        if (!trigger.getJob().isBuildable()) {
```
rsandell: To be safe, should check `gerritTrigger.getJob() != null` first.

rsandell:
https://ci.jenkins.io/job/Plugins/job/gerrit-trigger-plugin/job/PR-380/1/checkstyleResult/new/

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTriggerTimerTask.java
cashlalala: Fixed. Thx

PR #384, opened 2018-10-03
Make filtering out gerrit user's comments configurable

olofek:
It may be useful, as it was for my team, to be able to trigger builds on comments from any user.  That is currently not possible since filtering out the Gerrit user's comments is hardcoded.

This change makes that behaviour configurable, with the old behaviour the default.


src/main/webapp/help-TriggerOnAllComments.html
```diff
@@ -0,0 +1,14 @@
+<p>
+    By default the Gerrit trigger plugin filters out comments posted by the
+    configured user from the event stream. This means that it is not possible
+    to trigger builds on these messages. Setting this option will give access to
+    the unfiltered stream.
+</p>
+<p>
+    <strong>Note</strong> that with this option enabled, it is possible to create infinite loops.
+    For example, a trigger on a Verified +1 will loop infinitally on successful
+    builds.
```
rsandell: Another reason why we filtered it out as early as possible in the stream was because it caused a significant boost in the amount of events that had to pass through and be checked for if they should trigger a build. So it could slow down the responsiveness in the triggering since there will be at least two more events per triggered build.

olofek: I'm currently on vacation but will review the help text next week. The performance issue should probably be included.

olofek: The help text has been updated with warning about possible performance degradation.  Can't think of anything else that needs adding.

olofek:
Not sure if I'm missing something, but the CI job isn't starting.  Has been pending for almost two weeks now.

PR #385, opened 2018-11-02
Add missing metadata on the maven-surefire-plugin configuration

dpursehouse:
The configuration for the maven-surefire-plugin was missing the groupId and version values.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -227,6 +227,7 @@
     <build>
         <plugins>
             <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
                 <configuration>
                     <systemProperties>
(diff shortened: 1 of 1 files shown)


dpursehouse:
Ah, I was wondering why Eclipse started to show "Overriding managed version" when there aren't any other references to this plugin in the project.

There are a couple more of the same warning.  One for the animal-sniffer plugin (which was updated by me in #377) and junit.  Should those versions then be removed?

rsandell:
possibly, I know I downgraded JUnit at one point because the managed version was giving me problems. But it might work now. Animal sniffer perhaps was upgraded because of some bug? A newer pom parent might have updates?

dpursehouse:
> The animal-sniffer plugin has the same version on the parent as here.

I take it back.  It uses the same version in the pom.xml that was linked above, but in the one currently referenced by this project it's at an earlier version (1.15).

dpursehouse:
Added commits to fix the warnings about animal sniffer in #386.

PR #389, opened 2019-01-10
Move to Java 8 & Jenkins 2.60

framillien:
This change allow use of Java 8 enhancements and  remove compatibility with Java 6 & 7.
Upgrade to Jenkins core 2.60 is needed to expose Java 8 requirement.

Some dependencies are updated to match new jenkins core and jenkins plugin POM parent changes, but many dependencies are still old: we could choose to upgrade them, too.

Dependencies update discussed in PR 383

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.30</version>
+        <version>3.32</version>
         <relativePath />
     </parent>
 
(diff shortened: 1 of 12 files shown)


pom.xml
```diff
@@ -53,17 +53,17 @@
...
+        <java.level>8</java.level>
         <no-test-jar>false</no-test-jar>
         <findbugs.failOnError>false</findbugs.failOnError>
         <concurrency>1C</concurrency>
         <powermock.version>1.6.2</powermock.version>
         <checkstyle.version>2.13</checkstyle.version>
         <mockito.version>1.10.19</mockito.version>
-        <workflow.version>1.14</workflow.version>
-        <workflow.version.test>1.14.1-beta-1</workflow.version.test>
+        <workflow.version>1.14.2</workflow.version>
```
rsandell: I believe we need to bump this to some 2.x version of workflow and split up the versions to whatever latest that supports core 2.60. 1.14 of workflow is ancient.

framillien: All right, I check that.

pom.xml
```diff
@@ -138,6 +137,13 @@
             <version>${workflow.version}</version>
             <optional>true</optional>
         </dependency>
+        <dependency>
+            <!-- Used for test with matrix permissions -->
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>matrix-auth</artifactId>
+            <version>2.3</version>
+            <optional>true</optional>
```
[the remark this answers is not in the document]
framillien: Yes, I fix it.

pom.xml
framillien: I need some advice for this point. If I use workflow-step-api plugin versions targeted to java 8 (core 2.60), I get the last version of this plugin. I can't wonder if it's a problem or not. If we do that, gerrit-trigger-plugin users will be forced to use the last workflow-step-api version to update gerrit-trigger-plugin ?

May be not an issue, but I can't guess ...

Version of workflow-step-api needing java 8 / jenkins core 2.60 : 2.18 (january 2019) (runtime optional)

framillien: workflow-aggregator needed for some tests introduce a lot of dependencies problems (maven enforcer). Do you think there is a good way to replace it for test, or we really need to deal with all depencies management ? Maybe we could get an example of dependencies management from other project using it ?

Or ... keep this versions of workflow plugins ? Since we do need new versions features, and their versions are drived Jenkins plugin management, it's an important issue ?

rsandell: We don't have to go with the latest bleeding edge versions but something 2.x so that debugging etc gets easier to do.
Also as of workflow suite 2.0 there is no longer a mono repo of all the basics of workflow instead each plugin in the suite has it's own release cycle so we can't use the agregator any more.
See #390 for examples.

rsandell:
I've fixed the test errors in #390 and think that's good enough to merge soon. It's this PR plus the workflow updates. And a bump to 2.121 instead of 2.60 which I guess is OK.
When that gets merged this PR should be auto merged as well since all the commits in this PR are in that as well.

PR #391, opened 2019-01-28
Specify a prefix for limiting the list of Gerrit Projects

lucamilanesio:
The GerritServer thread executes an unlimited gerrit ls-projects
at startup which has two major issues:

1. May take a very long time, because the list of projects could
   be huge, even hundreds of thousands of repositories.

2. May cause trouble to the Gerrit Server that has to load *ALL*
   projects in memory to access their configuration and evaluate
   the ACLs.

By introducing a projects prefix pattern, the GerritServer startup
speed could be increased by 20x or even 100x, depending on the
Gerrit server that is connecting to.

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritProjectListUpdater.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritProjectListUpdater.java
@@ -54,8 +54,9 @@ public class GerritProjectListUpdater extends Thread implements ConnectionListen
     /**
      * The command for fetching projects.
      */
-    public static final String GERRIT_LS_PROJECTS = "gerrit ls-projects";
+    public static final String GERRIT_LS_PROJECTS = "gerrit ls-projects --prefix ";
     private static final int MAX_WAIT_TIME = 64;
+    private final String projectPrefix;
 
     private AtomicBoolean connected = new AtomicBoolean(false);
     private boolean shutdown = false;
@@ -66,11 +67,13 @@ public class GerritProjectListUpdater extends Thread implements ConnectionListen
     /**
      * Default constructor.
      * @param serverName the name of the Gerrit server.
+     * @param projectPrefix of the projects to list from the Gerrit server.
      */
-    public GerritProjectListUpdater(String serverName) {
+    public GerritProjectListUpdater(String serverName, String projectPrefix) {
         this.setName(this.getClass().getName() + " for " + serverName + " Thread");
         this.setDaemon(true);
         this.serverName = serverName;
+        this.projectPrefix = projectPrefix;
         addThisAsListener();
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer.java
```diff
@@ -251,6 +252,26 @@ public String getName() {
...
+        return projectPrefix;
+    }
+
+    /**
+     * Set the prefix for fetching all project names.
+     *
+     * @param projectPrefix prefix of the projects to list from the Gerrit Server.
+     */
+    @Exported
+    public void setProjectPrefix(String projectPrefix) {
```
dpursehouse: Maybe I'm missing something, but where does this get called?

If it's not called from anywhere, the projectPrefix uninitialized which results in the full list being returned, i.e. the same behavior as before.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritProjectListUpdater.java
```diff
@@ -54,8 +54,9 @@
     /**
      * The command for fetching projects.
      */
-    public static final String GERRIT_LS_PROJECTS = "gerrit ls-projects";
+    public static final String GERRIT_LS_PROJECTS = "gerrit ls-projects --prefix ";
```
dpursehouse: Note: The `--prefix` option is not documented on the `ls-projects` command for 2.14, 2.15 or 2.16.

dpursehouse (changes requested):
See inline comments.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritProjectListUpdater.java
lucamilanesio: True, this needs to be fixed in Gerrit though.

With regards to the support in Gerrit < 2.13, needs to be documented in this plugin and checked as soon as the version is queried.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer.java
lucamilanesio: For backward compatibility, it will work as before.

Need to find a place to hook into it, do you have any suggestions? I’m not very familiar with this codebase :-(

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritProjectListUpdater.java
lucamilanesio: I've posted the documentation for review https://gerrit-review.googlesource.com/c/gerrit/+/212012

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer.java
rsandell: Yes this uses an outdated way of doing Jenkins form submission. See `public void doConfigSubmit(StaplerRequest req, StaplerResponse rsp) throws ServletException...` in the decriptor.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritProjectListUpdater.java
```diff
@@ -239,7 +242,8 @@ private void tryLoadProjectList() {
                         activeConfig.getGerritProxy(),
                         activeConfig.getGerritAuthentication()
                 );
-                List<String> projects = readProjects(sshConnection.executeCommandReader(GERRIT_LS_PROJECTS));
+                List<String> projects = readProjects(sshConnection
+                        .executeCommandReader(GERRIT_LS_PROJECTS + projectPrefix));
```
rsandell: Should we assume that htis works if `projectPrefix` is an empty string?
In that case I guess it won't work when it's `null`

PR #396, opened 2019-04-10
Ignore replication, replication-done, and reviewer added events

DarrienG:
In accordance with https://issues.jenkins-ci.org/browse/JENKINS-52636

This PR ignores all replication events and reviewer added events. We found these two events bogged down our Jenkins instance running on a machine with 256Gi of RAM and an enterprise grade Intel Xeon. We found when Jenkins was processing all of this data, it ended up just throwing a NoSuchMethodException, trying to reflect into the events it received looking for a method that didn't exist.

For our testing: We are internally running this patch on our prod instance of Jenkins with no issues.

If this change is too controversial, there should be some way to ignore these events from the UI programmatically (and it should probably be the default).


DarrienG:
In the odd case there is contention for this PR, we looked through our logs to see how many replication and reviewer added events would be bogging down Jenkins just a few hours after deployment (and after work hours) and the numbers are astounding:

```bash
[gerrit-server] # cat /var/log/jenkins/jenkins.log | grep ignored | wc -l
120000
```

Our gerrit trigger used to have a backlog for hours when this patch wasn't deployed. Ignoring these events though be the default.

DarrienG:
Just for the sake of it, after another 20 hours:

```
[jenkins-server] # cat /var/log/jenkins/jenkins.log | grep ignored | wc -l
1472364
```

Normally builds would be stalled for a very long time because of all of these events gerrit trigger thinks it can process that it eventually fails on.

pom.xml
```diff
@@ -9,7 +9,7 @@
 
     <groupId>com.sonyericsson.hudson.plugins.gerrit</groupId>
     <artifactId>gerrit-trigger</artifactId>
-    <version>2.29.1-SNAPSHOT</version>
+    <version>2.30.0-uninterested</version>
```
rsandell: ?

DarrienG: Sorry, I didn't mean to commit that to the PR. I just wanted an identifier to know we were internally running something different.

I'll remove when I get into work.

rsandell:
I would be happy if it was configurable with these as defaults.

And if there are as many events bogging down your instance as you say then You should try to stop the event from entering the system much earlier than in `GerritTrigger` as those checks will be done for each configured trigger. If you stopped it earlier in for example `GerritHandler` you would need to do the check only once. The problem there would be that the replication events are needed for another type of listener, so it would break that functionality.
So that means there are two modes for events you want to stop, one set that is global and one set that isn't interesting for Jobs but needs to be trickling down to other listeners, so in those cases `GerritTrigger.isInteresting` might be a good place to put the check.

There is already code in gerrit-events that stops certain events from being parsed.

Christoffer-Cortes:
I am working on a configurable filter solution. The one I'm working on allows for users to set a filter on a per server basis through jenkins gerrit trigger settings page. The servers then takes this setting and changes the boolean interesting in the GerritEventType enum. This stops events from being picked up into the queue in the first place but does require a change in the gerrit-events repo as well. I'm also working on an interval playback setting since we had issues with the disc writing speed slowing down the processing of each event. The solution I'm working on would mean playback can work by writting on each event or based on a timed interval. Setting it to just a 1 second cap is enough to severely drop the queue build up.

DarrienG:
Thanks Chris, I've been allocated for something else at work and haven't been able to find the time to work on this. Feel free to tag me in the PR if you want an extra set of eyes.

Christoffer-Cortes:
I don't know how to tag a user but I have a PR up now: https://github.com/jenkinsci/gerrit-trigger-plugin/pull/397
Most of the logic surrounds how to save and load the actual filter list.

DarrienG:
Thanks Chris. For future reference, you can do `@Christoffer-Cortes` or whoever else to mention them.

PR #397, opened 2019-04-29
EventFilter - Add capability to modify which events are interesting

Christoffer-Cortes:
This adds a new advanced configuration that allows the user to filter which events are interesting. The list is set globally for the gerrit trigger and gerrit event enum is then modified accordingly. It makes use of the already built in filtering with the interesting boolean.

The list configuration uses a javascript in order to use a more user friendly interface. It reads and writes to a hidden text field that is the actual config string.

Requires an update to gerrit-events:
https://github.com/sonyxperiadev/gerrit-events/pull/87


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/config/Config.java
```diff
@@ -410,9 +421,41 @@ public void setValues(JSONObject formData) {
...
+     * Get the filter event list from formdata.
+     * @param formData JSONObject
+     * @return value.
+     */
+    private List<String> getFilterInFromFormData(JSONObject formData) {
+        List<String> filter = new ArrayList<>();
+        String[] arrayIn = new String[0];
+        String stringIn = formData.optString("filterIn");
+        if (stringIn.length() > 0) {
+            arrayIn = stringIn.substring(1, stringIn.length() - 1).split(", ");
```
DarrienG: It looks like you're trying to remove the first and last character? Could you explain why this is necessary?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/config/Config.java
```diff
@@ -1127,6 +1175,25 @@ public void setRestVerified(boolean restVerified) {
         this.restVerified = restVerified;
     }
 
+    @Override
+    public int getEventTypesSize() {
```
DarrienG: This function was added, but it doesn't look like it was used anywhere.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/config/Config.java
Christoffer-Cortes: The string looks something like this: "[one, two, three]". So I remove the first and last character before I split. I don't really need the first and last character though when I create the string. I just made it like that for uniformity. When a List<String> is sent to a text field it is format in the same way. I dont know if you can turn a string like that directly back into a List<String> though.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/config/Config.java
Christoffer-Cortes: It is not used in the java code. It's used in jelly to resize the list depending on the number of available types.

DarrienG: Ditto for interface.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/config/Config.java
DarrienG: Looks like it's a JSON array (assuming I'm reading it correctly)? You it would be better off to use `getJSONArray` instead of parsing yourself. https://stackoverflow.com/a/22074360/2573646

Christoffer-Cortes: A json array wraps strings around quotation marks. I could add this in order to fetch it as a jsonarray or remove the brackets to skip the substring part.

Christoffer-Cortes: Ok so trying to get a json array from the form before it exists will throw an exception. Plus I still want  to handle it as a List so I'd need to convert the JSONArray as well. You could handle the exception and keep it as a JSONArray but the check contains you need to do a toString on the array. No solution is perfect.

DarrienG: That's fair. I figured since the JSON was already parsed, it would make more sense to use the pre-parsed version (especially if the input is big). If there's no way to get an empty list without an exception, it's probably fine either way. I'd just leave a comment near the code to explain what you're doing (and why).

rsandell:
This looks really complicated for something that should be fairly simple with what Jenkins provides in the standard lib. It is also going to complicate things when we finally get the time to modernize the form input to be CasC compliant. For example if the `filterIn` was a setter with a space separated list in a `String` it would make that work a lot easier as it follows a bit closer the standard conventions.

Do you have a screen capture or something to show what it looks like in the UI?

Christoffer-Cortes:
The reason why the setting looks complicated in code is to make it look and work like this in the UI:
settings
What do you think?

rsandell:
@chris-overend it looks great :)
Is it possible to get the data in the form posting to be a whitespace separated string instead of the semi json-array thingie instead? It would make future changes much simpler as it would more conform to what these types of data in Jenkins forms usually look like.
I don't think there will ever be an event type with an "id"  containing a whitespace.

Christoffer-Cortes:
I might have misunderstood your last feedback. Perhaps you meant that it should also be stored as a whitespace separated string list?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/config/EventFilterConfig.java
```diff
@@ -0,0 +1,23 @@
+package com.sonyericsson.hudson.plugins.gerrit.trigger.config;
+
+import java.util.List;
+
+/**
+ * Interface for Event filter configuration.
+ * @author Christoffer Cortes, christoffer.cortes.sjowall@ericsson.com
+ */
+public interface EventFilterConfig {
```
rsandell: It doesn't seem like this interface is needed? Is it a leftover from when you had the logic in gerrit-events?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/config/PluginConfig.java
```diff
@@ -178,4 +201,95 @@ public int getReplicationCacheExpirationInMinutes() {
...
+    private List<String> getFilterInFromFormData(JSONObject formData) {
+        // Getting the list as a JSONArray is not utilized since if the string is empty
+        // the get method will throw an exception. The logic is such that if the filter
+        // is null then the default of every event being interesting is assumed.
+        String stringIn = formData.optString("filterIn");
+        List<String> filter = null;
+        String[] arrayIn = null;
+        if (!stringIn.equals("null")) {
+            filter = new ArrayList<>();
+            arrayIn = stringIn.split(" ");
```
rsandell: might be better to split on `\\s+` to cover any white space.

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/config/PluginConfigTest.java
```diff
@@ -38,33 +43,74 @@
...
+     * filterIn using partial event list.
      */
     @Test
     public void testSetValues() {
+        String events = "change-abandoned change-merged change-restored";
         String formString = "{"
                 + "\"numberOfSendingWorkerThreads\":\"4\","
-                + "\"numberOfReceivingWorkerThreads\":\"6\"}";
+                + "\"numberOfReceivingWorkerThreads\":\"6\","
+                + "\"filterIn\":\"" + events + "\"}";
```
rsandell: Since you are modifying static data in this test (as the tests in gerrit-events) We probably need to make sure they are reset to default between tests to cover our behinds in the future.
I think it can be done with an `@After` method in this class.

rsandell:
I've released gerrit-events 2.16.0 just now so you should be able to update the pom to that version soon when the central sync is complete.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/config/EventFilterConfig.java
Christoffer-Cortes: It's used in PluginConfig. I saw that getter methods had interface stubs so I made them for mine as well.

Christoffer-Cortes: I decided to remove it if that's ok.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/config/PluginConfig.java
Christoffer-Cortes: Done

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/config/PluginConfigTest.java
Christoffer-Cortes: Done

DarrienG:
I'm really excited to see this merged in! Thanks so much for the great work guys!

p3ck:
+1 the plugin is very unstable in it's current form.  Looking forward to this being in the next release.  Any ETA?
Thanks

DarrienG:
I have a patched version in the meantime we're using internally here: https://github.com/DarrienG/gerrit-trigger-plugin/releases/tag/2.31.0-uninterested

It ignores replication and reviewer added events. Since the last rebase with master we haven't hit the queue is building up with builds dropping. We probably have a few thousand builds a day, so you could use it as a decent interim step.

src/main/webapp/help-EventsFilter.html
```diff
@@ -0,0 +1,8 @@
+Filters events that are queued from the gerrit stream events.
+<br>
+If only a few events are intresting to you and you want to reduce load on the queue, then you can control the filter with this setting.
+<br><br>
+<b>Warning</b>
+<br>
+This does not modify the event list in the job configuration view.
+If you use an event from that list that is filtered out by this setting then that trigger will not work.
```
rsandell: A nice future feature could possibly be to automatically filter based on what jobs are configured to trigger on

rsandell:
I've been on vacation so haven't been available to review until now.

@DarrienG is #396 obsolete now?

DarrienG:
Yup, closed!

PR #398, opened 2019-05-02
PlaybackManager - Persist to XML in a thread

Christoffer-Cortes:
In my work place we had issues with the queue getting a build up and delaying jenkins triggers. This is one solution that we propose to reduce the queue build up.

Workers are responsible for writing to XML and the synchronous persist method also writes to disk even when the timestamp has not changed.
The timestamp is then used to ask events-log in gerrit for the missing events.

This change moves the writing to XML to a separate thread and the workers do not have to wait for the disk writing which seems to be the part that takes longest and can cause queue build up.
The thread only checks every second since the smallest increment to the gerrit event timestamp is by whole seconds.

The result is that workers can keep working through the events queue. The persist thread writes consecutively, without a queue and no more than once a second. The downside is that it will check every second even when there is no activity. It could also miss a second in the timestamp.

An alternative solution could be to simple start a thread in what used to be the persist method and have the previous timestamp check there together with a simple boolean semaphore to check that writing is not busy.


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsPlaybackManager.java
```diff
@@ -92,7 +93,8 @@
     /**
      * Server Timestamp.
      */
-    protected EventTimeSlice serverTimestamp = null;
+    protected static EventTimeSlice serverTimestamp = null;
```
rsandell: This will make it not a **server**Timestamp but a global timestamp. Seems possible to mess things up making it static since there is one manager per server.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsPlaybackManager.java
```diff
@@ -554,4 +571,80 @@ public String getDisplayName() {
...
+            GerritMissedEventsPlaybackPersistRunnable.serverName = serverName;
+        }
+
+        /**
+         * Start the persistence thread loop.
+         */
+        public synchronized void start() {
+            if (!isRunning()) {
+                running.set(true);
+                worker = new Thread(this);
```
rsandell: There is `jenkins.util.Timer.get()` for these sorts of things, I'm debating with myself if that one can handle once a second... It probably would I think.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsPlaybackManager.java
Christoffer-Cortes: It's no longer static now.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsPlaybackManager.java
Christoffer-Cortes: I use jenkins scheduler to trigger the runnable every second.

PR #403, opened 2019-10-22
[JENKINS-56528] Always unblock executors waiting for dynamic trigger configuration

Dschoordsch:
Executors are waiting for the project to be ready. This is done so the
list of dynamic trigger projects has finished loading. To ensure the
tasks are not waiting forever, the list must always be set and the latch
released, even when the job is disabled or the trigger was disabled,
fetching of the trigger list failed or similar.

This fixes https://issues.jenkins-ci.org/browse/JENKINS-56528

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
@@ -1754,14 +1754,11 @@ public class GerritTrigger extends Trigger<Job> {
         }
         triggerInformationAction.setErrorMessage("");
         try {
-            dynamicGerritProjects = DynamicConfigurationCacheProxy.getInstance().fetchThroughCache(triggerConfigURL);
-
-            // Now that the dynamic project list has been loaded, we can "count down"
-            // the latch so that the EventListener thread can begin to process events.
-            if (projectListIsReady.getCount() > 0) {
-                logger.debug("Trigger config URL updated: {}; latch is currently {}; decrementing it.", job.getName(),
-                        projectListIsReady.getCount());
-                projectListIsReady.countDown();
+            // Check if dynamic trigger was disabled in the meantime
+            if (!dynamicTriggerConfiguration || job == null || !job.isBuildable()) {
+                dynamicGerritProjects = Collections.emptyList();
+            } else {
+                dynamicGerritProjects = DynamicConfigurationCacheProxy.getInstance().fetchThroughCache(triggerConfigURL);
             }
         } catch (ParseException pe) {
             String logErrorMessage = MessageFormat.format(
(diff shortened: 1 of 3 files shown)


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -1796,6 +1793,14 @@ public void updateTriggerConfigURL() {
...
                     "IOException when fetching dynamic trigger url: {0}", ioe.getMessage());
             triggerInformationAction.setErrorMessage(triggerInformationMessage);
+        } finally {
+            // Now that the dynamic project list has been loaded, we can "count down"
+            // the latch so that the EventListener thread can begin to process events.
+            if (projectListIsReady.getCount() > 0) {
+                logger.debug("Trigger config URL updated: {}; latch is currently {}; decrementing it.", job.getName(),
+                        projectListIsReady.getCount());
+            }
+            projectListIsReady.countDown();
```
rsandell: It looks like this was inside the above if block before, why the move outside of it?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTriggerTimerTask.java
```diff
@@ -60,12 +59,7 @@ public void run() {
         if (trigger == null) {
             return;
         }
-        if (StringUtils.isEmpty(trigger.getTriggerConfigURL())) {
-            return;
-        }
-        if (trigger.getJob() != null && !trigger.getJob().isBuildable()) {
-            return;
-        }
```
rsandell: This will risk Jenkins starting to hammer external services more than before, so why is this needed now that there are changes to the latch handling in `Gerrittrigger`?

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTriggerTest.java
```diff
@@ -1888,6 +1889,128 @@ public void testDynamicTriggerConfigurationTimeGap() throws Exception {
...
+        assertNotNull(Whitebox.getInternalState(trigger, "gerritTriggerTimerTask"));
+
+        // Wait until the timer task has run for the first time.
+        Thread.sleep(GerritTriggerTimer.DELAY_MILLISECONDS + 1000);
+
+        CountDownLatch latch = Whitebox.getInternalState(trigger, "projectListIsReady");
+        assertNotNull(latch);
+        assertEquals(0, latch.getCount());
+    }
+
```
rsandell: You should also try to make a non mocked integration test (just the happy path), my guess it that it would be less complex to write than this version with mocks ;)

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
Dschoordsch: It was moved outside since we need to unblock the worker in any case, otherwise a wrong URL would leave the worker being stuck in the latch indefinitely.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTriggerTimerTask.java
Dschoordsch: I moved this check into `trigger.updateConfigURL`:1758 since that one will release the workers, otherwise they will be blocked forever. These early returns are the underlying issue of the blocked workers. I also left a comment on the next line. I could improve it if you have any ideas.

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTriggerTest.java
Dschoordsch: I can try, checking the blocking condition was more important to me and I just assumed the happy case is already covered.

rsandell:
@Dschoordsch have you seen https://issues.jenkins-ci.org/browse/JENKINS-63000 ?

PR #405, opened 2019-12-02
Remove "previous version warnings" from readme

zbynek:
This content is displayed on the plugin site in the right column using update center API -- that part will be synchronized with security team's recommendations. It is not necessary to have a duplicate in the readme.


timja:
> This content is displayed on the plugin site in the right column using update center API -- that part will be synchronized with security team's recommendations. It is not necessary to have a duplicate in the readme.

I didn't even know that warning icon was clickable =/

PR #406, opened 2019-12-17
Lower logging level of "Notifying BuildStarted" et.al. send to Jenkins log.

jonsten:
If a user start posting very long review comments using gerrit trigger, then the system log for Jenkins becomes unreadable since the entire comment is posted there. This change fixes this by truncating all but the first and last line of the message (and concatenating them on the same line).

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/GerritNotifier.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/GerritNotifier.java
@@ -84,11 +84,11 @@ public class GerritNotifier {
                         parameterExpander.getBuildStartedCommand(build, taskListener, (ChangeBasedEvent)event, stats);
                 if (command != null) {
                     if (!command.isEmpty()) {
-                        logger.info("Notifying BuildStarted to gerrit: {}", command);
+                        logger.debug("Notifying BuildStarted to gerrit: {}", command);
                         cmdRunner.sendCommand(command);
                         GerritTriggeredBuildListener.fireOnStarted(event, command);
                     } else {
-                        logger.info("BuildStarted command is empty.  Gerrit will not be notified of BuildStarted");
+                        logger.debug("BuildStarted command is empty. Gerrit will not be notified of BuildStarted");
                     }
                 } else {
                     logger.error("Something wrong during parameter extraction. "
@@ -114,11 +114,11 @@ public class GerritNotifier {
 
                 if (command != null) {
                     if (!command.isEmpty()) {
-                        logger.info("Notifying BuildCompleted to gerrit: {}", command);
+                        logger.debug("Notifying BuildCompleted to gerrit: {}", command);
                         cmdRunner.sendCommand(command);
                         GerritTriggeredBuildListener.fireOnCompleted(memoryImprint, command);
                     } else {
-                        logger.info("BuildCompleted command is empty.  Gerrit will not be notified of BuildCompleted");
+                        logger.debug("BuildCompleted command is empty. Gerrit will not be notified of BuildCompleted");
                     }
                 } else {
                     logger.error("Something wrong during parameter extraction. "


jonsten:
@TWestling, @GLundh

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/GerritNotifier.java
```diff
@@ -84,7 +84,7 @@ public void buildStarted(Run build, TaskListener taskListener,
                         parameterExpander.getBuildStartedCommand(build, taskListener, (ChangeBasedEvent)event, stats);
                 if (command != null) {
                     if (!command.isEmpty()) {
-                        logger.info("Notifying BuildStarted to gerrit: {}", command);
+                        logger.info("Notifying BuildStarted to gerrit: {}", truncateCommandForLog(command));
```
jonsten: As an alternative we could log without the command on info and then log the command on debug (or lower).

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/GerritNotifier.java
```diff
@@ -129,4 +129,21 @@ public void buildCompleted(MemoryImprint memoryImprint, TaskListener listener) {
...
+     * @param str String to be truncated
+     * @return Truncated string
+     */
+    private static String truncateCommandForLog(String str) {
+        int firstBreak = str.indexOf('\n');
+        int lastBreak = str.lastIndexOf('\n');
+        if (firstBreak == -1) {
+            return str;
+        }
+        return str.substring(firstBreak) + " ... <truncated> ..." + str.substring(lastBreak, str.length());
```
TWestling: Does this work as intended?
line1
line2
line3
would truncate as:
line2
line3 ... <truncated> ... line3

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/GerritNotifier.java
rsandell: Yea, this shouldn't be info level at all I think

jonsten: Let's go that route, in general gerrit trigger is a very chatty plugin, changing this to debug would improve the situation somewhat :)

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/GerritNotifier.java
jonsten: Well this is embarrassing. But let's go with the decreased log level solution instead :)

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/GerritNotifier.java
jonsten: Btw. why are we logging to the logger and not to the listener which ends up in the build log?

rsandell: I'm not sure I can remember correctly that far back. But I think it started as a debug log to monitor that the plugin was working correctly. then someone might have wanted it as an info. Or it was an info to begin with because changing logger configs are hard and then found the utility of it good to have and left it there. And we didn't start with listening to build started, that was added a bit later and I guess was copied from `onCompleted` where we didn't dare logging if the logger was closed or something.
_TL DR:_ No good reason at all :)

jonsten: Thanks for the explanation :)

PR #407, opened 2019-12-18
Some controls ignore current state

testowyludek1234:
'Enable Code-Review' and 'Allow Other Plugins to Contribute Messages' are always selected despite other state in gerrit-trigger.xml. Cause is rather trivial - typos. This is troublesome since any modification must be done with caution otherwise user can override state unconsciously (especially when field is hidden). This patch should fix the issue.

### DIFF
--- a/src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer/index.jelly
+++ b/src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer/index.jelly
@@ -333,7 +333,7 @@
                                               with="gerritFrontEndUrl,gerritHttpUserName,gerritHttpPassword"/>
                             <f:entry title="${%Enable Code-Review}">
                                 <f:checkbox name="restCodeReview"
-                                            checked="${is.config.restCodeReview}"
+                                            checked="${it.config.restCodeReview}"
                                             default="true" />
                             </f:entry>
                             <f:entry title="${%Enable Verified}">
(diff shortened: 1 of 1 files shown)


rsandell:
Tests? Is there a Jira ticket as well that could be mentioned?

testowyludek1234:
I didn't create JIRA ticket, I just uploaded this patch.

To be fair, I have no clue how to provide Unit Tests for this. Is there any existing test case where test is provided for *.jelly file which you could point out as example so I can upload something analogous? The problem I see to provide test for this is that it looks like typical visual bug. For instance checkbox 'Enable Code-Review' is "enabled", you disable it, save configuration, page refreshes and it's still "enabled". Underlying config file have it disabled but as soon as you change anything in it and forgot to disabled it once more it will re-enable it's self during configuration save.

I guess this would be easiest to test during Integration Test via Selenium (you could simulate clicking to disable checkbox, save config and then evaluate if given element have disabled state) but I don't see such infrastructure configured in this project.

testowyludek1234:
I have did some more research and it seems Jenkins have it's own solution to this: j.createWebClient()
and here: https://wiki.jenkins.io/display/JENKINS/Unit+Test#UnitTest-HTMLscraping

I will try to create some tests next week.

rsandell:
OK, no need to create a ticket for a "trivial" bug like this. I just didn't want to loose any ticket if there already was one filed somewhere.

There might already be ["configRountrip"](https://javadoc.jenkins.io/component/jenkins-test-harness/org/jvnet/hudson/test/JenkinsRule.html#configRoundtrip--) tests written that you could just addon to. Though on second thought it is less likely since there is no configRoundtrip alternative to a custom page like this, though I might have written one at one point, hard to remember that far back :)

PR #408, opened 2019-12-19
SSH Review seems stop to working when REST enabled

testowyludek1234:
Sorry I'm uploading empty pull request, but I have no idea how to solve this. Maybe someone has...

I have configured SSH command for success as follows:

`gerrit review <CHANGE>,<PATCHSET> --message 'Something' --code-verification 1`

I have also added new Label Code-Verification on the Gerrit side. When "Enable Rest API" checkbox is **disabled**, vote is placed and everything seems to work.

When I enable "Enable Rest API" and enable "Code-Review" checkbox in this subsection, then --code-verification doesn't work, but that's something I would expected, and "Code-Review" is placed. To my understanding voting was switched to Rest API. But when I disable "Code-Review", there are no votes placed at all.

To summarize my problem I do not see option to have enabled "Rest API" (for missed events to work) and at the same time that SSH commands to be used for voting. To me "Enable Rest API" shouldn't switch to posting reviews using Rest API, but to fetch missed events only (unless some additional config will be provided by plugin and is explicitly checked by user). Also there is no way to influence "vote" payload when REST-API is used.

PS I'm using Jenkins LTS 2.190.3 (not sure what's version of plugin is installed there, latest though update center).


testowyludek1234:
Thank you for reply. When I have more spare time I will try look deeper into this. I would appreciate if this ticket will remain open, maybe someone else will look into this in meantime.

rsandell:
#465

PR #409, opened 2020-01-22
Prevent forbiddenFilePaths from corrupting following filePath matches

peter-walls:
When iterating over multiple jobs set to listen to Gerrit events
a job using forbiddenFilePaths will remove files from the filePaths
that will be checked for jobs that are checked later on. This has
been fixed by changing the file list object to a local copy in the
method doing the check instead of a reference to the object used
later on

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/GerritProject.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/data/GerritProject.java
@@ -32,6 +32,7 @@ import hudson.model.Descriptor;
 import hudson.model.Hudson;
 import hudson.util.ComboBoxModel;
 
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.Iterator;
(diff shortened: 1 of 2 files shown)


peter-walls:
Here is an overview to recreating the error:

Create a job (A) that listens to Gerrit events. Configure it to use forbiddenFilePaths with the following configuration:
[compareType: 'PLAIN', pattern: 'vars/rbGerrit.groovy']

Create a second job (B) that also listens to Gerrit events. Configure it to use filePaths with the following configuration:
[[compareType: 'PLAIN', pattern: 'vars/rbGerrit.groovy']]

Now upload a change to Gerrit where the file vars/rbGerrit.groovy has been changed. If job A is checked first the forbiddenFilePaths setting will remove the file from the file list and when job B is checked it will not find anything to trigger on

rsandell:
@peter-walls Thanks for the fix!
Could you make a unit test out of your description on how to recreate the bug please?

peter-walls:
Yes, I will look at creating a unit test.

peter-walls:
Hello @rsandell I have added a unit test which can be used to verify the fix.

PR #412, opened 2020-03-24
Only store events from configured server

ctudor:
Ran into an issue with missed events and the watchdog timeout when there are multiple gerrit instances defined.  Events were being stored for all defined servers when new events came in which caused the per server data file with the last seen event to have incorrect information for all servers except the one that produced the event.  This patch checks to see if the event is coming from the configured server before processing the event,  I have been running with this change for several weeks now and have not seen any issues.


rsandell:
Thanks! 
Would love to see a test or two in the future.

PR #418, opened 2020-10-27
Add listener for cleaning up removed triggers

jonsten:
If a gerrit trigger is removed from a job, then the listener won't be removed. This caused loads of log entires to be written to the Jenkins system log, warning that the trigger couldn't be found. E.g.:

> Couldn't find a configured trigger for Some/Job

This change fixes this issue by listening for save events on jobs and removing listeners if there is no gerrit trigger configured on the newly updated job.

### DIFF
--- a/src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritItemListenerTest.java
+++ b/src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritItemListenerTest.java
@@ -30,6 +30,7 @@ import com.sonyericsson.hudson.plugins.gerrit.trigger.mock.TestUtils;
 import com.sonymobile.tools.gerrit.gerritevents.GerritHandler;
 import com.sonymobile.tools.gerrit.gerritevents.dto.GerritEvent;
 import com.sonymobile.tools.gerrit.gerritevents.dto.events.PatchsetCreated;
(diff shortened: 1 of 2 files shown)


src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritItemListenerTest.java
```diff
@@ -171,18 +172,53 @@ public void testOnJobDeleted() throws Exception {
...
-     * Subscribes the job to the specified event.
-     * @param job the job.
-     * @param event the event.
+     * Test {@link GerritSaveableListener#onChange(hudson.model.Saveable, hudson.XmlFile)} to handle removed trigger.
+     *
      * @throws Exception if so.
      */
-    public void subscribeJobToEvent(FreeStyleProject job, PatchsetCreated event) throws Exception {
+    @Test
+    public void testOnJobUpdated() throws Exception {
```
jonsten, line 180: This is a little naughty, ideally, since I introduce a new listener, I should also introduce a new test class, however, I took this shortcut since there is an significant amount of boilerplate that needs to be copied from this class. If you wan't I can introduce a new class or rename this one.

TWestling:
@rsandell I am a bit biased here, since we are suffering from the issue that this pull request is fixing, but I would like to get this in, would you mind having a look?

rsandell (approved):

PR #419, opened 2020-10-29
Replace hudson.Plugin with jenkins.model.GlobalConfiguration

olivergondza:
Refactoring to make the JCasC (https://issues.jenkins-ci.org/browse/JENKINS-53250) easier to implement.

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/PluginImpl.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/PluginImpl.java
@@ -443,6 +443,7 @@ public class PluginImpl extends GlobalConfiguration {
      * @return gerritEventManager
      */
     public GerritHandler getHandler() {
+        if (gerritEventManager == null) throw new IllegalStateException("Plugin is not started yet, or it is stopped already");
         return gerritEventManager;
     }
 
@@ -596,7 +597,10 @@ public class PluginImpl extends GlobalConfiguration {
     /**
      * Startup hook.
      */
-    @Initializer(after = InitMilestone.PLUGINS_STARTED)
+    @Initializer(
+            after = InitMilestone.PLUGINS_STARTED, // Requires extensions registered
+            before = InitMilestone.EXTENSIONS_AUGMENTED // Need to be done before jobs start loading
+    )
     @Restricted(DoNotUse.class)
     public static void gerritStart() {
         PluginImpl.getInstance().start();
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/PluginImpl.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/PluginImpl.java
@@ -24,6 +24,7 @@
  */
 package com.sonyericsson.hudson.plugins.gerrit.trigger;
 
+import com.sonyericsson.hudson.plugins.gerrit.trigger.dependency.DependencyQueueTaskDispatcher;
 import com.sonyericsson.hudson.plugins.gerrit.trigger.replication.ReplicationQueueTaskDispatcher;
 import com.sonymobile.tools.gerrit.gerritevents.GerritHandler;
 import com.sonymobile.tools.gerrit.gerritevents.GerritSendCommandQueue;
@@ -498,6 +499,9 @@ public class PluginImpl extends GlobalConfiguration {
         return plugin.getConfiguredJobs(serverName);
     }
 
+    /**
+     * Start the plugin.
+     */
     public void start() {
         logger.info("Starting Gerrit-Trigger Plugin");
         logger.trace("Loading configs");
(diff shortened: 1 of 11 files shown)


olivergondza:
@rsandell, PTAL. Thanks!

olivergondza:
Boy, I hate checkstyle...

olivergondza:
@rsandell, can you please review? I am so looking forward to adding the JCasC support...

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/PluginImpl.java
```diff
@@ -420,12 +431,22 @@ public static void save_() throws IOException {
         plugin.save();
     }
 
+    // Reading from the location where the data used to be back when this implemented hudson.Plugin
+    @Override
+    protected XmlFile getConfigFile() {
+        return new XmlFile(Jenkins.XSTREAM, new File(Jenkins.get().getRootDir(), "gerrit-trigger.xml"));
```
timja, line 437: I had a look around and the only plugin I could find searching https://github.com/search?l=&p=7&q=org%3Ajenkinsci+gerrit.trigger.PluginImpl+fork%3Atrue&ref=advsearch&type=Code

was build failure analyzer and that seems to be using a specific extension point https://github.com/jenkinsci/build-failure-analyzer-plugin/blob/bf0ab28e79ff74a4e9447ba64d031f041c9cc30f/src/main/java/com/sonyericsson/jenkins/plugins/bfa/GerritMessageProviderExtension.java#L44

so this seems safe to me but I could be missing something

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/PluginImpl.java
```diff
@@ -139,7 +150,7 @@ public Api getApi() {
     public static PluginImpl getInstance() {
         Jenkins jenkins = Jenkins.getInstance();
         if (jenkins != null) {
-            return jenkins.getPlugin(PluginImpl.class);
+            return GlobalConfiguration.all().get(PluginImpl.class);
```
timja, line 153: can't you do this?
```suggestion
            return ExtensionList.lookupSingleton(PluginImpl.class);
```

timja (approved):
seems good, not manually tested

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/PluginImpl.java
olivergondza: There are places where powermock is used to mock the response of the static method. So I can do that, but it would require changing more test code, and depending on (power)mocking API out of the plugin source base. Which I choose to avoid.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/PluginImpl.java
olivergondza: Good point in searching the possible impact on other plugins, it would be worth retesting whether we do not break https://github.com/jenkinsci/configuration-as-code-groovy-plugin/blob/78d325d813c33b8515045b7267355b8624cd4508/demos/gerrit-trigger.yml.

Following appears safe too: https://github.com/jenkinsci/sonar-gerrit-plugin/blob/master/src/main/java/org/jenkinsci/plugins/sonargerrit/config/GerritAuthenticationConfig.java

timja: The first will definitely break but groovy compatibility isn’t normally as much of an issue

rsandell (approved):

rsandell:
no actual JCasC test? Or is this just the precursor for making it compatible?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/PluginImpl.java
```diff
@@ -69,7 +79,8 @@
  * @author Robert Sandell &lt;robert.sandell@sonyericsson.com&gt;
  */
 @ExportedBean
-public class PluginImpl extends Plugin {
+@Extension
+public class PluginImpl extends GlobalConfiguration {
```
rsandell, line 83: Might need to return some form of special category so that it isn't included in the normal global configuration? But I guess it should work since it used to be a `Plugin` that should also be listed normally.

olivergondza:
> no actual JCasC test? Or is this just the precursor for making it compatible?

Correct, it was not added yet.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/PluginImpl.java
olivergondza: > But I guess it should work since it used to be a Plugin that should also be listed normally.

Exactly. The empty jelly page is included.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/PluginImpl.java
Gustra: The first did break, eg. the Jenkins.instance.getPlugin(...) call returns `null`. Is there an another way to do this now, or should we pin 2.30.5 until https://issues.jenkins-ci.org/browse/JENKINS-53250 has been implemented?

olivergondza: @Gustra, `PluginImpl.getInstance()` should work before and after this fix.

Gustra: Thank you, @olivergondza. Works like a charm.

liejuntao001:
> > no actual JCasC test? Or is this just the precursor for making it compatible?
> 
> Correct, it was not added yet.

Is my understanding right?
This change prepared for the JCasC support, but the JCasC is not yet supported.
Would you share if the JCasC support will be implemented soon?

PR #421, opened 2020-12-14
Fixed  adding forbidden paths in dynamic config

valentinvarbanov:
When using dynamic files forbidden paths were not working properly. Changes in the forbidden paths triggered builds instead of being ignored.


valentinvarbanov:
@rsandell @TWestling Could you review this change?

rsandell (approved):

rsandell:
Would have been nice to have a test for this as well so the typo doesn't appear again in the future :)

PR #422, opened 2021-01-25
Fix retrieving review message.

thesvistun:
In case of NOT_BUILT status (for ex. when a build was removed from
the queue before it has been started) we get null in both labels:
Verify and Code-Review.
In this case we remove these labels from result command. After that
assuming the message is located between "--message '" and "'  --"
fails due to we got extra spaces after labels has been removed.
As a result, an option following the message is parsed as a part
of the message.

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ParameterExpander.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ParameterExpander.java
@@ -265,7 +265,7 @@ public class ParameterExpander {
         String command = gerritCommand;
         if (r != null && taskListener != null) {
             try {
-                System.out.println("EnvVars: " + r.getEnvironment(taskListener));
+                
                 command = r.getEnvironment(taskListener).expand(command);
             } catch (Exception ex) {
                 logger.error("Failed to expand env vars into gerrit cmd. Gerrit won't be notified!!", ex);
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ParameterExpander.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ParameterExpander.java
@@ -263,6 +263,7 @@ public class ParameterExpander {
         String command = gerritCommand;
         if (r != null && taskListener != null) {
             try {
+                System.out.println("EnvVars: " + r.getEnvironment(taskListener));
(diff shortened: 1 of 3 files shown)


thesvistun:
Hi @rsandell, can you review the PR, please.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ParameterExpander.java
```diff
@@ -704,13 +706,14 @@ public String getBuildStartedMessage(Run build, TaskListener listener, ChangeBas
...
-        String messageStart = "--message '";
-        String fromMessage = completedCommand.substring(completedCommand.indexOf(messageStart));
-        int endIndex = fromMessage.indexOf("' --");
-        if (endIndex <= -1) {
-            endIndex = fromMessage.length();
+        String message = "";
+        String messageRegex = "(?s)--message\\s+'(.*?)'";
+        Pattern p = Pattern.compile(messageRegex);
+        Matcher m = p.matcher(completedCommand);
+        while (m.find()) {
```
romanek-adam, line 713: Sorry but why use `while` instead of `if`?

romanek-adam (commented):
It would be good to write a test case to provide a clear understanding of what this change tries to fix. Also I doubt the maintainer would accept a fix without tests.

rsandell (changes requested):
Some tests are needed.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ParameterExpander.java
thesvistun: A message can be defined more then one time in commands, by a mistake.
In this situation we have at least three cases of how to act:

1. to choose the leftmost message, using `if`;
2. to choose the rightmost message, using `while`;
3. to compose a message out of all the definitions.

I supposed the second case could be the best one here.

rsandell (approved):

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ParameterExpander.java
```diff
@@ -263,6 +265,7 @@ private String expandParameters(String gerritCommand, Run r, TaskListener taskLi
         String command = gerritCommand;
         if (r != null && taskListener != null) {
             try {
+
```
rsandell, line 268:
```suggestion

```

PR #424, opened 2021-02-05
[JENKINS-64783] Better handle of change based event

mobangjack:
See also JENKINS-647883

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
@@ -44,15 +44,20 @@ import com.sonyericsson.hudson.plugins.gerrit.trigger.hudsontrigger.events.Plugi
 import com.sonyericsson.hudson.plugins.gerrit.trigger.hudsontrigger.events.PluginGerritEvent;
 import com.sonyericsson.hudson.plugins.gerrit.trigger.hudsontrigger.events.PluginPatchsetCreatedEvent;
 import com.sonyericsson.hudson.plugins.gerrit.trigger.version.GerritVersionChecker;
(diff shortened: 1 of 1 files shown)


rsandell:
I don't have permission to read that ticket, or it has been removed?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -901,6 +909,88 @@ public boolean equals(Object obj) {
...
+        if (topic == null) {
+            logger.trace("{} has no topic", change);
+            return set;
+        }
+
+        for (GerritServer server : PluginImpl.getServers_()) {
+            topic = new Topic(topic.getName()); // Use a new topic object to avoid getting old changes
+            logger.trace("query topic {} from {} ({}:{})", topic, server.getName(), server.getHostName(),
+                    server.getSshPort());
+            Map<Change, PatchSet> changes = topic.getChanges(new GerritQueryHandler(server.getConfig()));
```
rsandell, line 983: This is not good. Think of a Jenkins controller with thousands of jobs listening to events. For every change based event thousands of jobs would be sending a query to Gerrit, sinking both it and Jenkins.
You need to do as is done with the file paths and only query once for each event and cache it (where the extra info is needed) so that the following jobs in the list won't have to make the query.

rsandell (changes requested):
Since I can't read the ticket, I'm not sure what this is supposed to fix.
It looks like a fundamental change in how the trigger is behaving, so there will need to be some way of retaining the old behavior and opt in to this new one.

rsandell:
Also, please provide tests that should fail with the old and succeed with the new.

mobangjack:
Sorry, my bad. I've given the wrong ticket number. See #425

PR #425, opened 2021-02-09
[JENKINS-64783] Topic Association support

mobangjack:
See https://issues.jenkins.io/browse/JENKINS-64783

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
@@ -28,6 +28,7 @@ import com.google.common.collect.Iterators;
 import com.sonyericsson.hudson.plugins.gerrit.trigger.GerritServer;
 import com.sonyericsson.hudson.plugins.gerrit.trigger.PluginImpl;
 import com.sonyericsson.hudson.plugins.gerrit.trigger.VerdictCategory;
+import com.sonyericsson.hudson.plugins.gerrit.trigger.config.Config;
 import com.sonyericsson.hudson.plugins.gerrit.trigger.config.IGerritHudsonTriggerConfig;
 import com.sonyericsson.hudson.plugins.gerrit.trigger.config.ReplicationConfig;
 import com.sonyericsson.hudson.plugins.gerrit.trigger.dependency.DependencyQueueTaskDispatcher;
@@ -47,12 +48,16 @@ import com.sonyericsson.hudson.plugins.gerrit.trigger.version.GerritVersionCheck
 import com.sonymobile.tools.gerrit.gerritevents.GerritHandler;
 import com.sonymobile.tools.gerrit.gerritevents.GerritQueryHandler;
 import com.sonymobile.tools.gerrit.gerritevents.dto.attr.Approval;
+import com.sonymobile.tools.gerrit.gerritevents.dto.attr.Change;
+import com.sonymobile.tools.gerrit.gerritevents.dto.attr.PatchSet;
 import com.sonymobile.tools.gerrit.gerritevents.dto.attr.Provider;
 import com.sonymobile.tools.gerrit.gerritevents.dto.events.ChangeBasedEvent;
 import com.sonymobile.tools.gerrit.gerritevents.dto.events.CommentAdded;
 import com.sonymobile.tools.gerrit.gerritevents.dto.events.GerritTriggeredEvent;
 import com.sonymobile.tools.gerrit.gerritevents.dto.events.RefUpdated;
 import com.sonymobile.tools.gerrit.gerritevents.dto.events.TopicChanged;
+import com.sonymobile.tools.gerrit.gerritevents.dto.rest.Topic;
+
 import hudson.model.Action;
 import hudson.model.Job;
 import hudson.model.ParametersAction;
@@ -80,6 +85,7 @@ import java.util.Collections;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.CountDownLatch;
 import java.util.regex.PatternSyntaxException;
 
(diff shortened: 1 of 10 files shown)


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -901,6 +908,92 @@ public boolean equals(Object obj) {
...
+            logger.trace("{} has no topic", change);
+            return set;
+        }
+
+        for (GerritServer server : PluginImpl.getServers_()) {
+            topic = new Topic(topic.getName()); // Use a new topic object to avoid getting old changes
+            try {
+                logger.trace("query topic {} from {} ({}:{})", topic, server.getName(), server.getHostName(),
+                        server.getSshPort());
+                Map<Change, PatchSet> changes = topic.getChanges(new GerritQueryHandler(server.getConfig()));
```
mobangjack, line 983: As @rsandell mentioned in https://github.com/jenkinsci/gerrit-trigger-plugin/pull/424:
This is not good. Think of a Jenkins controller with thousands of jobs listening to events. For every change based event thousands of jobs would be sending a query to Gerrit, sinking both it and Jenkins.
You need to do as is done with the file paths and only query once for each event and cache it (where the extra info is needed) so that the following jobs in the list won't have to make the query.

There is more work to be done.

mobangjack (commented):
There is more work to be done

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -901,6 +908,92 @@ public boolean equals(Object obj) {
...
+
+        try {
+            for (GerritServer server : PluginImpl.getServers_()) {
+                gerritQueryHandlers.add(new GerritQueryHandler(server.getConfig()));
+            }
+            Map<Change, PatchSet> changes = topic.getChanges(gerritQueryHandlers);
+            logger.trace("found {} changes with topic {}", changes.size(), topic);
+            for (Change item : changes.keySet()) {
+                logger.trace("{}", item);
+            }
```
rsandell, line 988:
```suggestion
            if (logger.isTraceEnabled()) {
                for (Change item : changes.keySet()) {
                    logger.trace("{}", item);
                }
            }
```
or just remove.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -901,6 +908,92 @@ public boolean equals(Object obj) {
...
+     */
+    private boolean isChangeInteresting(Change change, GerritProject project) {
+        boolean shouldTrigger = false;
+        boolean containsFilePathsOrForbiddenFilePaths = ((project.getFilePaths() != null
+                && project.getFilePaths().size() > 0)
+                || (project.getForbiddenFilePaths() != null && project.getForbiddenFilePaths().size() > 0));
+
+        if (isFileTriggerEnabled() && containsFilePathsOrForbiddenFilePaths) {
+            if (project.isInteresting(change.getProject(), change.getBranch(), change.getTopic(),
+                    change.getFiles(new GerritQueryHandler(getServerByChange(change).getConfig())))) {
```
rsandell, line 943: Instead of doing the somewhat brittle `getServerBychange` you could return a `Map<Change, GerritQueryHandler>` from `getChanges` since in there you already know which querry handler was used for what change.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -901,6 +908,92 @@ public boolean equals(Object obj) {
...
+            return set;
+        }
+
+        List<GerritQueryHandler> gerritQueryHandlers = new ArrayList<GerritQueryHandler>();
+
+        try {
+            for (GerritServer server : PluginImpl.getServers_()) {
+                gerritQueryHandlers.add(new GerritQueryHandler(server.getConfig()));
+            }
+            Map<Change, PatchSet> changes = topic.getChanges(gerritQueryHandlers);
```
rsandell, line 984: compile error here.

rsandell (changes requested):
I am still a bit nervous about changing the trigger behavior this substantially, and I'd rather see this in a separate "mode" that you "activate" in the trigger configuration.

It also needs tests especially one that explains the scenario you are after to fix.

mobangjack:
@rsandell I converted this PR to a draft. There is a related PR needs to be reviewed first: https://github.com/sonyxperiadev/gerrit-events/pull/104.
Please take a look.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
mobangjack: This PR depends on another PR in gerrit-events (https://github.com/sonyxperiadev/gerrit-events/pull/104), please review, thanks

mobangjack:
> I am still a bit nervous about changing the trigger behavior this substantially, and I'd rather see this in a separate "mode" that you "activate" in the trigger configuration.
> 
> It also needs tests especially one that explains the scenario you are after to fix.

A separate mode and required tests were added, please review. This PR needs next gerrit-event release to compile.

mobangjack:
gerrit-events was updated to the required version.

mobangjack:
> I am still a bit nervous about changing the trigger behavior this substantially, and I'd rather see this in a separate "mode" that you "activate" in the trigger configuration.
> 
> It also needs tests especially one that explains the scenario you are after to fix.
A separate mode and required tests were added. Please review.

rsandell (approved):

rsandell:
Thanks for the contribution!
And for your patience in my slow reviewing :)

PR #427, opened 2021-02-19
[JENKINS-64341] Remove the table, tr and td tags from the forms that should not be tables

ironcerocloudbees:
[JENKINS-64639]

I checked the master version running from a Jenkins 2.279 and I found everything is working as in Jenkins 2.263.4 it did. However, there are two jelly files where the table tags were been used to draw fields in the configuration forms. Those table tags are that I removed in the new Jenkins version using the variable divBasedFormLayout.

Additionally, I found someplace where the table tags are used to draw real UI tables inside forms. I believe those tables should continue being tables. I checked and the save actions worked fine in Jenkins 2.279 using those table tags inside the forms.

cc @timja, @fqueiruga, @dariver


timja:
this one looks like there's some history here =/

I can't see anything _wrong_ with the conversion, but lots of the code in there shouldn't be needed at all.

I'm assuming there's some kind of odd layout? or it predates the jelly tags.

in general all the table tags should be changed or altered, tables should not be used for layout

ironcerocloudbees:
@timja,

Thank you so much for your review and your tip. Great idea.

I moved the tag declarations to the lib/plot folder.

Kind regards,

src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer/index.jelly
```diff
@@ -1,6 +1,6 @@
 <?jelly escape-by-default='true'?>
-<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout"
-         xmlns:t="/lib/hudson" xmlns:f="/lib/form" xmlns:i="jelly:fmt">
+<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:l="/lib/layout"
+         xmlns:t="/lib/hudson" xmlns:f="/lib/form" xmlns:i="jelly:fmt" xmlns:p="/lib/plot">
```
timja, line 3: that's a global resources folder, you should use a unique folder for the resources, the convention is `/lib/<plugin-artifact-id>`
```suggestion
         xmlns:t="/lib/hudson" xmlns:f="/lib/form" xmlns:i="jelly:fmt" xmlns:p="/lib/gerrit-trigger">
```

src/main/resources/lib/plot/blockWrapperTrValidationErrorArea.jelly
```diff
@@ -0,0 +1,16 @@
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:d="jelly:define" xmlns:f="/lib/form" xmlns:l="/lib/layout">
+    <!-- TODO remove and switch to div after baseline is 2.264 or newer -->
+    <j:choose>
+        <j:when test="${divBasedFormLayout}">
+            <div>
```
timja, line 6: this is missing it's validation?

This was one of the points I meant that this shouldn't be needed, something is wrong or weirdly laid out if it's needed
```suggestion
            <div class="validation-error-area">
```

src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritServer/index.jelly
ironcerocloudbees: You are right. Thank you so much for the tip. Changes applied.

timja (approved):
the changes here look like they implement a migration correctly.

the (historical) code itself looks wrong and could be improved but that can be separated out to an improvement for latter

fqueiruga (approved):
Code looks solid, pinging @rsandell

src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger/config.jelly
```diff
@@ -242,7 +252,7 @@
                             checkUrl="'descriptorByName/com.sonyericsson.hudson.plugins.gerrit.trigger.hudsontrigger.GerritTrigger/urlCheck?value='+escape(this.value)"/>
                     </f:entry>
                 </f:optionalBlock>
-            </table>
+            </p:blockWrapperTableWidth>
         </f:entry>
```
rsandell, line 257: The actual main configuration form is below this and contains lots of table layouts. What about that?

fbelzunc:
@rsandell Any chance we could perform a release?

rsandell:
I was hoping to get a response to my question. At least that's one of the reasons that I didn't perform a release on Friday :)

PR #428, opened 2021-03-10
JENKINS-65075 Update PluginPatchsetCreatedEvent shouldTriggerOn

bkhouri:
When a Jenkins job is configured with a PatchSet Created, and the
"commit message contains regular expression" is populated, using the
"Query and Trigger Gerrit Patches" link found at
 <jenkins_url>/gerrit_manual_trigger/ will incorrectly trigger the job.

Update the PatchSetCreatedEvent shouldTriggerOn function to check for
the commit message before inspecting if it's a ManualPatchSetCreated
event.

Issue: JENKINS-65075

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/events/PluginPatchsetCreatedEvent.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/events/PluginPatchsetCreatedEvent.java
@@ -225,6 +225,14 @@ public class PluginPatchsetCreatedEvent extends PluginGerritEvent implements Ser
         if (excludeDrafts && ((PatchsetCreated)event).getPatchSet().isDraft()) {
             return false;
         }
+        if (StringUtils.isNotEmpty(commitMessageContainsRegEx)) {
+            if (commitMessagePattern == null) {
+                commitMessagePattern = Pattern.compile(
+                        this.commitMessageContainsRegEx, Pattern.DOTALL | Pattern.MULTILINE);
+            }
+            String commitMessage = ((PatchsetCreated)event).getChange().getCommitMessage();
+            return commitMessagePattern.matcher(commitMessage).find();
+        }
         if (event instanceof ManualPatchsetCreated) {
             // always trigger build when the build is triggered manually
             return true;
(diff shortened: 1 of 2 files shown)


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/events/PluginPatchsetCreatedEvent.java
```diff
@@ -225,6 +225,14 @@ public boolean shouldTriggerOn(GerritTriggeredEvent event) {
         if (excludeDrafts && ((PatchsetCreated)event).getPatchSet().isDraft()) {
             return false;
         }
+        if (StringUtils.isNotEmpty(commitMessageContainsRegEx)) {
+            if (commitMessagePattern == null) {
+                commitMessagePattern = Pattern.compile(
+                        this.commitMessageContainsRegEx, Pattern.DOTALL | Pattern.MULTILINE);
+            }
+            String commitMessage = ((PatchsetCreated)event).getChange().getCommitMessage();
+            return commitMessagePattern.matcher(commitMessage).find();
```
rsandell, line 234: Returning this early will will nullify all the other exclude checks below

rsandell (changes requested):
As I see it, it works as intended.
A better fix would be to allow a custom commit message to be provided when manually triggering and then remove the check for `event instanceof ManualPatchsetCreated`.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/events/PluginPatchsetCreatedEvent.java
bkhouri: True. I didn't think of that.  

In my opinion, the `ManualPatchsetCreated` event should behave just like a `PatchsetCreated` event.  That is, the two triggers should be identical.  it's just that one allow you to retrigger jobs without having to actually create a new Gerrit patch set.

If that is the case, can't we just remove the `event instanced ManualPatchsetCreated` block?

bkhouri (commented):
> As I see it, it works as intended.
> A better fix would be to allow a custom commit message to be provided when manually triggering and then remove the check for `event instanceof ManualPatchsetCreated`.

I don't think allow a custom commit message to be provided when manually triggering the right way of doing this as the commit message should already be available (it should be the commit message of the current patch set).

It is possible for the `ManualPatchsetCreated` to contain t he same information as the `PatchsetCreated` event?

Any thoughts on how to proceed with this?

rsandell:
Yea, I didn't think of that. IIRC you can now get the full message from the query. It could be that before you couldn't get that info or we chose to not include it in the search result in fear of requesting too much data.
Can't remember off the top of my head how the manual patch set is created from the manual trigger. But you probably could just get the result from the query and add it to the created object.
Though if it is created from the json data in the javascript you might need to do some http POST magic to get the text out of it correctly and safely.

PR #429, opened 2021-04-19
[JENKINS-63000] Bypass dynamic trigger fetch after timeout

Gammal-Skalbagge:
The root cause of not opened latch was not found so proposed timeout is only workaround to be able to process events without building up the queue. 

Together with the workaround a modification of dynamic configuration timer handling is proposed to avoid scheduling growing number of timer task every time user apply job config with dynamic trigger enabled.


Gammal-Skalbagge:
Hi, this is kind of stopper for us preventing upgrade to WIP & Private support. Though a timeout seems to be trivial approach and its arbitrary value can be discussed it does not necessarily ruin latch concept. In case of broken connectivity or URL there is no point to block resources endlessly however it would be nice to wait politely for new configuration to apply.
@rsandell  what would you say?

rsandell (approved):

Gammal-Skalbagge:
@rsandell , thank you!

PR #437, opened 2021-05-11
Fix Gerrit Trigger not updating the list of gerrit projects automatically

mariahbb1:
Fix Gerrit Trigger not updating the list of gerrit projects automatically by creating a timer that schedule the update of gerrit project list every five minutes.

The issue that is fixed by this change can be reproduced following the steps below:
Environment requirements:
Jenkins v2.249.2
Gerrit v.2.14.20
Gerrit Trigger v2.34.0
1) Create a new log record:
a) Go to Manage Jenkins > System Log > Add new log recorder
b) In the configure page, in the Name text area, add the name of the logger. (e.g. Gerrit Trigger log)
c) Click on button "Add" to add a new logger. In the Logger text area, add "com.sonyericsson.hudson.plugins.gerrit.trigger.GerritServer" and leave Log level as default "All"
d) Click on button "Add" to add a second new logger. In the Logger text area, add "com.sonyericsson.hudson.plugins.gerrit.trigger.GerritProjectListUpdater" and leave Log level as default "All"

2) Configure Gerrit Trigger Server
a) Go to Manage Jenkins > Gerrit Trigger > Add New Server
b) Insert the information and credentials to your Gerrit Server
c) Click on button "Advanced"
d) In section "Miscellaneous", in Fetch Project List Settings, modify the value of "Project list fetch interval (seconds)" to 60
e) Save the configuration

3) Check the Gerrit Trigger log on Jenkins
a) The following lines should be shown:

Trying to load project list.
May 07, 2021 10:09:00 AM INFO com.sonyericsson.hudson.plugins.gerrit.trigger.GerritProjectListUpdater tryLoadProjectList
Project list from <name of Gerrit Server> contains <number of projects> entries
May 07, 2021 10:09:00 AM INFO com.sonyericsson.hudson.plugins.gerrit.trigger.GerritProjectListUpdater run
ProjectCreated events supported by Gerrit Server <name of Gerrit Server>. Will now listen for new projects...

4) Create new project on Gerrit
a) Login to your Gerrit Server, click on menu Projects > Create New Project
b) Name the project and inherit from the project you wish (e.g. All-Projects)
c) Click on "Create Project"

5) Check the Gerrit Trigger log on Jenkins
a) Issue: No new projects were detected by the plugin. 
b) Wait a few minutes to check if plugin detects the new Gerrit Project.
c) Plugin didn't detected the new Gerrit project.

Plugin will be forever listening for new projects, but no new projects will be added to the Gerrit Trigger project list.

6) Check if new project is listed in the Gerrit Event section
a) Create a new Freestyle Job
b) In "Build Triggers" section, mark the "Gerrit event" option
c) Select your server in "Choose a Server"
d) In the Gerrit Project panel, select Type "Plain" and in Pattern text area, start to write the name of your new project.
e) The auto-completion function will not show your new project, because it is not added to Gerrit Trigger project list.

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritProjectListUpdater.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritProjectListUpdater.java
@@ -220,7 +220,7 @@ public class GerritProjectListUpdater extends Thread implements ConnectionListen
     /**
      * Try to load entire project list from Gerrit server.
      */
-    private void tryLoadProjectList() {
+    public void tryLoadProjectList() {
         int interval = 1;
         while (!isConnected() && !shutdown) {
             logger.info("Not connected to {}, waiting for {} second(s)", serverName, interval);
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritProjectListUpdater.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritProjectListUpdater.java
@@ -152,15 +152,14 @@ public class GerritProjectListUpdater implements ConnectionListener, NamedGerrit
      */
     public void initProjectListUpdater() {
         logger.info("Init project list updater");
-        if(timer == null){
+        if (timer == null) {
             // Never query this Gerrit-server for project list.
             if (!getConfig().isEnableProjectAutoCompletion()) {
                 return;
             }
             timer = new Timer(serverName);
             scheduleProjectListUpdate(getConfig().getProjectListFetchDelay());
-        }
-        else{
+        } else {
             logger.error("Can't create two timers for the same Gerrit instance: " + serverName);
         }
     }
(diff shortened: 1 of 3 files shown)


mariahbb1:
Hi @rsandell , would you please take a look at this PR?

rsandell:
Hi @mariahbb1 thanks for the fix!
I took a look last week, but got forced onto other stuff in the middle of it.

It looks to me like this solution is just a bandaid on top of something that should work. More precisely a Thread added on Top of a Thread, and that doesn't smell good to me :) Sorry for sounding vague, I can't provide any exact technical examples.

There is already a thread that tries to do this and I would rather like to see a fix in there instead of piling something on top. The fix might be to rewrite it into a Timer task.

mariahbb1:
Hi @rsandell 

I modified the GerritProjectListUpdater class to use a timer instead of a Thread. Could you please re-check?

rsandell (approved):

ctudor:
Since this change was merged I am getting failures in the unit tests.  It looks like the default update period of 5 minutes is triggering just before the connection goes down so isConnected() is true in tryLoadProjectList, but the connection is down by the time SshConnectionFactory.getConnection is called so the list is not updated.  Not sure if this is a problem with the code or the test.  The failing test is GerritProjectListUpdaterFunctionalTest.  As a work around I changed the update frequency to 4 and the tests pass, but that is not a good long term fix.  Any help or pointers would be appreciated.

ottofds1:
I'm currently working on this bug and its as @ctudor says, the test timeout is shorter than the period to fetch/update the projects from gerrit, increasing the test timeout fixes the issue but this test already takes a long time to run. My fix was to make the period changeable, since it was fixed at 5 minutes, and decrease the update period to 1 minute in the test so that runs faster, this fix makes the test end 50% faster.

PR #440, opened 2021-05-13
[JENKINS-65481] Add choosing strategy with newrev support

eric-isakson:
When gerrit automatically generates a merge commit, we should be building
the merge commit when the change is merged, not the patchset commit.

Added a new choosing strategy that honors newrev and updated README.adoc with instructions on how to use the new strategy.

I originally attempted modifying the original strategy here: https://github.com/eric-isakson/gerrit-trigger-plugin/tree/choose-newrev-for-change-merged

But that attempt caused build failures for existing jobs when an automatic merge commit is created so I decided to add a new strategy rather than trying to adapt the old one so as not to break existing behavior.

Details here: https://issues.jenkins.io/browse/JENKINS-65481

- [x ] Please describe what you did
- [x ] Link to relevant issues in GitHub or Jira
- [x ] Link to relevant pull requests, esp. upstream and downstream changes
- [x ] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTriggerBuildChooserWithMergeCommitSupport.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTriggerBuildChooserWithMergeCommitSupport.java
@@ -1,0 +1,327 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2010 Andrew Bayer. All rights reserved.
+ * Copyright 2013 Sony Mobile Communications AB. All rights reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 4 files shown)


eric-isakson:
@rsandell the offending field that SpotBugs is complaining about is:

```java
    /**
     * Used by XStream for something.
     */
    @SuppressWarnings("unused")
    private final String separator = "#";
```

Which I copied from GerritTriggerBuildChooser, not sure how to make SpotBugs happy or if this field is still needed. Any advice?

Looks like it was added here:

https://github.com/jenkinsci/gerrit-trigger-plugin/commit/1f8d0dfda1e51725c5f1f6c13d342aef8a1250f1

rsandell:
It is probably not needed in the new class since it has never been serialized before.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTriggerBuildChooserWithMergeCommitSupport.java
```diff
@@ -0,0 +1,327 @@
...
+ * in a backward compatible way.
+ */
+public class GerritTriggerBuildChooserWithMergeCommitSupport extends BuildChooser {
+    private static final long serialVersionUID = 2003462680723330645L;
+
+    /**
+     * Used by XStream for something.
+     */
+    @SuppressWarnings("unused")
+    private final String separator = "#";
```
rsandell, line 80: I don't think you need this. But write some tests to verify just in case. For example a restartable jenkins rule with a config roundtrip of a freestyle projectand then check that it got loaded correctly on the next startup.

rsandell (commented):
Besides the check for the unused field it looks good.

eric-isakson:
Thank you for the feedback. I'll see if I can figure out how to write the test you asked for. Any chance you can point me to something similar that I can derive from? I'm new to this testing framework and don't normally work on Jenkins plugins so I'm going to be digging around the codebase trying to figure that out.

rsandell:
There are plenty of `configRoundTrip(p)` tests in this plugin you can look at. But we haven't used `RestartableJenkinsRule` yet.
But there is an example one here https://github.com/jenkinsci/cloudevents-plugin/blob/0bb2b3079848c07e2158d8cc472d6d42963a545a/src/test/java/io/jenkins/plugins/sample/SampleConfigurationTest.java

And [many more](https://github.com/search?p=2&q=org%3Ajenkinsci+restartablejenkinsrule&type=Code) in other plugins. :)

PR #446, opened 2021-08-21
Reduce usages of Guava

basil:
Reduces usages of Guava in preparation for the upcoming Jenkins core Guava upgrade. CC @rsandell


rsandell (approved):

basil:
@rsandell Would it be possible to get a release of this?

PR #456, opened 2021-11-17
Fix GerritProjectListUpdaterFunctionalTest so that it doesnt time out

ottofds1:
GerritProjectListUpdaterFunctionalTest has a timeout of 500 seconds, but the time that it takes to fetch updates about projects in gerrit is 300 seconds, making the timeout time insufficient to run the full test.

My fix was to change the update period in GerritProjectListUpdater so that it can be customizable. I've also changed the update period used in the test to 1 minute, making so the timeout doesn't interfere with the normal course of the test.

This issue was reproduced and fixed with a windows env, but also works with Linux.

bug appearance:
https://github.com/jenkinsci/gerrit-trigger-plugin/pull/437


ottofds1:
Hi guys,
Could you give a follow up on this pull request?

ottofds1:
Any feedback on the pull request?

PR #430, opened 2021-04-20
Support voting patchsets with the same topic

mobangjack:
- [ ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your master branch!
- [ ] Ensure that the pull request title represents the desired changelog entry
- [ ] Please describe what you did
- [ ] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue

PR #468, opened 2022-07-10
[JENKINS-65349] Feature voting patchsets with the same topic

ckreisl:

This change is a follow up (or second try) of the PR #430 .

This **feature** enables the possibility to send the Gerrit commands also to changes which are assigned to a Gerrit topic.
Right now only the change which has triggered the latest event receives the feedback.

I've created new classes and adapted the way how notifications are send to Gerrit a tiny bit in order to add this feature in a dedicated way.

@mobangjack hope taking your first try as input here and re-upload it as new PR is fine.


ckreisl:
@rsandell any change to get a review of this PR?

ckreisl:
Seems like this plugin is dead :/

jgarver:
@ckreisl I hope it's not dead.  I count on it every day.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/GerritNotifier.java
```diff
@@ -68,6 +73,34 @@ public GerritNotifier(IGerritHudsonTriggerConfig config, GerritCmdRunner cmdRunn
...
+        if (!notification.isValid()) {
+            logger.error("Something wrong during parameter extraction. "
+                    + "Gerrit will not be notified of BuildStarted!");
+            return;
+        }
+
+        NotificationCommands notifyCommands = notification.getCommands();
+        cmdRunner.sendCommand(notifyCommands.getCommand());
+
+        if (notification.isVoteSameTopic() && notifyCommands.hasTopicChanges()) {
```
jgarver, line 97: I wonder if this conditional is necessary.  If it was false, shouldn't notifyCommands.getCommandsTopicChanges() be an empty list?  Requiring the conditional put a burden on the caller to understand a bit about how topics are handled.

jgarver (commented):
Otherwise, looks good to me.

ckreisl:
> @ckreisl I hope it's not dead. I count on it every day.

At least a overall review from the maintainer would be great :)

What concerns me a little bit more is the latest blog post from the Jenkins community: [BlogPost Java11](https://www.jenkins.io/blog/2022/06/28/require-java-11/). Right now the CI/CD pipeline setup for this plugin is based on Java 8. I compiled it locally with Java 11 which breaks due to the `powermock` package which is used within the unit tests.

I'm not sure what the future plans are but i'll assume from the blog post that even Java 17 will be the next target which is then tackled soon - and from my opinion this should be the goal then for this plugin as well. @rsandell if you could find some time it would be great to talk about what are the future plans of this gerrit-trigger-plugin from your end?

rsandell:
I am working on modernizing the plugin parent and all the other dependencies, but it is taking a lot of time and I don't have much spare time to work on it as I have other Jenkins stuff to deal with in my workdays.

basil:
It's running on Java 11 now. :smile:

ckreisl:
> It's running on Java 11 now. 😄

Nice, thanks for the great work! If there is anything we can support or help just give me a ping :)

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/notification/Notification.java
```diff
@@ -0,0 +1,128 @@
...
+        Topic topic = event.getChange().getTopicObject();
+
+        if (topic == null) {
+            logger.error("Topic is null not querying changes!");
+            return Collections.emptyMap();
+        }
+
+        GerritServer server = PluginImpl.getServer_(event);
+        if (server == null) {
+            logger.error("Could get gerrit server based on event!");
```
rsandell, line 103:
```suggestion
            logger.error("Could not get gerrit server based on event!");
```

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/notification/INotification.java
```diff
@@ -0,0 +1,52 @@
+/*
+ *  The MIT License
+ *
+ *  Copyright 2010 Sony Ericsson Mobile Communications. All rights reserved.
+ *  Copyright 2012 Sony Mobile Communications AB. All rights reserved.
```
rsandell, line 5: _nit_ Since you are the author of these new files you should technically add your own copyright (or your employer's depending on your contract).
I don't know if you can give away your copyright like this, but IANAL :shrug: In the git world the code is associated with your email anyway, so the original author can be found.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/GerritNotifierFactory.java
```diff
@@ -50,18 +50,18 @@
  *
  * @author Robert Sandell &lt;robert.sandell@sonyericsson.com&gt;
  */
-public class NotificationFactory {
-    private static final Logger logger = LoggerFactory.getLogger(NotificationFactory.class);
-    private static NotificationFactory instance;
+public class GerritNotifierFactory {
```
rsandell, line 53: Renaming a public class as this is generally frowned upon in Jenkins plugins. It is hard to find out if this is used by any other plugin. And if some other plugin happens to rely upon it they will break since the class is called something else in the version of the plugin they compile against.

Though in the context of this plugin it is highly unlikely that some other plugin would have such a dependency.

rsandell (commented):
Great work! Thanks!
I've relied mostly on the other fantastic reviewers and mostly just skimmed through it. And I've not found anything that would stop me from merging this.
But I'll leave it for a few days to allow you time to address my small comments if you want to.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/GerritNotifier.java
ckreisl: That's right. Removed the additional check and only kept checking if `isVoteSameTopic()` is enabled.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/notification/Notification.java
ckreisl: Done :)

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/notification/INotification.java
ckreisl: I worked on these changes in my spare time. Adapted it for the new files, but I don't mind in general, happy to give something back to the community.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/GerritNotifierFactory.java
ckreisl: I kept it for now in case of complains we can change it back. The reason why I changed the name is that it conflicts with the name of the new `notification` files.

ckreisl:
> Great work! Thanks! I've relied mostly on the other fantastic reviewers and mostly just skimmed through it. And I've not found anything that would stop me from merging this. But I'll leave it for a few days to allow you time to address my small comments if you want to.

Thanks! I've added one additional check that Gerrit changes which are already ABANDONED are not considered since I saw some SSH exceptions in the log if the plugin tries to add a +1/-1 verified to abandoned change sets. Not sure if this is the "best" way to tackle it but there are some other things I try to tackle which concern me a little bit.

This new option helps but requires other options like "Abort patchsets with same topic" be enabled as well, otherwise there might be conflicts from my point of view and right now i'm not sure how to tackle this.
E.g.
Repo A -> (triggers) -> verify-a
Repo B -> (triggers) -> verify-b

Patchset 1 (on repo A) & Patchset 2 (on repo B) combined as a Gerrit Topic. If the required option is disabled the feedback will be:
verify-a -> (reports result to) -> Repo A (main) & Repo B (due to topic)
verify-b -> (reports result to) -> Repo B (main) & Repo A (due to topic)

The fastest job wins, so could be that one reports -1 verified in the first place and the second one +1 verified which will override the previous result.

rsandell:
@ckreisl Perhaps att a warning/info in the help file, and perhaps even a description on the field in the form mentioning this limitation for now. Would that be enough?

rsandell (approved):

PR #474, opened 2022-10-06
Align PLAIN matching with docs

phoenikz:
Documentation states that PLAIN matches with case sensitivity, but the implementation is with case insensitivity. This aligns implementation with documentation

- [x ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [ x] Ensure that the pull request title represents the desired changelog entry
- [ x] Please describe what you did
- [ x] Link to relevant issues in GitHub or Jira
- [ x] Link to relevant pull requests, esp. upstream and downstream changes
- [ x] Ensure you have provided tests - that demonstrates feature works or fixes the issue


phoenikz:
Mutually exclusive with #475, proposing that maintainers select one and reject one - I have no opinion how it should work, just that is should be consistent

rsandell:
Better to keep the existing behavior and update the documentation instead. So I'll go for #475

PR #476, opened 2022-10-06
Fix SSH connection leaks

ride2ubaye:
When the Configuration As Code changes the instances of GerritServer can be refreshed, before that we need to stop the server to be sure the opened SSH connections are properly released.
Otherwise it is possible to create a 'zombie' SshConnection objects detached from the Gerrit Server and so never released.

How to reproduce:
Go to Manage Jenkins menu and open Configuration As Code and click on the button Reload existing configuration.
Once the reload is performed look at the connection to the port 29418, there is a new connection but the previous is not closed.
(e.g. on centos: ss -tr dport = 29418)

This behavior was observed with our Gerrit Server version 2.14.2 and Jenkins Master 2.346.2.

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/config/GerritJcascConfigurator.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/config/GerritJcascConfigurator.java
@@ -47,7 +47,6 @@ import java.util.Calendar;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/config/GerritJcascConfigurator.java
```diff
@@ -91,7 +90,12 @@ public Class<PluginImpl> getTarget() {
     protected void configure(
         Mapping config, PluginImpl instance, boolean dryrun, ConfigurationContext context
     ) throws ConfiguratorException {
-        List<GerritServer> oldServers = new ArrayList<>(instance.getServers());
+        if (!dryrun) {
+            for (GerritServer oldServer : instance.getServers()) {
+                oldServer.stopConnection();
```
rsandell, line 95: But if `super.configure(config, instance, dryrun, context);` now fails you will be left with the old configuration but stopped connections.

rsandell (changes requested):

rsandell (approved):

PR #478, opened 2022-10-24
[JENKINS-67761] Fix cancelling jobs multiple times

ckreisl:

In case N (N > 1) jobs are triggered via one event the build will be aborted N times if the build policy "Only build latest patchset" is enabled. Related to [JENKINS-67761](https://issues.jenkins.io/browse/JENKINS-67761). Instead of parsing `null` to `cancelMatchingJobs(...)` as `jobName` we actually parse the correct job name which should be cancelled.

In addition related to [JENKINS-69338](https://issues.jenkins.io/browse/JENKINS-69338) "zombie" leftovers in the queue marked with "Stopped" will be cancelled properly. During debugging noticed that `items.getCauses()` returns an empty list in case the build is a Pipeline project. Maybe there is a better way how we can handle it but right now I just added one additional cancel call before the actual thread is terminated.

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/RunningJobs.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/RunningJobs.java
@@ -8,6 +8,7 @@ import static com.sonyericsson.hudson.plugins.gerrit.trigger.hudsontrigger.Gerri
 import com.sonyericsson.hudson.plugins.gerrit.trigger.hudsontrigger.data.BuildCancellationPolicy;
 import com.sonymobile.tools.gerrit.gerritevents.dto.events.ChangeBasedEvent;
 import com.sonymobile.tools.gerrit.gerritevents.dto.events.GerritTriggeredEvent;
+import hudson.model.FreeStyleProject;
 import hudson.model.Cause;
 import hudson.model.Computer;
 import hudson.model.Executor;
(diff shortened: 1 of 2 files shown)


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/RunningJobs.java
```diff
@@ -229,23 +235,29 @@ private void cancelMatchingJobs(GerritTriggeredEvent event, String matchOnJobNam
...
-                       Run<?, ?> run = (Run<?, ?>)currentExecutable;
-                       if (checkCausedByGerrit(event, run.getCauses())) {
-                           if (matchOnJobName == null || matchOnJobName.equals(run.getParent().getFullName())) {
-                               e.interrupt(
-                                       Result.ABORTED,
-                                       new NewPatchSetInterruption()
-                               );
-                           }
-                       }
+                   if (!(this.job instanceof FreeStyleProject)) {
```
ckreisl, line 257: Need to check on this again, since there could be other job types as well...

rsandell: There are some others it could be, like `MatrixProject`. Probably checking if it is `AbstractProject` would work and cover all of them except `WorkflowJob`.

ckreisl: Do you have any clue why `item.getCauses()` does not work for `hudson.model.Queue$BuildableItem:ExecutorStepExecution.PlaceholderTask` queue items? Even if the build is triggered / scheduled via the Plugin the causes are missing at least this is what I got for WorkflowJobs.

ckreisl: Just did some tests and as soon as the build is not displayed as "part of xyz" we can actually get the assigned causes. Since I'm not so deep into the whole Jenkins dev yet I try to understand the different behaviors and states in the queue.
However with this additional check we still get leftovers in the queue marked as "Stopping part of  xyz".
I think some additional logic is required to really take care of that. At least for now the job is aborted just once and not multiple times.

rsandell (approved):

panicking:
@ckreisl this change or the follow one add some regression as explain in https://issues.jenkins.io/browse/JENKINS-70162. Do you have some suggestion how fix it?

ckreisl:
@panicking related to the linked issue it can only be this change which broke the behavior since the `RunningJobs.java` class was not touched by the other changes I did. Before a new version is released we might need to revert the change or provide a proper fix. The bug which pops up now is not covered by a unit or integration that's why at least i missed it. So if you want to provide a fix cover the broken behavior with a test first.

PR #480, opened 2022-11-17
Provide commit message in manual queries

muffgaga:
[Gerrit 3.4.6](https://www.gerritcodereview.com/3.4.html#346) (cf. [gerrit issue 15941](https://bugs.chromium.org/p/gerrit/issues/detail?id=15941)) defaults to not show commit messages unless `--commit-message` is provided.
This yields a mismatch between "automatically" and "manually" triggered changesets in the plugin (`GERRIT_CHANGE_COMMIT_MESSAGE` missing in the latter case). 
This commit switches to a different overload of `GerritQueryHandler`'s `queryJava()` providing this functionality.

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/actions/manual/ManualTriggerAction.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/actions/manual/ManualTriggerAction.java
@@ -332,7 +332,7 @@ public class ManualTriggerAction implements RootAction {
             session.setAttribute("queryString", queryString);
 
             try {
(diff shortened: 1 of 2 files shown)


rsandell (approved):

rsandell:
The tests that are failing are the tests for the class that you have made changes to, so it deserves an investigation.

muffgaga:
> The tests that are failing are the tests for the class that you have made changes to, so it deserves an investigation.

Yep, and I did initially forget to modify the query commands the mocked gerrit server expects (we added `--commit-message` to the query, as this was the implicit default before Gerrit 3.6.4).

muffgaga:
@rsandell it's green :D…

PR #481, opened 2022-11-18
[JENKINS-70111] Provide options to prevent triggering jobs when Change status != NEW

ckreisl:

When enabling the option "Enable Topic Association" within the job configurations every job will be triggered as soon as the topic is of interest. This can lead to jobs triggered even if the Change matching the topic has the status MERGED or ABANDONED.

In general you can't submit a whole topic in Gerrit if there is one change which has been abandoned. Since topics are not unique in Gerrit and it always can happen that some developers assign a change to another topic as reference there should be an option to prevent triggering builds if the change of interest assigned to the topic is already in a specific state.

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
@@ -27,12 +27,12 @@ package com.sonyericsson.hudson.plugins.gerrit.trigger.hudsontrigger;
 import com.sonyericsson.hudson.plugins.gerrit.trigger.GerritServer;
 import com.sonyericsson.hudson.plugins.gerrit.trigger.PluginImpl;
 import com.sonyericsson.hudson.plugins.gerrit.trigger.VerdictCategory;
-import com.sonyericsson.hudson.plugins.gerrit.trigger.config.Config;
 import com.sonyericsson.hudson.plugins.gerrit.trigger.config.IGerritHudsonTriggerConfig;
 import com.sonyericsson.hudson.plugins.gerrit.trigger.config.ReplicationConfig;
 import com.sonyericsson.hudson.plugins.gerrit.trigger.dependency.DependencyQueueTaskDispatcher;
 import com.sonyericsson.hudson.plugins.gerrit.trigger.gerritnotifier.ToGerritRunListener;
 import com.sonyericsson.hudson.plugins.gerrit.trigger.hudsontrigger.actions.GerritTriggerInformationAction;
+import com.sonyericsson.hudson.plugins.gerrit.trigger.hudsontrigger.data.TopicAssociation;
 import com.sonyericsson.hudson.plugins.gerrit.trigger.hudsontrigger.data.BuildCancellationPolicy;
 import com.sonyericsson.hudson.plugins.gerrit.trigger.hudsontrigger.data.GerritProject;
 import com.sonyericsson.hudson.plugins.gerrit.trigger.hudsontrigger.data.GerritSlave;
@@ -142,7 +142,7 @@ public class GerritTrigger extends Trigger<Job> {
     private Integer gerritBuildAbortedVerifiedValue;
     private Integer gerritBuildAbortedCodeReviewValue;
     private boolean silentMode;
-    private boolean enableTopicAssociation = Config.DEFAULT_ENABLE_TOPIC_ASSOCIATION;
+    private TopicAssociation topicAssociation;
     private String notificationLevel;
     private boolean silentStartMode;
     private boolean escapeQuotes;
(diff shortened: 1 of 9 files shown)


ckreisl:
This might be related to PR #432 @chenhuahuan.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -133,7 +133,7 @@ public class GerritTrigger extends Trigger<Job> {
     private Integer gerritBuildAbortedVerifiedValue;
     private Integer gerritBuildAbortedCodeReviewValue;
     private boolean silentMode;
-    private boolean enableTopicAssociation = Config.DEFAULT_ENABLE_TOPIC_ASSOCIATION;
```
ckreisl, line 136: Question from my side, can such a change break the job options if the update is applied?

ckreisl:
Would close and tackle exactly this this kind of issue [JENKINS-65462](https://issues.jenkins.io/browse/JENKINS-65462).

ckreisl:
@rsandell any chance to provide a review? :)

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
rsandell: Yes it will fail to be read from disk. You need to mark the old field as `transient` and then in `readResolve` if the field is set migrate it to the new field.
See https://www.jenkins.io/doc/developer/persistence/backward-compatibility/

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -941,24 +941,39 @@ private boolean isChangeInteresting(Change change, GerritProject project, Gerrit
...
+
+            if (change.equals(event.getChange())) {
+                continue;
+            }
+
+            if (!isChangeInteresting(change, project, server.getQueryHandler())) {
+                continue;
+            }
+
+            if (!topicAssociation.isInterestingChangeStatus(change)) {
```
rsandell, line 970: `topicAssociation` could be `null`

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -941,24 +941,39 @@ private boolean isChangeInteresting(Change change, GerritProject project, Gerrit
...
+
+        GerritServer server = PluginImpl.getServer_(event);
+
+        if (server == null) {
+            return false;
+        }
+
+        Map<Change, PatchSet> changes = topic.getChanges(server.getQueryHandler());
+        for (Change change : changes.keySet()) {
+
```
rsandell, line 961: keeping some `trace` or `debug` logging here would be nice I think if it is not behaving as expected for someone.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -1837,23 +1864,67 @@ public void setSilentMode(boolean silentMode) {
...
+        this.topicAssociation.setEnabled(enable);
+    }
+
     /**
      * Check if topic association is enabled.
+     * Used for jelly file.
+     *
      * @return true if so.
      */
-    public boolean isEnableTopicAssociation() {
```
rsandell, line 1844: old method signatures should be kept as much as possible for backwards binary compatibility.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -1837,23 +1864,67 @@ public void setSilentMode(boolean silentMode) {
...
-     * When topic association is on the job can be triggered by project not in its configuration
-     * as long as there are projects of interest in the same topic.
-     * @param enableTopicAssociation true if it should be enabled.
+     * Returns true if a change in state NEW should be ignored otherwise false.
+     * Used for jelly file.
+     *
+     * @return true if it should be ignored otherwise false
      */
-    @DataBoundSetter
-    public void setEnableTopicAssociation(boolean enableTopicAssociation) {
```
rsandell, line 1855: same here, just mark them as `@Deprecated` if it is not intended to be used any more. `@DataBoundSetter` might be removed, i'm not sure if that would break any JCasC scenarios or not.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -1837,23 +1864,67 @@ public void setSilentMode(boolean silentMode) {
...
+        return topicAssociation != null && topicAssociation.isIgnoreNewChangeStatus();
+    }
+
+    /**
+     * Returns true if a change in state NEW should be ignored otherwise false.
+     * Used for jelly file.
+     *
+     * @return true if it should be ignored otherwise false
+     */
+    public boolean isIgnoreMergedChangeStatus() {
```
rsandell, line 1916: I don't see these new methods being used anywhere in particular? Seems a bit excessive to introduce them when a simple `@ChackForNull getTopicAssisiation` could be added? I am trying to keep the size of this class from not growing  too much bigger than it already is :)

src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger/config.jelly
```diff
@@ -236,12 +236,32 @@
...
                        checked="${it.silentMode}"/>
         </f:entry>
-        <f:entry title="${%Enable Topic Association}"
-                 field="enableTopicAssociation"
-                 help="/plugin/gerrit-trigger/trigger/help-EnableTopicAssociation.html">
-           <f:checkbox name="enableTopicAssociation"
-                       checked="${it.enableTopicAssociation}"/>
-        </f:entry>
+
+        <f:optionalBlock name="topicAssociation"
```
rsandell, line 240: I think this could be made a bit more sleek with a `config.jelly` view for the `TopicAssociation` class and a `<f:optionalProperty/>` here instead.

rsandell (changes requested):

rsandell:
Great!
I pushed my suggested changes and if the build is green I'll merge it :)

rsandell (approved):

ckreisl:
Great! Before creating a new release i would check on the issue mentioned here: #478 .
Related to: https://issues.jenkins.io/browse/JENKINS-70162

PR #484, opened 2023-02-03
[JENKINS-70558] Fix TopicAssociation not set after Jenkins restart

ckreisl:

The PR #481 introduced a small bug that the `TopicAssociation` field is not set after a Jenkins restart.

Description from Jenkins issue: https://issues.jenkins.io/browse/JENKINS-70558 

In the job configuration if TopicAssociation is enabled and Jenkins is restarted the option is not set anymore.
The reason is that the `readResolve()` function checks the config value 
`private transient boolean enableTopicAssociation = Config.DEFAULT_ENABLE_TOPIC_ASSOCIATION;`
 (GerritTrigger.java) which is always **false**. Since this condition is always the case the topicAssociation value in the GerritTrigger class is set to "null".

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
@@ -2176,11 +2176,8 @@ public class GerritTrigger extends Trigger<Job> {
         if (projectListIsReady == null) {
             projectListIsReady = new CountDownLatch(0);
         }
-
-        if (enableTopicAssociation) {
+        if (topicAssociation == null) {
             topicAssociation = new TopicAssociation();
-        } else {
-            topicAssociation = null;
         }
 
         return super.readResolve();
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
@@ -1898,7 +1898,6 @@ public class GerritTrigger extends Trigger<Job> {
      * @param enable true or false.
      */
     @Deprecated
-    @DataBoundSetter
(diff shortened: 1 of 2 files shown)


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -1893,10 +1893,12 @@ public void setSilentMode(boolean silentMode) {
 
     /**
      * Enable or disable Topic Association option.
+     * Replaced by {@link #setTopicAssociation(TopicAssociation)} ()
      *
      * @param enable true or false.
      */
     @Deprecated
+    @DataBoundSetter
```
ckreisl, line 1901: I think that's the way how we want it. If the old build configuration xml is read it will set already create the new TopicAssociation class.

Ich think line 2181 becomes then obsolete?

rsandell: `config.xml` deserialization doesn't care about `@DataBoundSetter` et.al. Those are only relevant for form submission and CasC.

rsandell (changes requested):
You can write a test using [JenkinsSessionRule](https://javadoc.jenkins.io/component/jenkins-test-harness/org/jvnet/hudson/test/JenkinsSessionRule.html) or [RealJenkinsRule](https://javadoc.jenkins.io/component/jenkins-test-harness/org/jvnet/hudson/test/RealJenkinsRule.html) to test that the configuration actually survives a restart.

ckreisl:
> You can write a test using [JenkinsSessionRule](https://javadoc.jenkins.io/component/jenkins-test-harness/org/jvnet/hudson/test/JenkinsSessionRule.html) or [RealJenkinsRule](https://javadoc.jenkins.io/component/jenkins-test-harness/org/jvnet/hudson/test/RealJenkinsRule.html) to test that the configuration actually survives a restart.

Thanks for the pointer, added a test now and removed the `@DataBoundSetter`.

rsandell (approved):

PR #485, opened 2023-02-08
[JENKINS-70494] Fix retrigger button visibility

ckreisl:

Rel to: https://issues.jenkins.io/browse/JENKINS-70494 

With the latest security patch in version 2.38.1 the behavior of the visibility of the retrigger buttons has changed due to the newly added action.jelly files.

Security patch change: https://github.com/jenkinsci/gerrit-trigger-plugin/commit/691d76fdf54d659f8585ea3cbc3cce60d9edfec8

With the introduced pre-check in this jelly file whether the button should be visible or not we could get rid of some obsolete code.

The only "issue" i faced now during debugging is that the  `response.sendRedirect2(entity.getProject().getAbsoluteUrl());` did not redirect me to the build job page. There was just a small popup by clicking the "Retrigger" button stating "Done.".
I've justed tested it on Win 11 - Google Chrome.

Any feedback from your end @rsandell ?

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/actions/RetriggerAllAction.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/actions/RetriggerAllAction.java
@@ -116,7 +116,7 @@ public class RetriggerAllAction implements Action {
      * checks if the current user has permission to build/retrigger this and the other projects.
      * @return true if so.
      */
-    public boolean hasPermission() {
+    private boolean hasPermission() {
         if (context == null || context.getThisBuild() == null || context.getThisBuild().getProject() == null) {
             return false;
         }
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/actions/RetriggerAllAction.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/actions/RetriggerAllAction.java
@@ -86,7 +86,7 @@ public class RetriggerAllAction implements Action {
      * It does a null check on the context before calling.
      * @return true if there are any other builds in the context.
      */
-    public boolean hasOthers() {
+    private boolean hasOthers() {
         if (context != null) {
             return context.hasOthers();
         } else {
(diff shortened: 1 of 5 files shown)


rsandell (changes requested):
I don't understand why you have removed all security checks.

ckreisl:
Can you guide me here or provide a Jenkins docs about security at all?
My understanding was, at least for the `getIconFileName()`, `getDisplayName()` and `getUrlName()`we could get rid of the checks since it moved to an upper level in `action.groovy` with `<j:if test="${action.isVisible()}">`. Since this groovy combined with the function is now responsible for the button visibility.

For the `doIndex()` POST method it's true from my understanding it should still be there.
Another question is how `@Restricted(NoExternalUse.class)` works in the Jenkins context.

ckreisl:
Just checked: https://www.jenkins.io/doc/developer/security/ now and based on that it seems that as mentioned the checks in the mentioned functions seems not requried except the POST method `doIndex()`.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/actions/RetriggerAllAction.java
```diff
@@ -68,37 +68,24 @@ public RetriggerAllAction() {
...
+        return "gerrit-trigger-retrigger-all";
     }
 
     /**
      * Checks {@link TriggerContext#hasOthers()} if the context exists.
      * It does a null check on the context before calling.
      * @return true if there are any other builds in the context.
      */
-    @Restricted(NoExternalUse.class)
     public boolean hasOthers() {
```
rsandell, line 89: Why give permission for other plugins to call this method? The restriction is there so that the method can be removed if it is no longer needed. Was that your intention instead? Or perhaps you intended to make it `private`? All that is possible since it has been restricted for use in other plugins since it's inception :)

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/actions/RetriggerAllAction.java
```diff
@@ -129,7 +116,6 @@ private boolean isBuilding() {
      * checks if the current user has permission to build/retrigger this and the other projects.
      * @return true if so.
      */
-    @Restricted(NoExternalUse.class)
```
rsandell, line 132: Why give permission for other plugins to call this method?

rsandell (changes requested):
OK the motivation for permission removal seems to be sound. Found some other comments to address though.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/actions/RetriggerAllAction.java
ckreisl: ok, if this is the use of `@Restricted(NoExternalUse.class)` then yes, just marked the functions now as `private`.

rsandell (approved):

PR #490, opened 2023-07-10
[JENKINS-71616] Use full project name instead of related name for comparing job names

Mkoz:
https://issues.jenkins.io/browse/JENKINS-71616

If project has complicated name like folder/project-name then buildCurrentPatchesOnly feature does not work because of cancelMatchingJobs gets parametr jobName as full name but for comparing it uses just relative name without folder or parent project item.task.getName()).

Cannot get fullName form the task. Workaround is transform fullDisplayName with replacing separator from " » "to "/"
item.task.getFullDisplayName().replace(" » ", "/")

Don't have any idea how to prepare tests, looks like test implementation requires much more time then fix.

### Testing done

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/RunningJobs.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/RunningJobs.java
@@ -28,6 +28,7 @@ import jenkins.model.CauseOfInterruption;
 import jenkins.model.Jenkins;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import java.util.regex.Pattern;
 
 /**
 * Class for maintaining and synchronizing the runningJobs info.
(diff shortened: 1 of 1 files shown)


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/RunningJobs.java
```diff
@@ -233,8 +233,12 @@ private void cancelMatchingJobs(GerritTriggeredEvent event, String jobName, Caus
            List<Queue.Item> itemsInQueue = Queue.getInstance().getItems((Queue.Task)getJob());
            for (Queue.Item item : itemsInQueue) {
                if (checkCausedByGerrit(event, item.getCauses())) {
-                   if (jobName.equals(item.task.getName())) {
+                   // Use transformed fullDisplayNameFor to avoid
+                   // ambigous comparing between same name projects but located at different folders
+                   // FolderName » jobName to FolderName/jobName
+                   if (jobName.equals(item.task.getFullDisplayName().replace(" » ", "/"))) {
```
rsandell, line 239: The problem here is that the `displayName` can be anything and change almost at any time and doesn't need to be anything like `jobName` that is gained from `getName`/`getFullName`

Mkoz: Another option is change the strict comparison between jobNames to regexp something like ".*jobName$"
But in this case we will have collision with projects located at the different projects folders with same names. This is quite less possible scenario, then I'm going to fix. 
@rsandell what do you think is it ok to fix it this way?

rsandell (approved):
Looks like the lesser evil for this issue.
Could you provide some tests for it as well please?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/RunningJobs.java
```diff
@@ -233,8 +234,16 @@ private void cancelMatchingJobs(GerritTriggeredEvent event, String jobName, Caus
            List<Queue.Item> itemsInQueue = Queue.getInstance().getItems((Queue.Task)getJob());
            for (Queue.Item item : itemsInQueue) {
                if (checkCausedByGerrit(event, item.getCauses())) {
-                   if (jobName.equals(item.task.getName())) {
+                   // Use regexpt .*jobName to avoid
```
mawinter69, line 237: Item task is always an instance of  `hudson.model.Item` I think, so you could check if this is the case and if yes cast to `Item` and then use `getFullName()`  for the comparison. No need use a regex here.

panicking:
Found the same and I have a fix for this in my pull request too

PR #493, opened 2023-07-16
[JENKINS-66535] Fix situation when failed builds are missing and only successful builds remain

Pijukatel:
Solution proposal for situation when failed builds are missing and only successful builds remain. In such situations currently verified score will be calculated from remaining successful builds.
With the fix the verified score is saturated if any builds are failing even if those builds are no longer available to use them in score calculation.

https://issues.jenkins.io/browse/JENKINS-66535

### Testing done
Added two tests to cover the scenario described in the ticket.

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ParameterExpander.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ParameterExpander.java
@@ -388,7 +388,8 @@ public class ParameterExpander {
      * @return the lowest verified value.
      */
     @CheckForNull
-    public Integer getMinimumVerifiedValue(MemoryImprint memoryImprint, boolean onlyBuilt, Integer maxAllowedVerifiedValue) {
+    public Integer getMinimumVerifiedValue(MemoryImprint memoryImprint, boolean onlyBuilt,
+                                           Integer maxAllowedVerifiedValue) {
         Integer verified = Integer.MAX_VALUE;
         for (Entry entry : memoryImprint.getEntries()) {
             if (entry == null) {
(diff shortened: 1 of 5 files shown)


GuillemSeCa (approved):

rsandell (approved):
Looks ok, just need to fix the build error.

Pijukatel:
I will fix the formatting and add few more tests to cover all new lines.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ParameterExpander.java
```diff
@@ -545,13 +549,17 @@ public String getBuildCompletedCommand(MemoryImprint memoryImprint, TaskListener
         } else {
             //Just as bad as failed for now.
             command = config.getGerritCmdBuildFailed();
+            // Some builds could have failed, but are already deleted and not
+            // available for score calculation.
+            // Set pessimistic upper boundary on verified value.
+            maxAllowedVerifiedValue = config.getGerritBuildFailedVerifiedValue();
```
Pijukatel, line 555: The only relevant branch for setting this. In other branches of this if-else statement this code would be redundant as correct verified values will be set as consequence of entry conditions permitting execution of those branches.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ParameterExpander.java
```diff
@@ -381,12 +381,15 @@ protected Integer getVerifiedValue(Result res, GerritTrigger trigger) {
...
-     * @param onlyBuilt only count builds that completed (no NOT_BUILT builds)
+     *
+     * @param memoryImprint    the memory.
+     * @param onlyBuilt        only count builds that completed (no NOT_BUILT builds)
+     * @param maxAllowedVerifiedValue Upper boundary on verified value.
      * @return the lowest verified value.
      */
     @CheckForNull
-    public Integer getMinimumVerifiedValue(MemoryImprint memoryImprint, boolean onlyBuilt) {
+    public Integer getMinimumVerifiedValue(MemoryImprint memoryImprint, boolean onlyBuilt,
```
rsandell, line 391: Normally a public method should not be changed for binary compatibility concerns, the old signature should be left intact but `@Deprecated` that just calls the new method with default values.
But I doubt that there are any usages of this class in other plugins, so I'll let it slide for now.

rsandell (approved):

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/ParameterExpander.java
Pijukatel: Now when I think of it, maybe I should have written overloaded function instead and do the saturation there:

`
getMinimumVerifiedValue(MemoryImprint memoryImprint, boolean onlyBuilt,Integer maxAllowedVerifiedValue){
    return Math.min(maxAllowedVerifiedValue, getMinimumVerifiedValue(memoryImprint, onlyBuilt));
}
`

PR #496, opened 2023-08-18
[JENKINS-71328] Add handling for events with Null ChangeSet attributes

poorlyDefined:
Fix for https://issues.jenkins.io/browse/JENKINS-71328

Gerrit documentation for Topic events clearly shows that these events can be sent without ChangeSet information.
See https://gerrit-review.googlesource.com/Documentation/cmd-stream-events.html#_topic_changed

From the same document we see the following comment:

> SCHEMA
> The JSON messages consist of nested objects referencing the change, patchSet, account involved, and other attributes as appropriate.
> 
> Note that any field may be missing in the JSON messages, so consumers of this JSON stream should deal with that appropriately.

In case of JENKINS-71328, the consumer (RunningJobs.shouldIgnoreEvent(..) is processing events without any awareness that attributes being processed are null, resulting in NullPointer exceptions. This behavior in turn causes events to not be properly processed.

The fix modifies the code to check if the attribute exists before trying to access it.

### Testing done

A reproduction of the failing state can be followed by these steps:
Repro Steps
0. ensure gerrit trigger for the Jenkins job configured to trigger a build when the change is rebased.
1.  create a change with a topic
2. push the change
3. edit the change and remove the topic
4. rebase the change

Expected:
The gerrit plugin triggers a Jenkins build.

Actual:
mvn hpi:run with a breakpoint at line 194 of src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/RunningJobs.java
should show the resulting NPE.

With the patch, the Expected result is obtained.

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/RunningJobs.java
+++ b/src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/RunningJobs.java
@@ -191,9 +191,18 @@ public class RunningJobs {
            return true;
        }
 
-       boolean shouldCancelPatchsetNumber = policy.isAbortNewPatchsets()
-               || Integer.parseInt(runningChangeBasedEvent.getPatchSet().getNumber())
+       // events of "type": "topic-changed" are not required to set a PatchSet
+       boolean hasPatchSets = runningChangeBasedEvent.getPatchSet() != null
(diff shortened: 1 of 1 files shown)


rsandell (approved):
A unit test would be nice so that future changes doesn't miss this scenario again.

PR #500, opened 2023-10-31
Added GERRIT_EVENT_UPDATED_APPROVALS parameter

eternity74:
Added GERRIT_EVENT_UPDATED_APPROVALS parameter to know gerrit label value changes from jenkins script.
GERRIT_EVENT_UPDATED_APPROVALS will contains json string with label name as a key and object value which contians label value and old_value.
By this change, we can add trigger of multiple verdict values on the same verdict and make different job depending on the value.

We can do same thing by call ssh to gerrit to get the label value but gerring the value from the parameter is much easier.


rsandell:
Sorry on the delayed review. I've made some changes to your code to make the generated json less brittle and I also added a test. I was unable to push to your branch to show my suggested changes, so I merged them directly to master.
If you don't agree with my changes please file a new PR with your suggestions.

eternity74:
@rsandell Thanks for review and accepting my pull request.
I checked your suggested changes and it looks much simpler and better!
I could learn a bit from your suggestion.

PR #501, opened 2023-12-29
TriggerMonitor: Fix grey icon anime bug and modernize icon animation

panicking:
Use the new icon from jenkins build class and drop legacy icon

Falling this conversation https://github.com/jenkinsci/jenkins/pull/8806

The grey animation and grey icon are dropped from jenkins core. We need to move to use the new way

### Testing done

Deploy plugin in my jenkins instance

Screenshot from 2023-12-29 09-49-32

### DIFF
--- a/src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritCause/description.jelly
+++ b/src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritCause/description.jelly
@@ -44,7 +44,7 @@
                     <j:choose>
                         <j:when test="${other.hasBuild()}">
                             <td width="18">
-                                <img src="${imagesURL}/16x16/${other.build.iconColor.image}"/>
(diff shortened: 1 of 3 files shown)


rsandell (approved):

src/main/resources/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritCause/description.jelly
```diff
@@ -44,7 +44,7 @@
                     <j:choose>
                         <j:when test="${other.hasBuild()}">
                             <td width="18">
-                                <img src="${imagesURL}/16x16/${other.build.iconColor.image}"/>
+                                <l:icon class="${build.build.iconColor.iconClassName}"/>
```
mawinter69, line 47: Shouldn't that be `other.build.iconColor.iconClassName`?
Also in line 57 the change to `<l:icon class="icon-grey-anime icon-sm"/>` is missing similar to the change in ajaxTriggerMonitor.jelly

PR #504, opened 2024-02-18
Should fix  [JENKINS-70162]   [JENKINS-72193]   [JENKINS-70539]   [JENKINS-69338] Fix topic cancelling patches (this close races in abortBuild)

panicking:
Pull request fix bug connected of abort jobs coming from the same topic. What we try to get is the same behavior
of 2.36.1.

If we have a topic A->B->C we would like that the whole topic get built so build it from C. The problem was
that the jenkins try to build A , B and C for each new patcheset. There were mainly to problem:
- one was the name matching
-  other was block syncronization. I did not find for now a better way

### Testing done

Create a project in gerrit, allow the project to be triggered and confirm that topic was really aborted


panicking:
@rsandell For me is building but now it fails on CI, did not find any reason

rsandell:
Looks like the build was taking too long and got aborted.

panicking:
@rsandell Do you have time to review patches? Like the UI change and this one? I have already deployed both in two instance

rsandell (approved):

PR #516, opened 2025-04-24
Require Jenkins 2.479.3 or newer and Jakarta EE 9

strangelookingnerd:
## Require Jenkins 2.479.3 or newer and Jakarta EE 9

Jenkins 2.479.3 provides Jakarta EE 9, Eclipse Jetty 12, Spring Security 6, and Java 17.

* Update plugin pom and baseline
* Remove deprecations
* Use Java 17 language features
* Minor cleanup

### Testing done

`mvn clean verify`

### Submitter checklist
* [x]  Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
* [x]  Ensure that the pull request title represents the desired changelog entry
* [x]  Please describe what you did
* [x]  Link to relevant issues in GitHub or Jira
* [x]  Link to relevant pull requests, esp. upstream and downstream changes
* [x]  Ensure you have provided tests - that demonstrates feature works or fixes the issue


rsandell (approved):

rsandell (approved):

panicking:
@strangelookingnerd  I have rebased here and add one patch https://github.com/jenkinsci/gerrit-trigger-plugin/pull/518

rsandell (approved):

rsandell:
> Cannot invoke "hudson.ExtensionList.isEmpty()" because "strategies" is null

I've seen this error someplace before, has to do with a bad mock I think.

rsandell:
The tests that are timing out seems to all have the same symptoms, perhaps something has changed that is making the ssh mock server not working?

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/GerritProjectListUpdaterFunctionalTest.java
```diff
@@ -99,7 +99,7 @@ public void setUp() throws Exception {
      */
     @After
     public void tearDown() throws Exception {
-        sshd.stop(true);
+        // sshd.stop(true);
         sshd = null;
```
panicking, line 103: Please explain why is needed and instead comment drop it

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/LockedDownGerritEventTest.java
```diff
@@ -112,7 +112,7 @@ public void setUp() throws Exception {
      */
     @After
     public void tearDown() throws Exception {
-        sshd.stop(true);
+        // sshd.stop(true);
         sshd = null;
```
panicking, line 116: Same here

panicking (dismissed):

rsandell:
bc download error fixed, test timeouts still happening though.

strangelookingnerd:
> bc download error fixed, test timeouts still happening though.

I'll try and look into those as well.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/model/BuildMemory.java
```diff
@@ -532,7 +532,7 @@ public GerritTriggeredEvent getEvent() {
          * @return the memory entries.
          */
         public synchronized Entry[] getEntries() {
-            return list.toArray(new Entry[list.size()]);
+            return list.toArray(new Entry[0]);
```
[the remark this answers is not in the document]
strangelookingnerd: > There are two styles to convert a collection to an array:
> 
>     * A pre-sized array, for example, c.toArray(new String[c.size()])
> 
>     * An empty array, for example, c.toArray(new String[0])
> 
> In older Java versions, using a pre-sized array was recommended, as the reflection call necessary to create an array of proper size was quite slow.
> 
> However, since late updates of OpenJDK 6, this call was intrinsified, making the performance of the empty array version the same, and sometimes even better, compared to the pre-sized version. Also, passing a pre-sized array is dangerous for a concurrent or synchronized collection as a data race is possible between the size and toArray calls. This may result in extra nulls at the end of the array if the collection was concurrently shrunk during the operation.

https://www.jetbrains.com/help/inspectopedia/ToArrayCallWithZeroLengthArrayArgument.html

rsandell:
I've fixed all tests and checkstyle warnings, but on the commit before #da6f92d, so I'll be having some merge fun first.

rsandell:
Ugh, it's one of those "works on my machine but not on CI" test problems :exploding_head: .

rsandell (approved):

PR #1010, opened 2026-05-06
Interface creation

### LINKED ISSUE #1008: [DS] Create interfaces pattern for memory management

ironcerocloudbees:
# Add Extension Points Pattern for Storage and Notification Claiming

Fixes #1008 

## Summary

This PR implements the Extension Points pattern for two core components: **BuildMemoryStorage** and **NotificationClaimStrategy**. This architectural refactoring enables switching between local (TreeMap-based) and other distributed implementations (e.g., Hazelcast-based) without if/else conditionals.

### 1. BuildMemoryStorage abstract class (2 new files)

Created a storage abstraction layer for BuildMemory:

- **`spi/BuildMemoryStorage.java`** - Abstract class defining 20 storage operations (CRUD, queries, metadata, reporting)
- **`storage/LocalBuildMemoryStorage.java`** - TreeMap-based extending class (extracted from BuildMemory)

**Refactored:**
- **`gerritnotifier/model/BuildMemory.java`** - Now delegates all storage operations to `BuildMemoryStorage` abstract class.

### 2. NotificationClaimStrategy abstract class (2 new files)

Created a claiming abstraction layer for notification coordination:

- **`spi/NotificationClaimStrategy.java`** - Abstract class defining 2 methods (claim/release notification rights)
- **`gerritnotifier/LocalNotificationClaimStrategy.java`** - Local extending class (always returns true)

### Entering point to choose the implementation

- **`coordination/LocalCoordinationProvider.java`** - Provider with `@Extension(ordinal=-1000)`
- **`coordination/CoordinationModeFactory.java`** - Singleton (Extension) factory with ExtensionList discovery

## Architecture Pattern

Both abstract class follow the same Jenkins Extension Points pattern:

Client Code (BuildMemory / ToGerritRunListener)
        ↓
    Factory (discovers via ExtensionList.lookup())
        ↓
    Provider (@Extension with priority + isAvailable())
        ↓
    Implementation (e.g. Local)

**Selection Algorithm:**
1. Jenkins discovers all providers via `ExtensionList.lookup()`
2. Filters to available providers (`isAvailable() == true`)
3. Selects provider with highest priority
4. Creates singleton instance

**Current Priority:**
- Local implementations: `ordinal=-100` (fallback, always available)
- Future cluster implementations: `ordinal=100` (higher priority when cluster mode enabled)

### Zero if/else Proliferation
- Avoiding the usage of if/else to use different implementations.

### Zero Coupling
- BuildMemory has no knowledge of storage implementation
- No imports for TreeMap or future classes
- Business logic separated from storage concerns

### Jenkins-Native Pattern
- Uses standard @Extension mechanism
- Familiar to Jenkins plugin developers
- Jenkins automatically discovers implementations

### Zero changes required to existing code
Factory will automatically select cluster implementations when available.

### ✅ 100% backward compatible - no breaking changes:
- Same TreeMap-based storage (via LocalBuildMemoryStorage)
- Same thread safety (synchronized methods)
- Same public API
- Same behavior for all BuildMemory operations
- Deprecated methods in RunningJobs still work (delegate to BuildMemory)

### Automated Testing
- 624 tests passing (0 failures, 0 errors, 2 skipped)
- All existing unit tests pass without modification
- Integration tests verify build cancellation still works correctly
- Test isolation ensured with factory reset in tearDown methods

### Test Coverage
- BuildMemoryTest.java - 36 tests covering all storage operations
- BuildCancellationIntegrationTest.java - 6 integration tests for cancellation policy
- ParameterModeJenkinsTest.java - 16 tests with JenkinsRule
- Full regression suite confirms no behavior changes

### Manual Verification
1. ✅ Compilation succeeds (mvn compile)
2. ✅ Checkstyle passes (0 violations)
3. ✅ All 624 tests pass (mvn test)
4. ✅ Build memory operations maintain identical behavior
5. ✅ Cancellation policy works correctly (builds properly marked as ABORTED)
6. ✅ Factory correctly discovers local providers in test environment

### Code Quality
- No checkstyle violations
- Proper documentation (Javadoc on all public methods)
- Follows existing code style and conventions
- Logger statements for debugging provider selection

### Submitter checklist
- Make sure you are opening from a topic/feature/bugfix branch (right side) and not your main branch!
- Ensure that the pull request title represents the desired changelog entry
- Please describe what you did
- Link to relevant issues in GitHub or Jira
- Link to relevant pull requests, esp. upstream and downstream changes
- Ensure you have provided tests that demonstrate the feature works or the issue is fixed

### Changelog Entry
Add Extension Points pattern for BuildMemoryStorage and NotificationClaimStrategy to enable pluggable implementations (preparation for HA/HS cluster mode support)


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/model/BuildMemory.java
```diff
@@ -57,16 +59,18 @@
...
 
 import static com.sonyericsson.hudson.plugins.gerrit.trigger.PluginImpl.getServerConfig;
 import static com.sonyericsson.hudson.plugins.gerrit.trigger.utils.Logic.shouldSkip;
 
 /**
  * Keeps track of what builds have been triggered and if all builds are done for specific events.
+ * <p>
+ * This class now delegates all storage operations to a {@link BuildMemoryStorage} implementation,
+ * discovered via the Extension Points pattern. This allows switching between local (TreeMap) and
+ * distributed (Hazelcast) storage without changing this class.
```
sboardwell:
```suggestion
 * distributed (e.g. Hazelcast) storage without changing this class.
```

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/model/BuildMemory.java
```diff
@@ -461,23 +393,26 @@ public void cancelOutdatedEvents(
...
+                    // IMPORTANT: We need to get the imprint from storage again to modify the real one,
+                    // not the copy from getAllEvents()
+                    MemoryImprint storageImprint = storage.getMemoryImprint(runningEvent);
+                    if (storageImprint != null) {
+                        for (Entry imprintEntry : storageImprint.getEntries()) {
+                            if (imprintEntry.isProject(jobName)
+                                    && !imprintEntry.isBuildCompleted()
+                                    && !imprintEntry.isCancelled()) {
+                                imprintEntry.setCancelled(true);
+                                imprintEntry.setBuildCompleted(true);
```
sboardwell: Still not sure what we agreed upon here. If something happens between here and `cancelMatchingJobs` we now have an inconsistent state. The intent is important here (it is scheduled to be cancelled and should not be considered), but we are effectively lying about the state for a few milliseconds/seconds. wdyt?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/model/BuildMemory.java
```diff
@@ -707,19 +642,8 @@ public void scheduled(ChangeBasedEvent event,
...
-                if (entry.getBuild() != null) {
-                    list.add(entry.getBuild());
-                }
-            }
-            return list;
-        } else {
-            return null;
-        }
+    public List<Run> getBuilds(GerritTriggeredEvent event) {
+        return storage.getBuilds(event);
```
sboardwell, line 649: Is the `LinkedList` important in this method (as opposed to the normal List)?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/model/BuildMemory.java
```diff
@@ -991,7 +898,7 @@ public synchronized String getStatusReport() {
          * @param project the project.
          * @return the entry or null if nothing is found.
          */
-        private Entry getEntry(@NonNull Job project) {
+        public Entry getEntry(@NonNull Job project) {
```
sboardwell, line 904: Why does everything need to be public here? Not saying it is incorrect, but seems a little strange when it was either private or protected before

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/NotificationClaimStrategyFactory.java
```diff
@@ -0,0 +1,214 @@
...
+            logger.info("Selected NotificationClaimStrategy provider: {} (priority={})",
+                bestProvider.getName(), bestProvider.getPriority());
+
+            NotificationClaimStrategy strategy = bestProvider.createStrategy();
+            logger.info("Created NotificationClaimStrategy: {}", strategy.getClass().getSimpleName());
+
+            return strategy;
+
+        } catch (Exception e) {
+            logger.warn("Failed to discover NotificationClaimStrategy via ExtensionList, using fallback", e);
```
sboardwell: This is a smell IMO. If the unit tests need something, then we should give it explicitly. This could trigger a fallback, hiding a real exception.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/LocalNotificationClaimStrategy.java
```diff
@@ -0,0 +1,52 @@
+/*
+ *  The MIT License
+ *
+ *  Copyright (c) 2010, 2014 Sony Mobile Communications Inc. All rights reserved.
```
sboardwell:
```suggestion
 *  Copyright (c) 2026 CloudBees. All rights reserved.
```

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spi/BuildMemoryStorageProvider.java
```diff
@@ -0,0 +1,126 @@
+/*
+ *  The MIT License
+ *
+ *  Copyright 2010 Sony Ericsson Mobile Communications. All rights reserved.
+ *  Copyright 2012 Sony Mobile Communications AB. All rights reserved.
```
sboardwell: Copyright

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spi/NotificationClaimStrategy.java
```diff
@@ -0,0 +1,65 @@
+/*
+ *  The MIT License
+ *
+ *  Copyright (c) 2010, 2014 Sony Mobile Communications Inc. All rights reserved.
```
sboardwell: Copyright

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spi/GerritTriggerModeProvider.java
```diff
@@ -0,0 +1,141 @@
+/*
+ *  The MIT License
+ *
+ *  Copyright (c) 2010, 2014 Sony Mobile Communications Inc. All rights reserved.
```
sboardwell: Copyright

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/storage/BuildMemoryStorageFactory.java
```diff
@@ -0,0 +1,177 @@
+/*
+ *  The MIT License
+ *
+ *  Copyright 2010 Sony Ericsson Mobile Communications. All rights reserved.
+ *  Copyright 2012 Sony Mobile Communications AB. All rights reserved.
```
sboardwell: Copyright

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/storage/LocalBuildMemoryStorage.java
```diff
@@ -0,0 +1,310 @@
+/*
+ *  The MIT License
+ *
+ *  Copyright 2010 Sony Ericsson Mobile Communications. All rights reserved.
+ *  Copyright 2012 Sony Mobile Communications AB. All rights reserved.
```
sboardwell: Copyright

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/storage/LocalBuildMemoryStorage.java
```diff
@@ -0,0 +1,310 @@
...
+        if (pb == null) {
+            //Shouldn't happen but just in case, keep the memory.
+            pb = new MemoryImprint(event);
+            memory.put(event, pb);
+        }
+        pb.set(project);
+        Entry entry = pb.getEntry(project);
+        entry.setCancelled(true);
+        entry.setBuildCompleted(true);
+    }
```
sboardwell, line 153: Deferring to Bobby here, but I can't shake the feeling that this type of stuff (in other methods as well) is not abstracting enough of the logic. We will now have to create the same/similar logic in the other implementation.

If someone changes the logic, it will need to be changed in both implementations.

I am wondering if it could be abstracted further so that the only difference is the storage or notification provider instantiated.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/storage/LocalBuildMemoryStorageProvider.java
```diff
@@ -0,0 +1,70 @@
...
+ * <p>
+ * This is the default storage implementation, always available and used
+ * in standalone Jenkins instances (non-cluster mode).
+ * <p>
+ * Priority is set to 0 (lowest), meaning distributed implementations
+ * (e.g., Hazelcast with priority 100) will take precedence when available.
+ *
+ */
+@Extension(ordinal = 0)
+public class LocalBuildMemoryStorageProvider extends BuildMemoryStorageProvider {
```
sboardwell: There are two extensions, one of storage and one for the notifications, but they always be coupled to an "implementation" and will never be different, say local notification impl and cluster storage impl.

I would probably go for a single extension which provides both notificationImpl and storageImpl

sboardwell (changes requested):
Added comments for discussion. Waiting for input from @rsandell of some of them.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/model/BuildMemory.java
ironcerocloudbees: I see your point. That block of code is necessary to address the race condition I mentioned yesterday (or two days ago). But we can "easily" change the location.

After reviewing the entire method, I believe we could move the block into the `cancelMatchingJobs` method. The `cancelMatchingJobs` method is called only inside this `cancelOutdatedEvents` method. And it could make sense to mark the job in our memory as canceled at the same time (immediately before) we throw the Jenkins action of aborting. 
So, I think we can move the code into the `cancelMatchingJobs` as:

```
private void cancelMatchingJobs(
            GerritTriggeredEvent event,
            String jobName,
            Job job,
            CauseOfInterruption cause) {

        try {
            synchronized (this){
                MemoryImprint storageImprint = storage.getMemoryImprint(event);
                if (storageImprint != null) {
                    for (Entry imprintEntry : storageImprint.getEntries()) {
                        if (imprintEntry.isProject(jobName)
                                && !imprintEntry.isBuildCompleted()
                                && !imprintEntry.isCancelled()) {
                            imprintEntry.setCancelled(true);
                            imprintEntry.setBuildCompleted(true);
                        }
                    }
                }
            }            
            
            if (!(job instanceof Queue.Task)) {
                logger.error("Error canceling job. The job is not of type Task. Job name: {}", job.getName());
                return;
            }
            // Remove any jobs in the build queue
            List<Queue.Item> itemsInQueue = Queue.getInstance().getItems((Queue.Task)job);
            for (Queue.Item item : itemsInQueue) {
                if (checkCausedByGerrit(event, item.getCauses())) {
                    Job tJob = (Job)item.task;
                    if (jobName.equals(tJob.getFullName())) {
                        Queue.getInstance().cancel(item);
                    }
                }
            }
            ...
```

We could consider moving it between the if and the list (in the "// Remove any jobs in the build queue" message). But I don't know in what kind of cases we can have an element in memory (with a build) that is referring to a job that is not a task. Maybe @rsandell knows.

In general terms I think this is a good idea. Just to decide if at the beginning of the `cancelMatchingJobs` method or after the `if (!(job instanceof Queue.Task))`.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/model/BuildMemory.java
ironcerocloudbees: The original code was using `LinkedList` in the implementation and `List` in the return. I just moved the code from the original location inside the Local implementation. I know that the kind of implementation (if LinkedList, ArrayList, etc) is important in terms of performance (some implementations have inner synchronization that could cause a bottleneck or deadlock directly), which is why I did not change the implementation (and leave the LinkedList).

Maybe @rsandell can help to clarify it.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/model/BuildMemory.java
ironcerocloudbees: In some cases, like the present one, the method was being used internally by the own class. However, in the current approach, the local implementation needs to use the same method (since the previous logic was moved inside the local implementation). Since the local implementation is in a different package (and obviously in a different class), we needed to change the scope to public to allow access from other packages.

For example:
The `getEntry` method is called by the `cancelled`/`setEntryCustomUrl`/`setEntryUnsuccessfulMessage` methods that are now in the `LocalBuildMemoryStorage` class.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/NotificationClaimStrategyFactory.java
ironcerocloudbees: Why is it hiding the real exception? We can discuss it online because I do not have a clear idea what you are talking about.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spi/GerritTriggerModeProvider.java
ironcerocloudbees: Done.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/storage/BuildMemoryStorageFactory.java
ironcerocloudbees: Done

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/storage/LocalBuildMemoryStorage.java
ironcerocloudbees: Done

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/storage/LocalBuildMemoryStorage.java
ironcerocloudbees: I also leave it to @rsandell to decide here. 

IMO, there is no clear or direct approach to make this kind of method more abstracted. For example, the purpose of this method is to mark the builds as canceled and completed in memory. We need to recover the element (the entry) from memory and modify it. This is what the method is doing right now (essentially). 

Regarding adding the event to memory if not found, to be honest, I don't know why the logic is doing it. It appears to be a fix for a corner case found in the past. But I cannot confirm it.
We could have a common logic (the abstracted part) with the `setCancelled` and `setBuildCompleted` calls. But I think it is not enough to split the code. Another thing would be if in the future we plan to increase this cancellation logic

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/storage/LocalBuildMemoryStorageProvider.java
ironcerocloudbees: I think it is a good idea. I implemented the change to unify the extensions.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/NotificationClaimStrategyFactory.java
ironcerocloudbees: Since the idea here is to stop or interrupt the main functionality of the plugin if something happens here (at the level of the extension point), @rsandell's thoughts are welcome.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/model/BuildMemory.java
```diff
@@ -417,15 +348,17 @@ public void cancelOutdatedEvents(
...
+
+        Map<GerritTriggeredEvent, MemoryImprint> allEvents = storage.getAllEvents();
+        logger.info("BuildMemory has {} events in memory", allEvents.size());
 
         String jobName = job.getFullName();
         List<ChangeBasedEvent> outdatedEvents = new ArrayList<>();
         CauseOfInterruption cause = new NewPatchSetInterruption();
 
         synchronized (this) {
             // Find all outdated events
```
rsandell, line 359: Since you have removed all the `synchronized` methods this block no longer surves any purpose and you risk a `ConcurrentModificationException` You eitrher need to synchronize on something else but probably rethink a bit how the synchronization is supposed to work here. Perhaps it needs to move into each storage implementation?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/model/BuildMemory.java
rsandell: As I explained before I don't think this is a good idea as, from what I can see, it will break the coordination and likely instead of sending one aborted message to gerrit for all the builds it will send one for each because `isAllBuildsCompleted` will be true for each one.
What is it that makes the normal events sent to the RunListener not enough here?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/model/BuildMemory.java
rsandell: Old CSci thinking; The list is only ever going to be iterated over, never indexed for a specific one, so a LinkedList should be more efficient on memory and performance. But modern CPUs with branch prediction et.al. tends to make that old type of thinking less efficient.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/GerritTriggerModeFactory.java
```diff
@@ -0,0 +1,254 @@
...
+            int bestPriority = Integer.MIN_VALUE;
+
+            for (GerritTriggerModeProvider provider : providers) {
+                logger.debug("Checking provider: {} (priority={}, available={})",
+                    provider.getModeName(),
+                    provider.getPriority(),
+                    provider.isAvailable());
+
+                if (provider.isAvailable()) {
+                    if (provider.getPriority() > bestPriority) {
```
rsandell: You shouldn't need a `getPriority()` because `ExtensionList.lookup` will give you an ordered list based on [`Extension.ordinal()`](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/Extension.java#L85) so you should be able to just pick the first in the list that `isAvailable()` just make sure that the default/fallback implementation has a low ordinal like -1000 or something.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/GerritTriggerModeFactory.java
```diff
@@ -0,0 +1,254 @@
...
+     * <p><strong>Not for production use:</strong> Calling this in production code
+     * could cause inconsistent behavior. The factory is designed to discover once
+     * at startup.</p>
+     *
+     * <p><strong>Thread Safety:</strong> This method is synchronized to prevent
+     * concurrent reset/get calls from causing race conditions.</p>
+     */
+    public static synchronized void reset() {
+        logger.debug("Resetting GerritTriggerModeFactory");
+        selectedMode = null;
```
rsandell: As I guess you've  discovered by this it is generally a bad design to have static fields holding state like this in Jenkins plugins.
It is better to make this class a singleton instance by annotating it with `@Extension` and just have a 

```
public static GerritTriggerModeFactory get() {
    return ExtensionList.lookupSingelton(GerritTriggerModelFactory.class)
}
```

that way your JenkinsRules, reload from disk etc will work better and less brittle.

Also could we think of a better name for this? TriggerMode makes it sound a bit more like the mode for how the triggering behaves and besides there are too many classes in this plugin prefixed with GerritTrigger already 😅 

Maybe CoordinationMode or MemoryMode? Naming things are hard 😄 
Also perhaps putting this and related classes in its own package?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/LocalModeProvider.java
```diff
@@ -0,0 +1,106 @@
...
+public class LocalModeProvider extends GerritTriggerModeProvider {
+
+    /**
+     * Returns the priority of this provider.
+     * Lowest priority (0) since this is the fallback mode.
+     *
+     * @return 0 (lowest priority)
+     */
+    @Override
+    public int getPriority() {
```
rsandell: Not needed when you have ordinal above.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/LocalNotificationClaimStrategy.java
rsandell: This is a completely new class with new code in, so no `Copyright (c) 2010, 2014 Sony Mobile Communications Inc. All rights reserved.`

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
```diff
@@ -2,7 +2,7 @@
  *  The MIT License
  *
  *  Copyright (c) 2010, 2014 Sony Mobile Communications Inc. All rights reserved.
- *  Copyright (c) 2014, CloudBees, Inc. All rights reserved.
+ *  Copyright (c) 2026, CloudBees, Inc. All rights reserved.
```
rsandell, line 5: Probably not nessesary but...
```suggestion
 *  Copyright (c) 2014, 2026, CloudBees, Inc. All rights reserved.
```

rsandell: oh, actually this file is not touched, so you shouldn't update the year at all.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spi/BuildMemoryStorage.java
```diff
@@ -0,0 +1,269 @@
...
+ *   <li>Local: TreeMap-based storage (standalone Jenkins)</li>
+ * </ul>
+ * <p>
+ * Implementations are discovered via Jenkins Extension Points pattern using
+ * {@link com.sonyericsson.hudson.plugins.gerrit.trigger.spi.GerritTriggerModeProvider}.
+ *
+ * @see com.sonyericsson.hudson.plugins.gerrit.trigger.spi.GerritTriggerModeProvider
+ * @see com.sonyericsson.hudson.plugins.gerrit.trigger.gerritnotifier.model.BuildMemory
+ */
+public interface BuildMemoryStorage {
```
rsandell: I guess using interfaces are ok now with "newer" java version having the posibility of default methods. But it still jerks me a bit seeing it used, the modus operandi in Jenkins is to use abstract classes when possible as they are easier to migrate in the future. But if that isn't possible then I guess an interface is fine to use.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spi/BuildMemoryStorage.java
```diff
@@ -0,0 +1,269 @@
...
+import edu.umd.cs.findbugs.annotations.CheckForNull;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Storage interface for BuildMemory operations.
+ * <p>
+ * This interface defines all storage operations needed by BuildMemory to track builds
+ * triggered by Gerrit events. Implementations provide different storage backends:
```
rsandell:
```suggestion
 * triggered by Gerrit events. Implementations provide different storage backends, for example:
```

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spi/BuildMemoryStorage.java
```diff
@@ -0,0 +1,269 @@
...
+ * <ul>
+ *   <li>Local: TreeMap-based storage (standalone Jenkins)</li>
```
rsandell:
```suggestion
 *   <li>{@link com.sonyericsson.hudson.plugins.gerrit.trigger.storage.LocalMemoryStorage}: TreeMap-based storage (standalone Jenkins)</li>
```

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spi/GerritTriggerModeProvider.java
rsandell: New file, so Sony Mobile need(/should)n't be involved.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spi/GerritTriggerModeProvider.java
```diff
@@ -0,0 +1,188 @@
...
+ * }</pre>
+ *
+ * <h2>Selection Algorithm:</h2>
+ * <ol>
+ *   <li>Jenkins discovers all providers via ExtensionList.lookup()</li>
+ *   <li>Filters to only available providers (isAvailable() == true)</li>
+ *   <li>Selects the provider with highest priority</li>
+ *   <li>Creates all component instances from selected provider</li>
+ * </ol>
+ *
```
rsandell: An `@see LocalModeProvider` should be enough imo.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spi/GerritTriggerModeProvider.java
```diff
@@ -0,0 +1,188 @@
...
+     * <p><b>Recommended values:</b></p>
+     * <ul>
+     *   <li>0 = Local/fallback mode (always available)</li>
+     *   <li>100 = Cluster mode (conditionally available)</li>
+     *   <li>Custom implementations can use any value to order themselves</li>
+     * </ul>
+     *
+     * @return the priority value (higher = higher priority)
+     */
+    public abstract int getPriority();
```
rsandell: not needed.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spi/NotificationClaimStrategy.java
rsandell: same note about being a new file

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/storage/LocalBuildMemoryStorage.java
rsandell: Yes, this contains old copied code so the old copyright should be kept intact according to the license.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/storage/LocalBuildMemoryStorage.java
```diff
@@ -0,0 +1,311 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import edu.umd.cs.findbugs.annotations.CheckForNull;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
+
```
rsandell, line 44: I am going to assume that this is just copied verbatim from the original `BuildMemory` class so that I don't waste time verifying that part :)

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/GerritTriggerModeFactory.java
```diff
@@ -0,0 +1,254 @@
+/*
+ *  The MIT License
+ *
+ *  Copyright (c) 2010, 2014 Sony Mobile Communications Inc. All rights reserved.
```
rsandell:
```suggestion
```

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/storage/LocalBuildMemoryStorage.java
rsandell: Maybe a `protected MemoryImprint getOrCreateMemoryImprint(@NonNull GerritTriggeredEvent event)` after  `getMemoryImprint` could be usefull and reused in `cancelled`, `completed`, `started` and `triggered` *(`retriggered` has some special handling and needs to know if the imprint needed to be created)* and then the hazelcast implementation would have less to implement I guess.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/model/BuildMemory.java
rsandell: or I guess you are relying on the copy that `getAllEvents` returns to avoid the concurrent modification problems?

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/model/BuildMemoryTest.java
```diff
@@ -3,6 +3,7 @@
  *
  *  Copyright 2010 Sony Ericsson Mobile Communications. All rights reserved.
  *  Copyright 2012 Sony Mobile Communications AB. All rights reserved.
+ *  Copyright 2026 CloudBees, Inc.
```
rsandell: Not enough changes to warrant a copy claim imo but IANAL

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/BuildCancellationIntegrationTest.java
```diff
@@ -1,3 +1,26 @@
+/*
+ *  The MIT License
+ *
+ *  Copyright 2026 CloudBees, Inc.
```
rsandell, line 4: Same here, but the original copyright needs to be here if you are adding the header.

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/model/BuildMemoryTest.java
```diff
@@ -89,6 +91,8 @@ public void setupFull() {
     @After
     public void tearDown() throws Exception {
         jenkinsMockedStatic.close();
+        // Reset the factory to ensure clean state for next test
+        GerritTriggerModeFactory.reset();
```
rsandell: Not needed if you go with my `@Extension` singelton suggestion.

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/BuildCancellationIntegrationTest.java
```diff
@@ -71,6 +95,8 @@ public class BuildCancellationIntegrationTest {
      */
     @Before
     public void setUp() throws Exception {
+        // Reset factory to ensure clean state for each test
+        GerritTriggerModeFactory.reset();
```
rsandell: Not needed if you go with my `@Extension` singelton suggestion.

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/BuildCancellationIntegrationTest.java
```diff
@@ -95,6 +121,8 @@ public void tearDown() throws Exception {
             sshd.stop(true);
             sshd = null;
         }
+        // Reset factory after test
+        GerritTriggerModeFactory.reset();
```
rsandell: Should not be needed if you go with my `@Extension` singelton suggestion.

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spec/gerritnotifier/SpecGerritVerifiedSetterTest.java
```diff
@@ -3,6 +3,7 @@
  *
  *  Copyright 2010 Sony Ericsson Mobile Communications. All rights reserved.
  *  Copyright 2012, 2013 Sony Mobile Communications AB. All rights reserved.
+ *  Copyright 2026 CloudBees, Inc.
```
rsandell: dito

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spec/gerritnotifier/SpecGerritVerifiedSetterTest.java
```diff
@@ -123,7 +125,8 @@ public void tearDown() throws Exception {
         messageProviderMockedStatic.close();
         jenkinsMockedStatic.close();
         triggeredBuildListenerMockedStatic.close();
-
+        // Reset the factory to ensure clean state for next test
+        GerritTriggerModeFactory.reset();
```
rsandell: dito

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spec/ParameterModeJenkinsTest.java
```diff
@@ -1,3 +1,26 @@
+/*
+ *  The MIT License
+ *
+ *  Copyright 2026 CloudBees, Inc.
```
rsandell, line 4: dito

rsandell (changes requested):

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/model/BuildMemory.java
ironcerocloudbees: Hey @rsandell. Yes, the copy in the getAllEvents prevents ConcurrentModificationException (since we are working with a copy). 

However, this case of synchronization is complex. I understand that:
The synchronized block in the BuildMemory-> cancelOutdatedEvents is about "this", which means it will block for the BuildMemory instance. The other sync in the LocalBuildMemoryStorage methods (for example, `getAllEvents`) are about the LocalBuildMemoryStorage instance. So:

The idea (or my idea) of the sync block in the `BuildMemory-> cancelOutdatedEvents` method was to serialize the cancel events. However, it is true that if other methods (such as LocalBuildMemoryStorage methods) modify the data structure, cancelOutdatedEvents can become obsolete for the current run.

Thinking of alternatives, I believe using the following block in the cancelOutdatedEvents method could resolve our problems:

```
synchronized (storage) {
```

Since we will block the code with the instance of the storage, and we will do the same in the LocalBuildMemoryStorage methods, we ensure the serialization of the cancelOutdatedEvents execution and, at the same time, we will block the BuildMemory structure while the cancellation is happening.

That said, your knowledge and experience with distributed systems (multi-threads) is drastically better than mine. So, I would greatly appreciate any tips or suggestions you can provide.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/model/BuildMemory.java
ironcerocloudbees: Are you talking about the proposed modification in my last comment (about moving the imprintEntry.setBuildCompleted(true); block into the cancelMatchingJobs method? Or about the previous method (cancelOutdatedEvents)?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/model/BuildMemory.java
ironcerocloudbees: Hey folks, do you still think we need to modify the linkedlist usage? Alternative?

sboardwell: No, I think we are good.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/NotificationClaimStrategyFactory.java
sboardwell: IMO: An exception should not occur here and thus should be thrown if encountered.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spi/BuildMemoryStorage.java
ironcerocloudbees: Hey @rsandell, where are you seeing a default method in this interface? I've re-visited, and I did not find any definition of method in this interface (only signatures). I also checked any other interfaces created in the PR and did not find any default methods. Am I misunderstanding what you meant?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/storage/LocalBuildMemoryStorage.java
ironcerocloudbees: Yes

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/LocalNotificationClaimStrategy.java
ironcerocloudbees: You are right. Done.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/storage/LocalBuildMemoryStorage.java
ironcerocloudbees: Good idea. Done.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
ironcerocloudbees: The class was updated if we diff with Master:
* abortBuild method.
* removing the runningJobs attribute.
* getRunningJobs method.
* etc

It was updated in the previous PR.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spi/BuildMemoryStorage.java
ironcerocloudbees: Done

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spi/BuildMemoryStorage.java
ironcerocloudbees: Done.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spi/GerritTriggerModeProvider.java
ironcerocloudbees: Ok. Done.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/GerritTriggerModeFactory.java
ironcerocloudbees: Done.

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/model/BuildMemoryTest.java
ironcerocloudbees: Ok. Done.

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spec/gerritnotifier/SpecGerritVerifiedSetterTest.java
ironcerocloudbees: Done.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/GerritTriggerModeFactory.java
ironcerocloudbees: Got it. I refactored the factory with CoordinationMode... one.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/GerritTriggerModeFactory.java
ironcerocloudbees: Done

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/LocalModeProvider.java
ironcerocloudbees: Done

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/model/BuildMemoryTest.java
ironcerocloudbees: Done

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/BuildCancellationIntegrationTest.java
ironcerocloudbees: Done

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/BuildCancellationIntegrationTest.java
ironcerocloudbees: Done.

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spec/gerritnotifier/SpecGerritVerifiedSetterTest.java
ironcerocloudbees: Done

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spi/GerritTriggerModeProvider.java
ironcerocloudbees: Done.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spi/BuildMemoryStorage.java
rsandell: I am talking about future extendability, when enevitably in the future a method needs to be added to the BuildMemory you can't just go and add a new definition to this interface because that would break any other plugin that is implementing it, and you would need to make sure that all plugins are implementing the new method and then coordinating a release.
Design wise in Jenkins core and plugins we solve that issue by having these things be abstract classes instead and any new method that needs to be added to the api is not added as an abstract method but an actual implementation that does something default so any existing implementations doesn't break.

in Java 8 that issue was solved for interfaces by introducing the default method definition, but many places in the Jenkins  code is keeping with tradition having abstract classes wherever possible instead of interfaces and dealing with default methods. You get some other benefits as well using actual classes instead of interfaces but the details escape me at the moment, but for example protected methods.

In short; having this be an inteface is technically ok, but I would prefer it to be an abstract class if possible :)

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/BuildCancellationIntegrationTest.java
rsandell: now nothing has changed in this class except for adding a copyright header. I normally don't have license and copyright headers on test classes, so please remove the header.

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spec/ParameterModeJenkinsTest.java
rsandell: same for this one

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/storage/LocalBuildMemoryStorage.java
```diff
@@ -0,0 +1,320 @@
...
+     * @return the MemoryImprint
+     */
+    protected MemoryImprint getOrCreateMemoryImprint(
+            @NonNull GerritTriggeredEvent event, String warningMessage) {
+        MemoryImprint pb = memory.get(event);
+        if (pb == null) {
+            //Shouldn't happen but just in case, keep the memory.
+            pb = new MemoryImprint(event);
+            if (warningMessage != null) {
+                logger.warn(warningMessage);
```
rsandell, line 148: nice

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/GerritTrigger.java
rsandell: ok

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/EventListener.java
```diff
@@ -2,6 +2,7 @@
  *  The MIT License
  *
  *  Copyright (c) 2014, CloudBees, Inc. All rights reserved.
+ *  Copyright 2026 CloudBees, Inc.
```
rsandell: So is it the same for this as GerritTrigger.java?

The lawyers told me that copyright claim should be ammended if a significant portion was added or changed. What is meant by a significant ammount is in the eye of the beholder 😁

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/model/BuildMemory.java
rsandell: I am confised by what you mean by serialize (to me that means writing the object to disk :) )

Anyway synchronizing on the storage class would be almost technically correct, you will be using the same monitor. But alot can have happened between the time when the synchronization was released by the return of `storage.getAllEvents()` and when you regain the monitor in this block. So even though there won't be any concurrent modification going on you would still be working with outdated data. And it is design wise very bad to have to know about and do external synchronization like that. And we will have to revisit again for the hazelcast implementation where I guess we will have to do some other form of distributed synchronization?

Maybe in BuildMemory having something like 
```java
public synchronized void forEachEvent(BiConsumer<GerritTriggeredEvent, MemoryImprint> c) {
    Iterator<Map.Entry<GerritTriggeredEvent, MemoryImprint>> it = memory.entrySet().iterator();
    while (it.hasNext()) { 
      Map.Entry<GerritTriggeredEvent, MemoryImprint> entry = it.next();
      c.apply(entry.getKey(), entry.getValue())
    }
}
```
and then use that here would solve this issue?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/model/BuildMemory.java
rsandell: I can't say anything about your proposal about moving into `cancelMatchingJobs` because I am still a bit thick in my brain about this block of code and why it's needed. I can't really understand the issue of what it is for. 

And I get stuck on the thing we discussed about the coordinated callback to Gerrit getting broken for multiple builds potentially with downstream builds etc.

ironcerocloudbees: Hi @rsandell,

I created this block because we are now using the BuildMemory to check the events and build for the cancellation (when we were using the RunningJobs before the changes of the previous Pull Request). So, I thought that marking the build as canceled made sense to avoid other methods/functions thinking that the build/event is in progress.

However, after our discussion in the previous Pull Request, I removed the block and tried to pass the tests. The tests failed. The error I found was a race condition between cancelOutdatedEvents() and isAllBuildsCompleted(), because Jenkins requires time to abort builds. 

I asked Claude for a compressible description of the problem, we can discuss it later on:

---
The imprintEntry.setBuildCompleted(true) block at line ~410 addresses a critical race condition between the cancellation decision and the actual Jenkins cancellation process. Here's the sequence without this pre-marking:

  1. Thread A (processing new patchset) enters cancelOutdatedEvents() and decides build X should be canceled
  2. Thread A calls cancelMatchingJobs() which removes items from queue and interrupts executors
  3. Race window opens: Jenkins is canceling build X, but this takes time (queue removal, executor interruption, cleanup)
  4. Thread B (another event or completion check) calls isAllBuildsCompleted() to determine if feedback should be sent
  5. Thread B sees build X is NOT marked complete (because we're waiting for RunListener callback)
  6. Thread B incorrectly believes there are still active builds for the old patchset
  7. Eventually Jenkins finishes canceling, RunListener fires, and build X is marked complete

  The problem: During step 3-6, our memory state is inconsistent with our intent. We've decided build X is obsolete and should not block feedback, but the memory doesn't reflect this until the async RunListener callback completes.
   This can delay feedback for new patchsets or cause duplicate notifications.

  Why RunListener events aren't sufficient: The RunListener callback (onCompleted) is asynchronous and unpredictable in timing. When we're canceling due to build cancellation policy (new patchset supersedes old), we need synchronous, immediate marking to maintain memory consistency. The builds we're canceling are being rejected by policy - they should not block the new patchset's feedback while waiting for Jenkins' async cleanup.

  Regarding coordinated callbacks: This pre-marking does NOT break coordination. The isAllBuildsCompleted() check happens AFTER all matching cancellations complete in the same synchronized block. For example, if we cancel 3 builds for an old patchset, all 3 are marked complete before the synchronized block exits, maintaining atomicity. The coordinated notification logic remains intact - it just now correctly treats policy-canceled builds as "complete" immediately rather than waiting for async Jenkins cleanup.

  Moving to cancelMatchingJobs wouldn't help because the race condition exists at the decision level (in cancelOutdatedEvents), not at the execution level. We need to mark ALL matching builds as complete atomically BEFORE any of them are physically canceled.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/model/BuildMemory.java
ironcerocloudbees: Hey @rsandell ,

> I am confised by what you mean by serialize (to me that means writing the object to disk :) )

I meant to avoid concurrency. To avoid two threads running the same cancelOudatedEvents at the same time. 

> Anyway synchronizing on the storage class would be almost technically correct, you will be using the same monitor. But alot can have happened between the time when the synchronization was released by the return of storage.getAllEvents() and when you regain the monitor in this block. So even though there won't be any concurrent modification going on you would still be working with outdated data. And it is design wise very bad to have to know about and do external synchronization like that. And we will have to revisit again for the hazelcast implementation where I guess we will have to do some other form of distributed synchronization?

I see. We could resolve this problem moving the `Map<GerritTriggeredEvent, MemoryImprint> allEvents = storage.getAllEvents();` inside the `synchronized (storage) {` block, right? AFAIK, in Java, we can re-lock the same item in the same thread without issues (the JVM will acquire the second lock immediately, since the thread already holds the first).  What do you think about it? I tested, and all tests passed.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/model/BuildMemory.java
sboardwell: Discussed and agreed upon adding a new attribute/method `cancelling/isCancelling` adjacent to `cancelled`

- cancelling set first and cancelled set during actual cancellation
- review all calls to `isCancelled` to set if we need to be checking the `cancelled` or `cancelling` status

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spi/BuildMemoryStorage.java
sboardwell: So are we saying make

```
public interface BuildMemoryStorage {
```
 into 
```
public abstract class BuildMemoryStorage {
```

with all the trimmings? Ok, we can do that.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/model/BuildMemory.java
ironcerocloudbees: Done.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/model/BuildMemory.java
ironcerocloudbees: Moved. Apparently, it is working fine after the change.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spi/BuildMemoryStorage.java
ironcerocloudbees: Done.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/hudsontrigger/EventListener.java
ironcerocloudbees: Done.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/model/BuildMemory.java
sboardwell: Nice! 👍

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/model/BuildMemory.java
sboardwell: Why have we kept the `imprintEntry.setBuildCompleted(true);` here? I would think it follows the same logic as the cancelling and should not be set BEFORE the build is actually completed, no?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/storage/LocalBuildMemoryStorage.java
```diff
@@ -117,6 +117,7 @@ public synchronized void cancelled(@NonNull GerritTriggeredEvent event, @NonNull
         pb.set(project);
         Entry entry = pb.getEntry(project);
         entry.setCancelled(true);
+        entry.setCancelling(false);
         entry.setBuildCompleted(true);
```
sboardwell, line 121: This is where the `entry.setBuildCompleted(true);` happens, not before. If there are places which use `isCompleted()`, these should also take `isCancelling()` into account instead, no?

sboardwell (commented):
Just one issue to clarify regarding `imprintEntry.setBuildCompleted(true);`

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/storage/LocalBuildMemoryStorage.java
ironcerocloudbees: You are right, this setBuildCompleted should be there. I've removed it.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/gerritnotifier/model/BuildMemory.java
ironcerocloudbees: Removed.

sboardwell (approved):
LGTM 🚀

rsandell (approved):
Awesome work!!

sboardwell:
@ironcerocloudbees Could you perhaps update the PR description to reflect the changes we discussed. I can merge now if you want, but we should make sure the PR "does what is says on the box"

ironcerocloudbees:
Updated (go idea).

Additionally, I added a last commit with some changes that we missed: we did not convert the interface NotificationClaimStrategy to an abstract class (as we did with the BuildMemoryStorage). The change is simple (and I'm currently running the tests to confirm that we don't break anything).

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/coordination/LocalCoordinationProvider.java
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * <p>The {@code @Extension(ordinal = FALLBACK_PRIORITY)} annotation registers this provider
+ * with Jenkins with a very low ordinal, ensuring it is only selected when no other modes
+ * are available. Higher ordinal values take precedence.</p>
+ *
+ * @see LocalBuildMemoryStorage
+ * @see LocalNotificationClaimStrategy
+ * @see CoordinationModeFactory
+ */
+// CHECKSTYLE:OFF MagicNumber - Ordinal must be literal in annotation, -1000 ensures fallback priority
```
rsandell, line 50: Oh, I missed this one, not needed since you are using a constant.

sboardwell: Question for @ironcerocloudbees and @rsandell: if we used the direct setting of a system property `-Dgerrit.trigger.coordination.mode=hazelcast`, then ordinal should not matter because (to quote the Highlander) there can be only one.

PR #1021, opened 2026-07-27
Rename CoordinationModeFactory to CoordinationMode

rsandell:
The Factory suffix was misleading: the class never creates anything from a caller's perspective. All call sites use it as CoordinationMode.get().getStorage() etc. to access cached singletons. Its actual responsibilities are to select the winning CoordinationModeProvider once, memoize the component instances that provider yields, and hand them out - i.e. it represents the active coordination mode, not a factory. Invoking the provider's create*() methods is an internal detail hidden behind double-checked locking.

The providers keep their names. Although their create*() methods make them resemble abstract factories, creation is only part of their contract: they also model a mode's availability (isAvailable), identity (getModeName) and lifecycle (initialize/shutdown). That matches the established Provider shape (java.security.Provider, Jenkins CredentialsProvider): a pluggable module that reports availability and supplies a coherent family of implementations. Renaming them to *Factory would undersell the lifecycle aspect.

After this rename, Factory means the same thing everywhere in the plugin (e.g. GerritNotifierFactory, which does create per call), and the relationship reads forwards: a CoordinationModeProvider provides the implementations backing the active CoordinationMode.

Also renamed getSelectedMode() to getSelectedProvider() (it returns a CoordinationModeProvider) so "mode" unambiguously refers to the facade, and reworded the Javadoc that described the class as a factory.

### Testing done

Verified with mvn clean verify (compile, checkstyle, spotbugs, tests) and the test-hazelcast profile. TopicAssociationTriggerTest fails under the test-hazelcast profile, but fails identically on the base branch without this change - pre-existing, unrelated to the rename.

### Submitter checklist


sboardwell (approved):
LGTM 🚀

PR #1022, opened 2026-07-28
Implementation plan for implementing alternative coordination mode using redis

rsandell:
Just a suggested plan.

### Submitter checklist

### DIFF
--- a/REDIS_COORDINATION_MODE_PLAN.md
+++ b/REDIS_COORDINATION_MODE_PLAN.md
@@ -1,0 +1,362 @@
+# Redis Coordination Mode — Implementation Plan & Spec
+
+Status: approved plan, not yet started.
+Base branch: `distributed-storage`.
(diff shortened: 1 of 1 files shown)


panicking:
@rsandell If we support webhook, is really need to have a redis DB?

rsandell:
If you put some form of router in front of Jenkins then you will distribute the events evenly. But for example if a new patchset event is delivered to a different Jenkins controller than the one that is building the previous one you won't get the build current patchset only functionality because there is no way for the other controller to know that there is a new patchset and cancel the current builds.

panicking:
> If you put some form of router in front of Jenkins then you will distribute the events evenly. But for example if a new patchset event is delivered to a different Jenkins controller than the one that is building the previous one you won't get the build current patchset only functionality because there is no way for the other controller to know that there is a new patchset and cancel the current builds.

You are right I was not really think about all those features. I will follow up later about other points, now I start to understand the whole story

PR #1023, opened 2026-07-29
Missed Events Playback - distributed coordination

sboardwell:
This pull request introduces a new cross-replica missed-events coordination strategy to the distributed event management system, along with supporting infrastructure for pluggable implementations. It also adds a utility for generating per-JVM unique instance identifiers, which will be used to avoid file collisions in shared storage scenarios. The documentation is significantly expanded to clarify the design decisions, trade-offs, and future work around distributed locking and event catch-up guarantees.

**Distributed Missed-Events Coordination and Locking:**

* Added a new `MissedEventsCoordinationStrategy` SPI and corresponding implementation plumbing to `CoordinationMode`, enabling robust, cross-replica mutual exclusion for missed-events playback (prevents duplicate catch-up fetches in multi-Jenkins scenarios). [[1]](diffhunk://#diff-a8c856a809c3be46ee6c08cf55bb6233a7b915440f8feb53eee043e3fe823041R29) [[2]](diffhunk://#diff-a8c856a809c3be46ee6c08cf55bb6233a7b915440f8feb53eee043e3fe823041R129-R134) [[3]](diffhunk://#diff-a8c856a809c3be46ee6c08cf55bb6233a7b915440f8feb53eee043e3fe823041L237-R263) [[4]](diffhunk://#diff-a8c856a809c3be46ee6c08cf55bb6233a7b915440f8feb53eee043e3fe823041L300-R338) [[5]](diffhunk://#diff-a8c856a809c3be46ee6c08cf55bb6233a7b915440f8feb53eee043e3fe823041R356) [[6]](diffhunk://#diff-6cf247f562814f1adc4f0470dfd205eb0e3bddcde3e83df97487a468be7c28a7R32) [[7]](diffhunk://#diff-6cf247f562814f1adc4f0470dfd205eb0e3bddcde3e83df97487a468be7c28a7R130-R141)
* Implemented a local (single-JVM) fallback for missed-events coordination, ensuring safe operation even without distributed coordination. [[1]](diffhunk://#diff-a8c856a809c3be46ee6c08cf55bb6233a7b915440f8feb53eee043e3fe823041R356) [[2]](diffhunk://#diff-6cf247f562814f1adc4f0470dfd205eb0e3bddcde3e83df97487a468be7c28a7R130-R141)

**Instance Identity Utility:**

* Introduced the `InstanceIdentity` utility class to generate a unique, per-JVM identifier, which is critical for distinguishing state files in environments where multiple Jenkins controllers share a filesystem (e.g., NFS-backed `JENKINS_HOME`).

**Documentation and Design Rationale:**

* Added a comprehensive design document (`DISTRIBUTED_EVENTS_TOPOLOGY.md`) detailing cluster topology options, distributed locking mechanisms, guarantees, known risks, and the roadmap for future FencedLock support. This document provides essential context for both operators and future contributors.

These changes lay the groundwork for robust, scalable event coordination in distributed Jenkins deployments, clarify the limitations of current locking approaches, and prepare for future enhancements such as stronger consistency via Hazelcast CP Subsystem locks.

### Testing done

In maven and locally.

### Submitter checklist
- [x] Ensure you have provided tests that demonstrate the feature works or the issue is fixed


src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/coordination/hazelcast/HazelcastBuildMemoryStorage.java
```diff
@@ -40,6 +40,7 @@
 import com.sonyericsson.hudson.plugins.gerrit.trigger.hudsontrigger.NewPatchSetInterruption;
 import com.sonyericsson.hudson.plugins.gerrit.trigger.hudsontrigger.PipelineAbortHelper;
 import com.sonyericsson.hudson.plugins.gerrit.trigger.spi.BuildMemoryStorage;
+import com.sonyericsson.hudson.plugins.gerrit.trigger.spi.EventIdGenerator;
```
rsandell, line 43: unused import?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/coordination/hazelcast/HazelcastMissedEventsCoordinationStrategy.java
```diff
@@ -0,0 +1,176 @@
...
+            try {
+                long newWatermark = catchUpAction.fetchAndTrigger(new Date(effectiveLowerBound));
+                map.put(serverName, newWatermark);
+                return MissedEventsCatchUpOutcome.PERFORMED;
+            } catch (IOException e) {
+                logger.error("Missed-events catch-up failed for server {}", serverName, e);
+                return MissedEventsCatchUpOutcome.FAILED;
+            }
+        } finally {
+            maintenanceAction.run();
```
rsandell: this could potentially throw a RuntimeException and then it won't be unlocked until the timeout?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/coordination/InstanceIdentity.java
```diff
@@ -0,0 +1,94 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.lang.management.ManagementFactory;
+
+/**
+ * Generates an identifier that is unique among concurrently-running JVM processes of this plugin.
+ *
+ * <p>Used to name per-instance state files (e.g. missed-events playback timestamps) so that
+ * multiple Jenkins controller JVMs sharing one NFS-mounted {@code JENKINS_HOME} never write to
+ * the same file. Stability across restarts is deliberately NOT a goal here - directories of
```
rsandell, line 35: Interesting, I didn't think of that to be an option :)

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/coordination/LocalMissedEventsCoordinationStrategy.java
```diff
@@ -0,0 +1,122 @@
...
+import java.util.concurrent.locks.ReentrantLock;
+
+/**
+ * Single-JVM implementation of {@link MissedEventsCoordinationStrategy}.
+ *
+ * <p>One instance of this class is shared (via {@link CoordinationMode}) by every {@code
+ * GerritMissedEventsPlaybackManager} in the JVM, so the per-server lock and watermark maps here
+ * are genuinely shared state - mirroring how a distributed implementation shares state via a
+ * cluster-wide map. In a single JVM there is no other process to coordinate with, so this reduces
+ * to a plain in-process lock: no lease is needed, since {@code finally} always releases it on the
```
rsandell: So why is a lock needed at all?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/coordination/LocalMissedEventsCoordinationStrategy.java
```diff
@@ -0,0 +1,122 @@
...
+            return MissedEventsCatchUpOutcome.LOCK_TIMEOUT;
+        }
+        try {
+            long watermark = watermarks.getOrDefault(serverName, 0L);
+            // See HazelcastMissedEventsCoordinationStrategy's identical fix: the watermark only
+            // reflects how far a PAST catch-up reached, not whether a NEW gap has opened since -
+            // gating on "watermark >= candidateCatchUpFrom" alone would permanently suppress
+            // catch-up after the first successful fetch on an otherwise-quiet server. Use the more
+            // advanced of the two as the actual fetch lower bound instead of skipping outright.
+            long effectiveLowerBound = Math.max(watermark, candidateCatchUpFrom);
```
rsandell: Looks like this would be needed on every implementation. Can it be shared somehow?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsPlaybackManager.java
```diff
@@ -86,18 +88,53 @@
...
+     * pruned. Default is 7 days.
+     */
+    private static final String STALE_INSTANCE_FILE_AGE_PROPERTY =
+            "gerrit.trigger.playback.instance.stale.age.hours";
+    private static final int DEFAULT_STALE_INSTANCE_FILE_AGE_HOURS = 168;
+    /**
+     * Used by {@link #roundUpToWholeSecond} to round a sub-second catch-up lower bound up to the
+     * whole-second resolution the events-log plugin's own {@code t1} query parameter is limited to.
+     */
+    private static final long MILLIS_PER_SECOND = 1000L;
```
rsandell: Use `TimeUnit` calculations instead or the `Duration` approach suggested above, if possible.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsPlaybackManager.java
```diff
@@ -630,24 +785,48 @@ public boolean isRunning() {
...
+            if (ownTimeSlice > lastPublishedTimeSlice) {
+                lastPublishedTimeSlice = ownTimeSlice;
+                coordinationStrategy.publishInstanceFreshness(serverName, ownTimeSlice);
+            }
+
+            // Persist the best currently-known value - this instance's own, or a fresher one
+            // borrowed from a peer via the shared signal - to this instance's own local file, so a
+            // later cold-start file scan sees cluster-wide freshness, not just what this specific
+            // instance itself ever directly observed.
+            long sharedTimeSlice = coordinationStrategy.getSharedInstanceFreshness(serverName).orElse(Long.MIN_VALUE);
```
rsandell: What if `publishInstanceFreshness` returned the actual "shared" value that is published? Wouldn't that remove one "network call" and make it a bit quicker?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/InstanceTimestampStore.java
```diff
@@ -0,0 +1,275 @@
...
+ * layout, {@code gerrit-server-event-data/<serverName>/}) and this JVM instance, since one
+ * {@link GerritMissedEventsPlaybackManager} - and therefore one store - already exists per
+ * configured server. {@code serverName} is deliberately NOT interchangeable between stores:
+ * the persisted timestamp is later used verbatim as the catch-up query's lower bound against
+ * that specific server's own front-end URL, so mixing servers would silently skip or re-trigger
+ * events for whichever server's watermark got overwritten.</p>
+ *
+ * <p>{@link #computeMaxTimestampAcrossInstances()} falls back to the legacy single-file path
+ * ({@link #getLegacyConfigXml()}) when the instances directory is empty, so upgrades from a
+ * pre-existing single-JVM deployment are not treated as "no known timestamp."</p>
```
rsandell, line 57: 👍

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/InstanceTimestampStore.java
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    /**
+     * Deletes instance files that have not been rewritten in more than {@code maxAgeMillis},
+     * indicating a permanently decommissioned JVM - orphaned files would otherwise accumulate
+     * forever. Never deletes this store's own instance file, regardless of its age.
+     *
+     * @param maxAgeMillis the staleness threshold, based on the file's last-modified time.
+     * @return the number of files removed.
+     */
```
rsandell, line 253: Do we want to include the legacy file in this perhaps?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spi/MissedEventsCatchUpAction.java
```diff
@@ -0,0 +1,46 @@
...
+ *
+ *  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ *  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ *  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ *  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ *  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ *  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ *  THE SOFTWARE.
+ */
+package com.sonyericsson.hudson.plugins.gerrit.trigger.spi;
```
rsandell, line 22: Shouldn't these belong ion the playback package?

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/InstanceTimestampStoreTest.java
```diff
@@ -0,0 +1,191 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.mockStatic;
+import static org.mockito.Mockito.when;
+
+/**
+ * Tests for {@link InstanceTimestampStore}.
+ */
+public class InstanceTimestampStoreTest {
+
+    private static final long TIMESTAMP_1000 = 1000L;
```
rsandell: I usually tell checkstyle to f off "because test data" with a `//CS INGNORE MagicNumber FOR NEXT...`

You can see examples in other test classes.

rsandell:
Overall the approach looks ok. I couldn't figure out why you need a lock at all in local mode, that might be due to the late hour that I'm reading it :)

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/coordination/hazelcast/HazelcastBuildMemoryStorage.java
sboardwell: not on my end. Still need to push some changes though

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/coordination/LocalMissedEventsCoordinationStrategy.java
sboardwell: still in flux

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spi/MissedEventsCatchUpAction.java
sboardwell: Computer says no 🙂 

> These four types (MissedEventsCatchUpAction, MissedEventsCatchUpOutcome, MissedEventsCatchUpResult, MissedEventsCoordinationStrategy) follow the same shape as the existing EventClaimStrategy/NotificationClaimStrategy pair already in spi: an abstract strategy discovered via CoordinationModeProvider's Extension Points pattern, with concrete implementations living in coordination (LocalMissedEventsCoordinationStrategy) and coordination/hazelcast (HazelcastMissedEventsCoordinationStrategy), and only consumed — not implemented — from playback.

> Putting them in playback would mean the coordination implementations depend on the playback package purely to satisfy an interface contract, which reverses today's dependency direction (playback depends on spi + coordination, not the other way around). Keeping them in spi matches the existing convention for pluggable coordination strategies in this codebase, so I'd lean toward leaving it as-is unless there's a reason to diverge from that pattern here.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/coordination/hazelcast/HazelcastBuildMemoryStorage.java
sboardwell: I think it's needed

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/coordination/hazelcast/HazelcastMissedEventsCoordinationStrategy.java
sboardwell: Added finally block to handle and unlock - fixes just in

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/coordination/LocalMissedEventsCoordinationStrategy.java
sboardwell: Yes. Javadoc updated - fixes just in

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/coordination/LocalMissedEventsCoordinationStrategy.java
sboardwell: Good catch. Shared via template class/function - fixes just in

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsPlaybackManager.java
sboardwell: Done - fixes just in

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsPlaybackManager.java
sboardwell: Done  - fixes just in

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/InstanceTimestampStore.java
sboardwell: Done  - fixes just in

src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/InstanceTimestampStoreTest.java
sboardwell: Nice. Done  - fixes just in

sboardwell:
All PR comments addressed.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsPlaybackManager.java
```diff
@@ -222,13 +361,33 @@ protected synchronized Date getDateFromTimestamp() {
...
+     *
+     * <p>This runs on every reconnect, not just process cold-start: each JVM keeps its own live
+     * Gerrit connection and event processing is already deduplicated across JVMs via the event
+     * claim strategy, so this cross-instance catch-up specifically covers the case where every
+     * JVM was simultaneously unable to process events (e.g. a full outage).</p>
+     *
+     * <p>Waits on {@link JobsLoadedGate} first, before computing anything below - including the
+     * candidate catch-up timestamp itself. This runs on {@code GerritConnection}'s own background
+     * thread (see that class), not on Jenkins' own initializer thread, so blocking here doesn't
+     * delay Jenkins' own boot - but on a cold start, this method can otherwise run (and replay a
```
rsandell, line 382: But you are blocking other `ConnectionListeners` from doing their thing. At this moment I think it's only the `GerritProjectListUpdater` and there's most likely no issues delaying that until jobs are loaded.

sboardwell: Agreed on `GerritProjectListUpdater` - it's registered before the playback manager, so it's already run its `connectionEstablished()` by the time this blocks. Not doing much there anyway (flag + `notify()`).

The wait also holds up `GerritConnection`'s read loop, since `notifyConnectionEstablished()` runs before the stream starts. But I thought that was expected and intentional - without it, live events could hit the same silent-drop race as replayed ones, before job triggers register. Blocking here closes both at once.

The gate runs before `checkIfEventsLogPluginSupported()` though, so it delays those servers without `events-log` too, even though catch-up doesn't apply to them. Might be intentional (same live-event race either way), just flagging though just in case.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsPlaybackManager.java
```diff
@@ -86,23 +92,155 @@
...
 
-    private static final String GERRIT_SERVER_EVENT_DATA_FOLDER = "/gerrit-server-event-data/";
     private static final Logger logger = LoggerFactory.getLogger(GerritMissedEventsPlaybackManager.class);
     static final String EVENTS_LOG_PLUGIN_NAME = "events-log";
     private static final String EVENTS_LOG_PLUGIN_URL = "a/plugins/" + EVENTS_LOG_PLUGIN_NAME + "/events/";
-    private static final String GERRIT_TRIGGER_SERVER_TIMESTAMPS_XML = "gerrit-trigger-server-timestamps.xml";
+    /**
+     * System property: maximum age in hours a per-instance timestamp file may go without being
+     * rewritten before it is considered orphaned (from a permanently decommissioned JVM) and
+     * pruned. Default is 7 days.
```
rsandell, line 102: The more "modern" way would be to use `Duration` I believe
Something like


```suggestion
    private static final Duration DEFAULT_STALE_INSTANCE_FILE_AGE = Duration.ofDays(7);
```

rsandell: I thought it was weird that you hadn't adressed this comment, somehow this wasn't posted from last time I reviewed. Nothing blocking though.

sboardwell: Sorry, just blindly applied the change and then realised the change wasn't actually a suggested change. Will revert @rsandell

sboardwell: Fixed in the latest commit

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/JobsLoadedGate.java
```diff
@@ -0,0 +1,110 @@
...
+    /**
+     * Upper bound on how long {@link #await()} will block before giving up and letting the
+     * caller proceed anyway. Not expected to ever be hit in practice ({@link
+     * hudson.init.InitMilestone#JOB_CONFIG_ADAPTED} is reached during ordinary Jenkins startup) -
+     * this exists only so a stalled/broken initializer elsewhere can't wedge missed-events
+     * catch-up for the lifetime of the JVM.
+     */
+    private static final long AWAIT_TIMEOUT_MINUTES = 5;
+
+    private static final CountDownLatch LATCH = new CountDownLatch(1);
```
rsandell, line 90: Again, having static fields like this is risky and your first instinct for creating a global value should be to use `@Extension`s. 
By using that you would potentially cover the situation where reload from disk etc reestablishes the JOB_CONFIG_ADAPTED milestone.
But is using the static field was intentional perhaps because my scenario isn't working as I think it does?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/JobsLoadedGate.java
```diff
@@ -0,0 +1,110 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(JobsLoadedGate.class);
+
+    /**
+     * Upper bound on how long {@link #await()} will block before giving up and letting the
+     * caller proceed anyway. Not expected to ever be hit in practice ({@link
+     * hudson.init.InitMilestone#JOB_CONFIG_ADAPTED} is reached during ordinary Jenkins startup) -
+     * this exists only so a stalled/broken initializer elsewhere can't wedge missed-events
+     * catch-up for the lifetime of the JVM.
+     */
+    private static final long AWAIT_TIMEOUT_MINUTES = 5;
```
rsandell: In my experience startup times can be waay longer than this. But perhaps there has been improvements in this area since I last measured it?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/JobsLoadedGate.java
sboardwell: I was under the impression we would do that later, no?

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/JobsLoadedGate.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    /**
+     * Opens the gate - called once, from the {@code @Initializer(after =
+     * InitMilestone.JOB_CONFIG_ADAPTED)} hook in {@link
+     * com.sonyericsson.hudson.plugins.gerrit.trigger.PluginImpl}. Safe to call more than once
+     * (e.g. under test); {@link CountDownLatch#countDown()} is a no-op once the count has already
+     * reached zero.
+     */
+    public static void open() {
```
rsandell, line 102: You should be able to annotate this method directly with `@Initializer()` instead of calling it from another one. But it's also a good idea to have plugin initialization gathered in one class to find it easier. So nothing to change here I think, just a note.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/JobsLoadedGate.java
rsandell: That was only for the HazelcastInstance class not new code.
But I am ok with letting this one go for now as I trust your testing has been thorough, and changing it would probably mean you need to retest a lot.

rsandell (approved):

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/GerritMissedEventsPlaybackManager.java
rsandell: because my first `Duration` comment didn't get posted maybe this one was confising? :)

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/JobsLoadedGate.java
sboardwell: The static field is intentional here, and I checked the specific scenario you raised against Jenkins
core's own source (jenkins-core 2.479.3, jenkins/model/Jenkins.java).

`doReload()` (the "Reload Configuration from Disk" action) calls `Jenkins#reload()`, whose javadoc
says explicitly:

> Beware that this calls neither ItemListener#onLoaded nor Initializers.

Its task graph in `loadTasks()` only *requires* `JOB_CONFIG_ADAPTED` for the "Finalizing set up" step
- it never re-*attains* that milestone, and no `@Initializer`-annotated method runs during reload. So
`gerritJobsLoaded()` (our `@Initializer(after = JOB_CONFIG_ADAPTED)` hook) won't fire again on a
config reload, and there's nothing for the gate to reset.

Given that, an `@Extension` wouldn't buy us anything for this milestone - there's no Jenkins event to
hook that re-establishes it. Happy to add a code comment citing this if it'd help future readers avoid
the same question.

(Separate from what you raised: a single job being resaved/reloaded later re-registers *that job's*
GerritTrigger listener without touching this milestone at all - a narrower race the gate was never
meant to cover, since it's scoped to boot-time timing rather than per-job listener registration.)

rsandell: Yes, a javadoc comment about it would be very appreciated.

rsandell:
There is a lot of commits here, do you want to squash merge it or do a normal merge? I
m fine either or.

src/main/java/com/sonyericsson/hudson/plugins/gerrit/trigger/playback/JobsLoadedGate.java
sboardwell: Done

sboardwell:
@rsandell the tests are timing out for some reason. Not happening in either JDK17 or 21 on my machine. Shall we still merge? I squash or merge. You choose 🙂

PR #1024, opened 2026-07-30
Fix TopicAssociationTriggerTest flakiness under Hazelcast coordination

rsandell:
The test asserted build numbers immediately after triggerAndWait(), which relies solely on JenkinsRule.waitUntilNoActivity(). But gerritServer.triggerEvent() is asynchronous: it hands the event to the GerritHandler worker queue and returns, and waitUntilNoActivity() only observes the Jenkins queue/executors, not the plugin's in-flight event processing. The Hazelcast event claim (distributed map init + get/putIfAbsent round-trips) adds enough per-event latency that waitUntilNoActivity() sees an idle Jenkins and returns before the build is scheduled, so getLastBuild() is null and the assertion NPEs. In local mode the window is microseconds, so it passed by luck.

Wait for the expected completed build via TestUtils.waitForBuilds() before reading build numbers. Because notifyListeners() dispatches the event to every listener synchronously and completes before any build executes, waiting for the job that IS expected to build also makes the "did not build" assertions on the other job deterministic.

Verified passing in both the default and -Ptest-hazelcast profiles, including single-CPU taskset stress runs.

### Submitter checklist


sboardwell (approved):
Looks good to me 🚀

PR #519, opened 2025-08-03
remove usages of commons lang 2

mawinter69:
replace with native java

### Submitter checklist
- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [ ] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [ ] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

PR #1030, opened 2026-08-08
Migrate the last Commons Lang 2 usage to native Java

timja:
- see https://github.com/jenkinsci/jenkins/issues/16404

#519 removed the Commons Lang 2 usages from this plugin, but it missed one file: `PluginChangeMergedEvent`, the sibling of `PluginPatchsetCreatedEvent` which that PR did migrate. It still has `import org.apache.commons.lang.StringUtils` on `master`, so the plugin will break at runtime once Jenkins core stops exposing Commons Lang 2 (jenkinsci/jenkins#26105).

This applies exactly the same transformation #519 used on the sibling file — `StringUtils.isNotEmpty(x)` becomes `x != null && !x.isEmpty()` — so no new dependency is introduced.

It also sets `<ban-commons-lang-2.skip>false</ban-commons-lang-2.skip>`. That property was never enabled here, which is precisely why the missed file went unnoticed: the enforcer rule was not running. With it on, a regression fails the build.

### Testing done

- `mvn clean install` passes with the enforcer enabled.
- Verified the enforcer is genuinely active, rather than silently skipping: restoring the Commons Lang 2 import with the property in place fails the build with `Reason: Use Commons Lang 2 (org.apache.commons.lang.*)`.
- `PluginChangeMergedEventTest` — 7 tests, 0 failures.

### A related problem this does not fix

`pct-build-failure-analyzer` in jenkinsci/bom#6234 fails with `NoClassDefFoundError: org/apache/commons/lang/StringUtils` thrown from `com.sonymobile.tools.gerrit.gerritevents.GerritHandler`, reached through this plugin. That is the `gerrit-events` library rather than plugin code, so it needs sonyxperiadev/gerrit-events#131 to merge and release, then a `gerrit-events` bump here.

---
This change was prepared with AI assistance (Claude Code); it has been reviewed and built locally by me.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -61,6 +61,7 @@
         <changelist>999999-SNAPSHOT</changelist>
         <gitHubRepo>jenkinsci/${project.artifactId}-plugin</gitHubRepo>
         <no-test-jar>false</no-test-jar>
+        <ban-commons-lang-2.skip>false</ban-commons-lang-2.skip>
         <findbugs.failOnError>false</findbugs.failOnError>
         <build-failure-analyzer-plugin.version>3.887.vc872cf1b_7dff</build-failure-analyzer-plugin.version>
         <checkstyle.version>3.6.0</checkstyle.version>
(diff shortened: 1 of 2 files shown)


rsandell (approved):

timja:
@rsandell possible to ship please?

This is one of five remaining PRs to get PCT passing for this migration

PR #1031, opened 2026-08-11
Create NOTICE

raul-arabaolaza:
Put the disclaimer about support in a visible place


docs/NOTICE
```diff
@@ -0,0 +1 @@
+Jenkins does not support multiple replicas or nodes for a single logical instance, this feature is not tested with Jenkins. This feature is provided for CloudBees CI (Enterprise Jenkins). This feature is provided as a community effort and is not endorsed or officially supported by CloudBees.
```
jtnord: one sentence per line.
Enterprise Jenkins name is no longer used (so remove).

```suggestion
Jenkins does not support multiple replicas or nodes for a single logical instance, this feature is not tested with Jenkins.
This feature is provided for use with CloudBees CI and is provided as a community effort which is neither endorsed nor supported by CloudBees.
```

but this is a generic NOTICE, so this would apply to which feature?  perhaps the whole plugin but then we should say plugin not feature. (but if we state the feature, is that implying that other features are supported?)

